PR #14, opened 2017-03-20
PLT-5896 Move user actions to use the available 3.7 v4 endpoints

jwilander:
#### Summary
Move user actions to use the available v4 endpoints that are in 3.7. Based on #13 

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-5986

#### Checklist

### DIFF
--- a/src/actions/preferences.js
+++ b/src/actions/preferences.js
@@ -3,7 +3,7 @@
 
 import {batchActions} from 'redux-batched-actions';
 
-import {Client} from 'client';
+import {Client, Client4} from 'client';
 import {Preferences} from 'constants';
 import {PreferenceTypes} from 'action_types';
 import {getMyPreferences as getMyPreferencesSelector} from 'selectors/entities/preferences';
@@ -13,12 +13,12 @@ import {getPreferenceKey} from 'utils/preference_utils';
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
 import {getLogErrorAction} from './errors';
 
-export function deletePreferences(preferences) {
+export function deletePreferences(userId, preferences) {
     return async (dispatch, getState) => {
         dispatch({type: PreferenceTypes.DELETE_PREFERENCES_REQUEST}, getState);
 
         try {
-            await Client.deletePreferences(preferences);
+            await Client4.deletePreferences(userId, preferences);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch);
             dispatch(batchActions([
@@ -65,17 +65,17 @@ export function makeDirectChannelVisibleIfNecessary(otherUserId) {
                 value: 'true'
             };
 
-            await savePreferences([preference])(dispatch, getState);
+            await savePreferences(currentUserId, [preference])(dispatch, getState);
         }
     };
 }
 
-export function savePreferences(preferences) {
+export function savePreferences(userId, preferences) {
     return async (dispatch, getState) => {
         dispatch({type: PreferenceTypes.SAVE_PREFERENCES_REQUEST}, getState);
 
         try {
-            await Client.savePreferences(preferences);
+            await Client4.savePreferences(userId, preferences);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch);
             dispatch(batchActions([
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -208,13 +208,13 @@ export function logout() {
     );
 }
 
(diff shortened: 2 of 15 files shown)


src/actions/users.js
```diff
@@ -224,6 +224,8 @@ export function getProfiles(offset, limit = General.PROFILE_CHUNK_SIZE) {
             return null;
         }
 
+        const profiles = profileListToMap(profilesList);
```
hmhealey: Should we maybe be changing the reducers to handle a list or profiles instead of converting the list into a map and then passing it down?

src/actions/users.js
```diff
@@ -376,31 +412,52 @@ export function getStatusesByIds(userIds) {
...
-export function revokeSession(id) {
-    return bindClientFunc(
-        Client.revokeSession,
-        UserTypes.REVOKE_SESSION_REQUEST,
-        [UserTypes.RECEIVED_REVOKED_SESSION, UserTypes.REVOKE_SESSION_SUCCESS],
-        UserTypes.REVOKE_SESSION_FAILURE,
-        id
-    );
+export function revokeSession(userId, sessionId) {
+    return async (dispatch, getState) => {
```
hmhealey: Why's this one not using `bindClientFunc`?

jwilander: Good question, I'll switch it over

src/actions/users.js
jwilander: I think it's better to be stored as a map in the store for ease of access but I could probably be convinced to store it as an array if I was presented with some good arguments for it. Either way it's a bit out of scope for this ticket and we can revisit when I start working on the store itself

src/actions/users.js
jwilander: Oh it's because I want to use `sessionId` when I dispatch the action and `bindClientFunc` only attaches the data from the driver response (which doesn't have the session id)

src/actions/preferences.js
```diff
@@ -13,12 +13,12 @@ import {getPreferenceKey} from 'utils/preference_utils';
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
 import {getLogErrorAction} from './errors';
 
-export function deletePreferences(preferences) {
+export function deletePreferences(userId, preferences) {
```
enahum: dont know about passing the userId as you should have it in the preferences, right?

src/actions/users.js
enahum: yeah I think its ok having a map in the store and the API can return whatever, we don't need to match the API or the store to the same object model, actually I think that was a mistake in APIv3

hmhealey: Oh sorry, I still want it a map in the store, but I meant we should could pass it as an array in the action since I'm almost certain most of those reducers just loop over the map anyway

src/actions/users.js
hmhealey: Ah gotcha

src/actions/users.js
jwilander: It does an assign https://github.com/mattermost/mattermost-redux/blob/master/src/reducers/entities/users.js#L103

src/actions/preferences.js
jwilander: What if you're the admin and you to delete someone else's preference?

enahum: this is an example of a preference

```js
preference = {
                user_id: currentUserId,
                category: Preferences.CATEGORY_DIRECT_CHANNEL_SHOW,
                name: otherUserId,
                value: 'true'
            };
```

so if you set the prop to the userId that you want, that should be enough, right?

jwilander: Right but you're passing in an array of preferences, I find it odd to pick one of the preferences in the array at random to use the user id for. I think it's much better to just be explicit

enahum: Alright then

src/actions/users.js
jwilander: Oh I think I know you mean. Are you saying pass the action the array then in the store convert to a map so we don't have to convert to a map in a bunch of places? I wouldn't be opposed to that but there is the case where you do have a map in the action and don't want to have to convert map -> array -> map

jwilander: What about two different action types RECEIVED_PROFILES for when you have the map and RECEIVED_PROFILES_LIST which will convert the list to a map?

hmhealey: Yeah that sounds good

PR #15, opened 2017-03-21
PLT-5518 RN: Allow user to upload + send image file attachments

csduarte:
#### Summary
This PR adds the ability for images to be attached to posts.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-5926

#### Checklist

#### Test Information
This PR was tested on:
iPhone 6 iOS 10
iPhone 7 iOS 10

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "deep-equal": "1.0.1",
     "harmony-reflect": "1.5.1",
     "isomorphic-fetch": "2.2.1",
+    "mime-db": "^1.27.0",
     "redux": "3.6.0",
     "redux-batched-actions": "0.1.5",
     "redux-thunk": "2.2.0",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "deep-equal": "1.0.1",
     "harmony-reflect": "1.5.1",
     "isomorphic-fetch": "2.2.1",
-    "mime-db": "^1.27.0",
+    "mime-db": "1.27.0",
     "redux": "3.6.0",
     "redux-batched-actions": "0.1.5",
     "redux-thunk": "2.2.0",
--- a/src/actions/files.js
+++ b/src/actions/files.js
@@ -36,3 +36,48 @@ export function getFilesForPost(teamId, channelId, postId) {
         ]), getState);
     };
 }
+
+const uploadFileRequests = {};
+export function uploadFile(teamId, channelId, fileFormData, formBoundary, rootId, requestId) {
+    return async (dispatch, getState) => {
+        dispatch({type: FilesTypes.UPLOAD_FILES_REQUEST}, getState);
+
+        uploadFileRequests[requestId] = true;
+
+        let files;
+        try {
+            files = await Client.uploadFile(teamId, channelId, fileFormData, formBoundary);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch);
+            dispatch(batchActions([
+                {type: FilesTypes.UPLOAD_FILES_FAILURE, error},
+                getLogErrorAction(error)
+            ]), getState);
+            return;
+        }
+
+        // check to see if our upload file request was cancelled
+        if (!uploadFileRequests[requestId]) {
+            return;
+        }
+
+        dispatch(batchActions([
+            {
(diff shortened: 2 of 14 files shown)


enahum:
@jarredwitt also take a look at why tests are failing

package.json
```diff
@@ -6,6 +6,7 @@
     "deep-equal": "1.0.1",
     "harmony-reflect": "1.5.1",
     "isomorphic-fetch": "2.2.1",
+    "mime-db": "^1.27.0",
```
jwilander: We should probably remove the `^` so it doesn't auto-upgrade

src/actions/files.js
```diff
@@ -36,3 +36,48 @@ export function getFilesForPost(teamId, channelId, postId) {
...
+            }
+        ]), getState);
+    };
+}
+
+export function cancelUploadFileRequest(requestId) {
+    uploadFileRequests[requestId] = false;
+    return {
+        type: FilesTypes.UPLOAD_FILES_CANCEL
+    };
```
jwilander: Unless I'm mistaken, this doesn't actually cancel the actual upload, it just prevents any actions being dispatched once the upload does complete. Maybe not a priority for now but we should at least create a ticket for actually cancelling the upload

src/actions/files.js
```diff
@@ -36,3 +36,48 @@ export function getFilesForPost(teamId, channelId, postId) {
         ]), getState);
     };
 }
+
+const uploadFileRequests = {};
+export function uploadFile(teamId, channelId, fileFormData, formBoundary, rootId, requestId) {
```
jwilander: Where does `requestId`  come from? Does the caller need to generate this?

test/actions/files.test.js
```diff
@@ -39,7 +39,7 @@ describe('Actions.Files', () => {
         const formBoundary = imageFormData.getBoundary();
 
         const fileUploadResp = await basicClient.
-            uploadFile(basicTeam.id, basicChannel.id, clientId, imageFormData, formBoundary);
+            uploadFile(basicTeam.id, basicChannel.id, imageFormData, formBoundary);
```
jwilander: I know we're testing the client driver with this but a whole action just got added that doesn't have any code coverage, can we add a unit test for the new action?

We also made changes to the create post action, can we update the unit test for that action to cover the new code?

src/actions/files.js
jarredwitt: Correct it does not cancel the actual upload. When the file is being uploaded a loader is displayed on the UI side since the upload request status is `started`. This is provided so that the loader can be removed from the UI and the user can continue on using the app.

src/actions/files.js
jarredwitt: Yes, the caller generates the requestId since it will be needed to cancel the request. Added this in case the user cancels an upload that would otherwise complete successfully. Since we aren't cancelling the actual upload we use this to check if the request was cancelled before adding the newly uploaded images to the files state.

test/actions/files.test.js
jarredwitt: Mistake on my part, I'll add a unit test for upload and update the create post action test.

package.json
jarredwitt: I'll make the change.

src/actions/files.js
jwilander: Ok, I think that's fine for now but I think in the future we should look at passing back the actual request ID generated by the driver (which can also be used to actually cancel the upload). I'm not familiar enough with this new fangled `await` and `dispatch` stuff but I think we'll need a new pattern for being to return a value while also returning the `async` to wait on

src/actions/files.js
jwilander: Cool, I've created a ticket here for it in the future https://mattermost.atlassian.net/browse/PLT-5990

src/actions/files.js
jarredwitt: Agreed, we can probably find a way to make it work with the redux-store instead of returning a value. That way we can keep using the `await` and `dispatch` like the other actions. Assume we need a ticket for this?

jwilander: I'll add it to the description of the ticket I made below

jarredwitt:
@jwilander added tests.

PR #24, opened 2017-03-23
PLT-5968 prevent profile request to update current user in the store

enahum:
#### Summary
Preventing the request for the profile of the current user and exclude the current user from requests that returns the user as part of it.

This is done because some requests for profiles return the user sanitized which in the case of the current user we don't want that.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-5968


enahum:
@jwilander build is passing.

Btw you should consider this for master

jwilander:
Yes, plan is to manually merge all the 3.7 changes over into `master` at some point before we switch over

PR #32, opened 2017-03-27
Move more actions to v4 and remove autocomplete and user search from store

jwilander:
#### Summary
Move more actions to v4 and remove autocomplete and user search from store.

#### Checklist


src/actions/general.js
```diff
@@ -49,7 +49,7 @@ export function getClientConfig() {
 
 export function getLicenseConfig() {
     return bindClientFunc(
-        Client.getLicenseConfig,
+        Client4.getClientLicenseOld,
```
enahum: what is with the old? API 4 returns something different?

src/actions/general.js
```diff
@@ -94,8 +94,8 @@ export function setServerVersion(serverVersion) {
 
 export function setStoreFromLocalData(data) {
     return async (dispatch, getState) => {
-        Client.setToken(data.token);
-        Client.setUrl(data.url);
+        Client4.setToken(data.token);
+        Client4.setUrl(data.url);
```
enahum: Im afraid that if we do this the `Client` won't have the token or the url set? can we keep it in both for now?

src/actions/general.js
jwilander: We return the client config and license as a map[string]string. I have a ticket in my backlog to fix it, hence why I'm doing this here

src/actions/general.js
jwilander: Yep, I can set both

PR #35, opened 2017-03-30
Fix android WS connection and weakmap when channel not in the store

enahum:
#### Summary
On Android there is an issue where the WebSocket won't connect if we don't remove the port from the URI if the port is 80 or 443, this will fix the issue.

Also this fixes a weakmap in android when we receive a post and the channel does not exists in the store

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -670,30 +670,34 @@ export function markChannelAsUnread(channelId, mentionsArray) {
         const state = getState();
         const {channels, myMembers} = state.entities.channels;
         const {currentUserId} = state.entities.users;
-        const channel = {...channels[channelId]};
-        const member = {...myMembers[channelId]};
 
-        if (channel && member) {
-            channel.total_msg_count++;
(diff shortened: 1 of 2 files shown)


src/actions/channels.js
```diff
@@ -670,30 +670,34 @@ export function markChannelAsUnread(channelId, mentionsArray) {
...
+        if (channels[channelId] && myMembers[channelId]) {
+            const channel = {...channels[channelId]};
+            const member = {...myMembers[channelId]};
 
-            let mentions = [];
-            if (mentionsArray) {
-                mentions = JSON.parse(mentionsArray);
-                if (mentions.indexOf(currentUserId) !== -1) {
-                    member.mention_count++;
+            if (channel && member) {
```
jarredwitt: Is this check still needed since line 675 would imply that we will have a channel and a member?

enahum: yes right I'll remove it, thanks

enahum:
@jarredwitt ready dude

PR #44, opened 2017-04-04
PLT-5960 Add admin entities for config, logs and audits (part1) 

jwilander:
#### Summary
Add admin entities for config, logs and audits. I'm using `nock` for mocking the server, it's really simple. Currently I'll only use it for admin endpoints that need to be mocked but I think long term we should move towards using it for all the action tests so that we can run them without having to stand-up and configure an MM server

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-5960

#### Checklist

### DIFF
--- a/test/actions/admin.test.js
+++ b/test/actions/admin.test.js
@@ -1,0 +1,130 @@
+// Copyright (c) 2017 Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+import assert from 'assert';
+import nock from 'nock';
+
+import * as Actions from 'actions/admin';
+import {Client, Client4} from 'client';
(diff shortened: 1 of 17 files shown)


test/actions/admin.test.js
```diff
@@ -0,0 +1,130 @@
...
+        nock.restore();
+        await TestHelper.basicClient.logout();
+        await TestHelper.basicClient4.logout();
+    });
+
+    it('getLogs', async () => {
+        nock(Client4.getBaseRoute()).
+            get('/logs').
+            query(true).
+            reply(200, [
```
enahum: this looks good, lets use this at least for now

PR #46, opened 2017-04-05
PLT-6208 Follow permission policies set in System Console (3.7)

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6208

### DIFF
--- a/src/constants/constants.js
+++ b/src/constants/constants.js
@@ -68,9 +68,6 @@ const PostsTypes = {
 
 const Permissions = {
     PERMISSIONS_ALL: 'all',
-    PERMISSIONS_CHANNEL_ADMIN: 'channel_admin',
-    PERMISSIONS_TEAM_ADMIN: 'team_admin',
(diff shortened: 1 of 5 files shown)


src/constants/constants.js
```diff
@@ -75,10 +66,35 @@ const PostsTypes = {
     REMOVE_FROM_CHANNEL: 'system_remove_from_channel'
 };
 
+const Permissions = {
+    PERMISSIONS_ALL: 'all',
+    PERMISSIONS_CHANNEL_ADMIN: 'channel_admin',
+    PERMISSIONS_TEAM_ADMIN: 'team_admin',
+    PERMISSIONS_SYSTEM_ADMIN: 'system_admin',
```
jwilander: We should probably call these `ROLE_SYSTEM_ADMIN`, etc. since they're roles and not permissions

enahum: cool let me change them, this was actually a copy/paste from the webapp, but it makes total sense

enahum:
@jwilander done

PR #50, opened 2017-04-06
Remove every comment from the deleted root post (master)

enahum:


jwilander:
Looks good but you have some build errors

enahum:
@jwilander is should be passing now

PR #57, opened 2017-04-10
Add initial state directly into service layer

jwilander:
#### Summary
Add the initial state here so that the users of the service don't need to worry about any of the internal reducers when initializing state. They still have the option to override if they choose to.


jwilander:
I haven't tried it but I can't see why it wouldn't

enahum:
Try it out to see, maybe ask @jarredwitt as I know it was causing issues in the past

jwilander:
Which part of this PR are you talking about? The initial state bit or the npm/babel changes?

enahum:
Npm babel

jwilander:
This repo shouldn't be using anything react native specific, otherwise the whole point of it is moot. If there's issues on RN because of that we can deal with those when we move it over to `master`

enahum:
Then we need to figure out a way not to transpile this on post-install

jwilander:
Sorry @enahum, I just pushed another commit. I've been using this branch with my work for the webapp :P

PR #58, opened 2017-04-11
RN-68: Make save and delete preferences optimistic

csduarte:
#### Summary
This PR adds the ability to handle optimistic actions for save and delete preferences. It also addresses a configuration issue with configureOfflineStore when building for release.

#### Ticket Link
https://mattermost.atlassian.net/browse/RN-68

#### Checklist

#### Test Information
This PR was tested on:
iPhone 6 iOS 10
iPhone 7 iOS 10

@jarredwitt


src/reducers/entities/preferences.js
```diff
@@ -10,6 +10,8 @@ function getKey(preference) {
 
 function myPreferences(state = {}, action) {
     switch (action.type) {
+    case PreferencesTypes.SAVE_PREFERENCES_OPTIMISTIC:
+    case PreferencesTypes.SAVE_PREFERENCES_ROLLBACK:
```
jwilander: I think you missed adding this to the key mirror

jarredwitt: Yeah, this PR is getting a major work over. I'm making the offline store the default as well as making the save and delete preferences optimistic on the redux side instead of the client side. Have it all working just need to fix the tests. Hopefully submit something soon.

jwilander: Ok cool, that was going to be my next question ;)

jarredwitt:
@jwilander @enahum ready for you.

src/reducers/entities/preferences.js
```diff
@@ -10,6 +10,7 @@ function getKey(preference) {
 
 function myPreferences(state = {}, action) {
     switch (action.type) {
+    case PreferencesTypes.SAVE_PREFERENCES:
```
enahum: why the reducer does the same when saving optimistically and when committing? I get that `RECEIVED_PREFERENCES` is called whenever we call `getPreferences` but this is adding the same preference twice once when saving and once on commit.

Also I thought we discuss that for `getMyPreferences` we were going to set the state to whatever was returned from the server.

I think im somehow lost here, care to clarify?

src/actions/preferences.js
```diff
@@ -1,43 +1,30 @@
...
-            },
-            {
-                type: PreferencesTypes.DELETE_PREFERENCES_SUCCESS
+    return async (dispatch) => {
+        dispatch({
+            type: PreferencesTypes.DELETE_PREFERENCES,
+            data: preferences,
+            meta: {
+                offline: {
+                    effect: () => Client.deletePreferences(preferences),
```
enahum: another worry that I have, what if this action is retrying after a while and the session has expired? can we stop retrying if the response has status 401?

jarredwitt: In the docs here https://github.com/jevakallio/redux-offline#giving-up-is-hard-to-do

"If server responded with HTTP 4xx client error, always discard"

src/actions/preferences.js
```diff
@@ -95,29 +82,18 @@ export function makeGroupMessageVisibleIfNecessary(channelId) {
...
+    return async (dispatch) => {
+        dispatch({
+            type: PreferencesTypes.SAVE_PREFERENCES,
+            data: preferences,
+            meta: {
+                offline: {
+                    effect: () => Client.savePreferences(preferences),
+                    commit: {
+                        type: PreferencesTypes.RECEIVED_PREFERENCES
+                    }
```
jwilander: Do we need to specify rollback or is that handled somewhere else?

src/reducers/entities/preferences.js
```diff
@@ -10,20 +10,26 @@ function getKey(preference) {
...
+        if (action.data) {
+            for (const preference of action.data) {
+                nextState[getKey(preference)] = preference;
+            }
         }
 
         return nextState;
     }
+    case PreferencesTypes.DELETE_PREFERENCES:
     case PreferencesTypes.DELETED_PREFERENCES: {
```
jwilander: Why two action types? Can't we just use the same one?

jwilander:
Overall looks good, just a couple comments

src/actions/preferences.js
jarredwitt: I can add one in. The rollback in this case won't run for an hour since it'll be ran against the default decaying schedule.

src/reducers/entities/preferences.js
jarredwitt: Yeah we should just use one action. I'll make the change.

jarredwitt:
@jwilander changes made.

PR #61, opened 2017-04-13
Preferences and GM (master)

enahum:
* Get preferences when websocket reconnects
* Fetch GM channel if not in the store


jwilander:
Got some build errors

PR #65, opened 2017-04-17
Fixes

enahum:
#### Summary
This PR adds the following:
* Fix preferences constants in channel utils
* Fix handleRequest helper to correctly parse an Error object
* Fix profilesListToSet in case the data is undefined
* Fix searchChannels and add logClientError
* Fix ping
* Fix uploadFile to include channelId or rootId when dispatching received files
* Fix load post on WS reconnect
* Fix members in team when data is undefined
* Fix options headers and add file Urls
* Add MyChannels and OtherChannels selectors
* Set existing notify_props for currentUser on getMe and made it so getProfiles do not replace currentUser
* Make WS to connect using APIv4
* Remove getMoreChannels and `MORE_CHANNELS` Action Types
* Add function to Filter profiles by term
* Make it so the current channel reset the count but does not update the `last_view_at` when marking the channel as read

#### Checklist


src/client/client4.js
```diff
@@ -643,6 +678,18 @@ export default class Client4 {
...
     };
 
+    logClientError = async (message, level = 'ERROR') => {
+        const body = {
+            message,
+            level
+        };
+
+        return this.doFetch(
+            `${this.getBaseRoute()}/general/log_client`,
```
jwilander: This endpoint doesn't exist yet, we should probably just use the v3 one for now

enahum:
@jwilander done

PR #70, opened 2017-04-19
Add a bunch of actions for the webapp

jwilander:
#### Summary
Add a bunch of actions/selectors for the webapp and do some clean-up.

#### Checklist


jwilander:
@enahum For changes you might need to care about for RN:
* Removed `updateNotifyProps` action and added `updateMe` which just does a patch (so just pass it `{notify_props: propsgohere}`)
 * I also didn't make it optimistic because I don't think we want it to be
* Refactored the `userAutocompleteInChannel` to be general `userAutocomplete`
* Switched the selector names for users from `getProfilesInThing` to `getUserIdsInThings` so it's more clear what they do

jwilander:
@jarredwitt in case you want to take a look as well

enahum:
@jarredwitt there are changes in this PR that we need to consider for RN, I know I've done a bunch of changes for https://github.com/mattermost/mattermost-mobile/pull/484 so lets merge that one first and then we can make the necessary changes if any

src/actions/users.js
```diff
@@ -339,7 +408,7 @@ export function getProfilesInChannel(channelId, page, perPage = General.PROFILE_
             },
             {
                 type: UserTypes.RECEIVED_PROFILES_LIST,
-                data: profiles
+                data: removeUserFromList(currentUserId, Object.assign([], profiles))
```
enahum: I think here instead of Object.assign you can do `[...profiles]`

src/actions/users.js
```diff
@@ -514,41 +591,62 @@ export function getUserAudits(userId, page = 0, perPage = General.AUDITS_CHUNK_S
...
+                    id: channelId
+                }
+            );
+        }
+
+        if (teamId) {
+            actions.push(
+                {
+                    type: UserTypes.RECEIVED_PROFILES_LIST_IN_TEAM,
+                    data: Object.assign([], data.users, data.out_of_channel),
```
enahum: same here, and a few more above and below

jwilander:
@enahum changes made

PR #73, opened 2017-04-21
Fix broken offlineCompose invocation

jwilander:
#### Summary
When building the webapp for production it was erroring out because the third argument was undefined. Related to the redux-offline changes I assume


jwilander:
Also if you could let me merge after review that'd be great

PR #76, opened 2017-04-24
RN-54 Make leave channel optimistic

enahum:
#### Summary
Makes leave channel action optimistic

also fixes a constant typo in preferences actions and adds `.babelrc` to `.npmignore` so whoever implements the library can transpile the lib using its own `.babelrc` file

#### Ticket Link
https://mattermost.atlassian.net/browse/RN-54

#### Device Information
This PR was tested on: 
* IOS 10.3.1 iPhone 6s
* Android 7.1.1 Nexus 7
* Android 6.0.1 Samsung J5

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -381,30 +381,25 @@ export function getChannelMembers(channelId, page = 0, perPage = General.CHANNEL
 
 export function leaveChannel(channelId) {
     return async (dispatch, getState) => {
-        dispatch({type: ChannelTypes.LEAVE_CHANNEL_REQUEST}, getState);
+        const state = getState();
+        const {currentUserId} = state.entities.users;
+        const {channels} = state.entities.channels;
+        const channel = channels[channelId];
 
-        const {currentUserId} = getState().entities.users;
-
-        try {
-            await Client4.removeFromChannel(currentUserId, channelId);
-        } catch (error) {
-            forceLogoutIfNecessary(error, dispatch);
-            dispatch(batchActions([
-                {type: ChannelTypes.LEAVE_CHANNEL_FAILURE, error},
-                getLogErrorAction(error)
-            ]), getState);
-            return;
-        }
-
-        dispatch(batchActions([
-            {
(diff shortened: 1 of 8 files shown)


src/actions/channels.js
```diff
@@ -381,30 +381,25 @@ export function getChannelMembers(channelId, page = 0, perPage = General.CHANNEL
...
+        dispatch({
+            type: ChannelTypes.LEAVE_CHANNEL,
+            data: {id: channelId, user_id: currentUserId},
+            meta: {
+                offline: {
+                    effect: () => Client4.removeFromChannel(currentUserId, channelId),
+                    commit: {type: ChannelTypes.LEAVE_CHANNEL},
+                    rollback: {
+                        type: ChannelTypes.RECEIVED_CHANNEL,
+                        data: channel
```
jwilander: Is this the proper rollback? Wouldn't it be join channel?

jwilander:
Also just a general question, how are we testing rollbacks? Is there an easy way to do this with the unit tests?

src/actions/channels.js
enahum: Join channel eventually dispatches `RECEIVED_CHANNEL` so this is basically the same thing, but what is not happening here is `RECEIVED_MY_CHANNEL_MEMBER` so this won't work, I'll need to make the change

enahum:
good question for testing rollbacks, let me see if I can find an easy way to do just that

jarredwitt:
I think I have an idea using the test store and the discard config.

src/actions/channels.js
jwilander: Also make sure that roles is set correctly when doing `RECEIVED_MY_CHANNEL_MEMBER`

enahum:
@jwilander @jarredwitt fixed the rollback and added tests to confirm that rollback and commit are working as expected.

Needed to fix the store config helper as the network detection needed a callback.

Be aware that this PR also adds `.babelrc` to npm ignore

PR #77, opened 2017-04-24
Add team actions and other changes for webapp

jwilander:
#### Summary
Add team actions and other changes for webapp. @enahum I think the only change to be known for RN is that the `createTeam` action no longer takes a user id as an argument

#### Checklist

### DIFF
--- a/src/reducers/entities/teams.js
+++ b/src/reducers/entities/teams.js
@@ -90,7 +90,9 @@ function membersInTeam(state = {}, action) {
         if (data && data.length) {
             const nextState = {...state};
             for (const member of data) {
-                if (!nextState[member.team_id]) {
+                if (nextState[member.team_id]) {
+                    nextState[member.team_id] = {...nextState[member.team_id]};
+                } else {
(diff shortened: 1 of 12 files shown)


src/reducers/entities/teams.js
```diff
@@ -140,7 +168,7 @@ export default combineReducers({
     // object where every key is the team id and has and object with the team members detail
     myMembers,
 
-    // object where every key is the team id and has a Set of user ids that are members in the team
+    // object where every key is the team id and has an of members in the team where the key is user id
```
enahum: this changed from set to object?

jwilander: Yeah, didn't make sense as a set of ids since we need to store the member details

enahum: ok, we just need to be careful when changing these cause when doing a rehydration it could error out and the store might be invalid at some point

jwilander: Yep agreed, we'll need a process to change entities once we go stable

PR #79, opened 2017-04-25
PLT-4457 Changed getProfilesAndStatusesForPosts to get at mentioned users

hmhealey:
This adds the `/users/usernames` added in https://github.com/mattermost/platform/pull/6218 and uses it to get the profiles for any users that were mentioned in the post. There's also some logic to attempt to capture usernames ending in periods as well as ones that don't which just end a sentence.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-4457

#### Checklist
- All new/modified APIs include changes to the drivers

#### Test Information

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -326,15 +326,11 @@ export function getNeededAtMentionedUsernames(state, posts) {
 
         const pattern = /\B@(([a-z0-9_.-]*[a-z0-9_])[.-]*)/gi;
 
-        while (true) {
+        let match;
+        while ((match = pattern.exec(post.message)) !== null) {
             // match[1] is the matched mention including trailing punctuation
             // match[2] is the matched mention without trailing punctuation
-            const match = pattern.exec(post.message);
-
-            if (!match) {
-                // No more matches in this post
-                break;
-            } else if (usersByUsername[match[1]] || usersByUsername[match[2]]) {
+            if (usersByUsername[match[1]] || usersByUsername[match[2]]) {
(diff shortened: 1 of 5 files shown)


src/actions/posts.js
```diff
@@ -264,31 +265,83 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
...
     const posts = list.posts;
-    const profilesToLoad = [];
-    const statusesToLoad = [];
 
-    Object.keys(posts).forEach((key) => {
-        const post = posts[key];
+    const userIdsToLoad = new Set();
+    const statusesToLoad = new Set();
+
+    Object.keys(posts).forEach((postId) => {
```
enahum: Why not do a `Object.values` so you get each post directly?

src/actions/posts.js
```diff
@@ -264,31 +265,83 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
...
+
+    return Promise.all(promises);
+}
+
+export function getNeededAtMentionedUsernames(state, posts) {
+    let usersByUsername; // Populate this lazily since it's relatively expensive
+
+    const usernamesToLoad = new Set();
+
+    Object.keys(posts).forEach((postId) => {
```
enahum: same here

src/actions/posts.js
hmhealey: Oh yeah, I can change that. I forgot that existed and was just following the old pattern :p

PR #83, opened 2017-04-28
Webapp actions/selectors for channels

jwilander:
#### Summary
More webapp actions/selectors. @enahum note that I changed the `members` channel entity to be `membersInChannel` and follow the same pattern as the one for teams.

#### Checklist


enahum:
@jwilander this is failing, now another question would be why the in the entities reducer?

src/actions/channels.js
```diff
@@ -117,6 +117,10 @@ export function createDirectChannel(userId, otherUserId) {
             last_update_at: created.create_at
         };
 
+        const preference = [{category: Preferences.CATEGORY_DIRECT_CHANNEL_SHOW, name: otherUserId, value: 'true'}];
+
+        savePreferences(userId, preference);
```
enahum: note to self: remove the save preference in the RN action when creating a direct channel as is going to be handled by this

src/actions/channels.js
```diff
@@ -128,7 +132,7 @@ export function createDirectChannel(userId, otherUserId) {
             },
             {
                 type: PreferenceTypes.RECEIVED_PREFERENCES,
-                data: [{category: Preferences.CATEGORY_DIRECT_CHANNEL_SHOW, name: otherUserId, value: 'true'}]
+                data: preference
```
enahum: can we name this `preferences` as is an array and not just the preference object?

jwilander:
It passed, locally...

You mean, why the `membersInChannel`? Because it's an object with channel id keys of objects with user id keys with members as values

src/reducers/entities/channels.js
```diff
@@ -181,33 +179,47 @@ function myMembers(state = {}, action) {
     }
 }
 
-function members(state = {}, action) {
-    const nextState = {...state};
-
+function membersInChannel(state = {}, action) {
```
enahum: note to self: Rename `members` to `membersInChannel` wherever this is used in RN

src/selectors/entities/channels.js
```diff
@@ -28,8 +32,8 @@ export function getMyChannelMemberships(state) {
     return state.entities.channels.myMembers;
 }
 
-export function getChannelMembers(state) {
-    return state.entities.channels.members;
+export function getChannelMembersInChannels(state) {
```
enahum: note to self: rename this too in the RN app

src/actions/channels.js
jwilander: Yeah good point, I'll update this

jwilander:
Merging because I'm pretty sure the random test failures aren't related to this PR

PR #87, opened 2017-05-02
Ability to retry failed file uploads

enahum:
#### Summary
With this we properly handle when upload a file fails so we can retry the upload before creating the post

### DIFF
--- a/src/actions/files.js
+++ b/src/actions/files.js
@@ -7,7 +7,6 @@ import {Client4} from 'client';
 import {FileTypes} from 'action_types';
 import {getLogErrorAction} from './errors';
 import {forceLogoutIfNecessary} from './helpers';
-import {parseClientIdsFromFormData} from 'utils/file_utils';
 
 export function getFilesForPost(postId) {
     return async (dispatch, getState) => {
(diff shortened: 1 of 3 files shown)


src/actions/files.js
```diff
@@ -46,8 +47,13 @@ export function uploadFile(channelId, rootId, fileFormData, formBoundary) {
             files = await Client4.uploadFile(fileFormData, formBoundary);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch);
-            dispatch(batchActions([
-                {type: FileTypes.UPLOAD_FILES_FAILURE, error},
+            dispatch(batchActions([{
+                type: FileTypes.UPLOAD_FILES_FAILURE,
+                clientIds: parseClientIdsFromFormData(fileFormData),
```
jwilander: Might be better to pass the client IDs in rather than parse them here but not a big deal 0/5

src/actions/files.js
```diff
@@ -46,8 +47,13 @@ export function uploadFile(channelId, rootId, fileFormData, formBoundary) {
             files = await Client4.uploadFile(fileFormData, formBoundary);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch);
-            dispatch(batchActions([
-                {type: FileTypes.UPLOAD_FILES_FAILURE, error},
+            dispatch(batchActions([{
```
hmhealey: Something is really funky with the indentation here. Do you mind making it clearer here which argument is for which function?

enahum:
@jwilander @hmhealey updated

PR #94, opened 2017-05-11
Add selectors for preferences and user admin checks

jwilander:
#### Summary
Add selectors for preferences and user admin checks.

#### Checklist

### DIFF
--- a/src/selectors/entities/preferences.js
+++ b/src/selectors/entities/preferences.js
@@ -1,16 +1,14 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See License.txt for license information.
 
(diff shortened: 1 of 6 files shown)


src/selectors/entities/preferences.js
```diff
@@ -4,3 +4,24 @@
 export function getMyPreferences(state) {
     return state.entities.preferences.myPreferences;
 }
+
+function getKey(category, name) {
```
enahum: this one already exists in `utils/preference_utils.js`

jwilander: Ah, didn't notice that. I'll switch to using that

jwilander:
Fixed @enahum

PR #95, opened 2017-05-11
Updates from the webapp branch

jwilander:
#### Summary
General fixes/updates I made to the webapp branch. @enahum not sure if the view channel changes will affect RN. If they don't work let me know and I can make it support the old behaviour as well

Note that `loadMe` changed to just use all the actions instead of the client directly. If that's an issue for RN we can chat about it as well

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -4,7 +4,6 @@
 import {General, Preferences} from 'constants';
 import {ChannelTypes, PreferenceTypes, UserTypes} from 'action_types';
 import {savePreferences} from 'actions/preferences';
-import {setServerVersion} from 'actions/general';
 import {batchActions} from 'redux-batched-actions';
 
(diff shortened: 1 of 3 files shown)


src/actions/channels.js
```diff
@@ -568,6 +561,8 @@ export function viewChannel(channelId) {
             return null;
         }
 
+        setServerVersion(Client4.getServerVersion())(dispatch, getState);
```
enahum: why this? there is an event that is fired anytime the serverVersion changes

jwilander: Does it update the store anytime it changes in the client? I wasn't seeing that happen so I just had it get set on a regular call like view channel

enahum: if you listen for the event you can use that to update the store accordingly

jwilander:
@enahum fixed

PR #97, opened 2017-05-15
PLT-6585 Properly leave channels when leaving a team

hmhealey:
We weren't removing the now-deleted channel members when we left a team so they would still show up in the sidebar upon rejoining the team.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6585

#### Test Information
This PR was tested on: Chrome on OS X

### DIFF
--- a/src/actions/teams.js
+++ b/src/actions/teams.js
@@ -4,7 +4,7 @@
 import {batchActions} from 'redux-batched-actions';
 import {Client4} from 'client';
 import {General} from 'constants';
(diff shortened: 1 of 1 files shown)


jwilander:
Got some build failures. Let me know if they're not related, if so we can just ignore them and I'll fix when I merge into master

hmhealey:
@jwilander A few of them were from changes I made, but the remaining eslint error is from code that I didn't touch

jwilander:
Ok cool, I'll fix that one when I merge into master

PR #108, opened 2017-05-22
Actions/selectors required for post list changes

jwilander:
#### Summary
Added a bunch of actions and selectors required for the post list. Added opengraph and reaction entities.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-5955
https://mattermost.atlassian.net/browse/PLT-5954

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,6 @@
     "build": "babel src --out-dir .",
     "check": "node_modules/.bin/eslint --ext \".js\" --ignore-pattern node_modules --quiet .",
     "test": "NODE_ENV=test mocha --opts test/mocha.opts",
-    "postinstall": "rm -f .babelrc && npm run build"
+    "postinstall": "npm run build"
   }
 }
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -7,6 +7,8 @@ import {Client4} from 'client';
 import {Preferences, Posts} from 'constants';
 import {PostTypes, FileTypes} from 'action_types';
 
+import * as Selectors from 'selectors/entities/posts';
+
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
 import {getLogErrorAction} from './errors';
 import {deletePreferences, savePreferences} from './preferences';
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -44,6 +44,6 @@
     "build": "babel src --out-dir .",
     "check": "node_modules/.bin/eslint --ext \".js\" --ignore-pattern node_modules --quiet .",
     "test": "NODE_ENV=test mocha --opts test/mocha.opts",
-    "postinstall": "rm -f .babelrc && npm run build"
+    "postinstall": "npm run build"
```
enahum: nothing to do here, lets go ahead and I'll test if this doesn't break anything in RN

src/actions/posts.js
```diff
@@ -123,6 +125,176 @@ export function editPost(post) {
...
+
+        return posts;
+    };
+}
+
+export function addReaction(postId, emojiName) {
+    return async (dispatch, getState) => {
+        dispatch({type: PostTypes.REACTION_REQUEST}, getState);
+
+        const currentUserId = getState().entities.users.currentUserId;
```
enahum: just as an FYI you can do

```js
```

src/actions/users.js
```diff
@@ -201,6 +201,25 @@ export function getProfiles(page = 0, perPage = General.PROFILE_CHUNK_SIZE) {
     };
 }
 
+export function getMissingProfilesByIds(userIds) {
```
enahum: nice!! there are a few places where we do this all the time, and we do it also for statuses like for `getProfilesAndStatusesForMembers` in `/src/actions/teams.js`

src/reducers/entities/posts.js
```diff
@@ -217,6 +217,66 @@ function currentFocusedPostId(state = '', action) {
     }
 }
 
+function reactions(state = {}, action) {
+    switch (action.type) {
+    case PostTypes.RECEIVED_REACTIONS: {
+        const reactionsList = action.data;
+        const nextReactions = {...(state[action.postId] || {})};
```
enahum: i'm 0/5 but we could declare this const before the switch so we don't need to declare them every time inside each case

src/selectors/entities/posts.js
```diff
@@ -21,6 +43,99 @@ export const getPostsInCurrentChannel = createSelector(
     }
 );
 
+export function makeGetPostsInChannel() {
```
enahum: This is great, @hmhealey maybe after this goes in you can replace the one we have in RN in `channel_post_list` with this one and have it only in one place ;)

PR #110, opened 2017-05-22
Multiple Fixes

enahum:
#### Summary
* Make getChannel selector to complete direct channel info
* markChannelAsRead if channel and member are in the store
* Mark channel as unread if post is from webhook


jwilander:
Looks good, but you've got a build failure

PR #117, opened 2017-05-26
More actions and selectors for posts/files/reactions

jwilander:
#### Summary
More actions and selectors for posts/files/reactions.

#### Checklist


src/actions/files.js
```diff
@@ -37,6 +37,19 @@ export function getFilesForPost(postId) {
     };
 }
 
+export function getMissingFilesForPost(postId) {
```
enahum: is the action `getFilesForPost` not enough?

src/store/configureStore.prod.js
```diff
@@ -15,7 +15,7 @@ import serviceReducer from 'reducers';
 import initialState from './initial_state';
 import {offlineConfig} from './helpers';
 
-export default function configureOfflineServiceStore(preloadedState, appReducer, userOfflineConfig, enableBuffer = true) {
+export default function configureOfflineServiceStore(preloadedState, appReducer, userOfflineConfig, getAppReducer, enableBuffer = true) {
```
enahum: `getAppReducer` should not be here as it is only use for the hot module loader

src/utils/file_utils.js
```diff
@@ -74,3 +75,15 @@ export function parseClientIdsFromFormData(formdata) {
 
     return clientIds;
 }
+
+export function getFileUrl(fileId) {
```
enahum: nothing to do here, @jarredwitt maybe we could replace the ones in RN for this methods?

src/actions/files.js
jwilander: Only want to get them if we don't have them, it'll be a perf problem otherwise since this is called on mount

src/store/configureStore.prod.js
jwilander: The prod and dev signatures for this function need to match, doesn't matter if that's not used here

PR #118, opened 2017-05-29
Include team reads and unreads

enahum:
#### Summary
Handle messages to mark the team as read or unread

#### Ticket Link
 https://mattermost.atlassian.net/browse/RN-137

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -810,7 +810,7 @@ export function markChannelAsRead(channelId, prevChannelId) {
 
                 let teamMsgCount = 0;
                 if (totalMsgCount) {
-                    teamMsgCount =  totalMsgCount - channelMember.msg_count;
+                    teamMsgCount = totalMsgCount - channelMember.msg_count;
                 }
                 teamMember.msg_count = teamMember.msg_count - teamMsgCount;
             }
@@ -837,7 +837,7 @@ export function markChannelAsRead(channelId, prevChannelId) {
 
                     let teamMsgCount = 0;
                     if (prevTotalMsgCount) {
-                        teamMsgCount =  prevTotalMsgCount - prevChannelMember.msg_count;
+                        teamMsgCount = prevTotalMsgCount - prevChannelMember.msg_count;
                     }
                     teamMember.msg_count = teamMember.msg_count - teamMsgCount;
                 }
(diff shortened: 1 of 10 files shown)


src/actions/channels.js
```diff
@@ -793,14 +793,22 @@ export function updateChannelPurpose(channelId, purpose) {
...
-        if (channels[channelId] && channelMember) {
-            totalMsgCount = channels[channelId].total_msg_count;
+        let teamMember;
+        if (channel && channelMember) {
+            totalMsgCount = channel.total_msg_count;
+
+            if (channel.team_id) {
+                teamMember = {...teamMembers[channel.team_id]};
+                teamMember.mention_count = teamMember.mention_count - channelMember.mention_count;
+                teamMember.msg_count = teamMember.msg_count - (totalMsgCount ? (totalMsgCount - channelMember.msg_count) : totalMsgCount);
```
jwilander: This line is a bit confusing to read but not a big deal

src/actions/channels.js
```diff
@@ -827,17 +843,34 @@ export function markChannelAsRead(channelId, prevChannelId) {
...
+        }
+
         if (actions.length) {
             dispatch(batchActions([...actions]), getState);
         }
     };
 }
 
-export function markChannelAsUnread(channelId, mentionsArray) {
+export function markChannelAsUnread(teamId, channelId, mentionsArray) {
```
jwilander: Is it possible to get the team id from the channel object so we don't need to pass it in here? If it's too much work then this is fine, 0/5

src/actions/users.js
```diff
@@ -151,11 +158,13 @@ export function loadMe() {
         const prefs = getMyPreferences()(dispatch, getState);
         const teams = getMyTeams()(dispatch, getState);
         const members = getMyTeamMembers()(dispatch, getState);
+        const teamUnreads = getMyTeamUnreads()(dispatch, getState);
 
         await me;
         await prefs;
         await teams;
         await members;
+        await teamUnreads;
```
jwilander: Do we need to wait on this or is it ok if it pops in a short time after loading? Just wondering if we can remove the `await`

src/actions/channels.js
enahum: The problem with that is you may get a post from a channel in a team that hasn't been loaded yet and then you don't have the channel in the store, no channel in the store means no team_id, sounds about right?

src/actions/users.js
enahum: can do

enahum:
@jwilander done

PR #119, opened 2017-05-29
Add action to get the status of a user

jwilander:
#### Summary
Add action to get the status of a user


src/action_types/users.js
```diff
@@ -52,6 +52,10 @@ export default keyMirror({
     PROFILES_STATUSES_SUCCESS: null,
     PROFILES_STATUSES_FAILURE: null,
 
+    PROFILES_STATUS_REQUEST: null,
```
enahum: not that really matters but we could make this `PROFILE_STATUS...` as it is only for one profile

jwilander: Oh yeah, I'll change that

PR #125, opened 2017-05-30
on reconnect always get myTeams

enahum:
#### Summary
On a previous PR we remove the `getMyTeams` from the WS reconnection, this adds it back as well as calling `getTeams` to be sure we are getting the open teams from the server


enahum:
closing and re-opening to trigger jenkins

PR #126, opened 2017-06-01
Add addCommand, regenCommandToken and deleteCommand actions and other…

94117nl:
… needed stuff


src/actions/integrations.js
```diff
@@ -134,3 +134,36 @@ export function updateOutgoingHook(hook) {
         hook
     );
 }
+
+export function addCommand(teamId, command) {
+    return bindClientFunc(
+        Client.addCommand,
```
jwilander: Please use `Client4`, we're moving away from API v3 which is what `Client` uses. You might need to add the functions into `src/client/client4.js` for any missing ones you need. You can use https://api.mattermost.com/v4 for reference

94117nl: Moved it to v4

jwilander:
To use these changes in the webapp you'll want to point the `yarn.lock` file to the latest commit on the `webapp-master` branch on this repo

PR #129, opened 2017-06-02
Add action for uploading brand image

jwilander:
#### Summary
Add action for uploading brand image. @enahum this endpoint is available in 3.8 https://github.com/mattermost/platform/blob/release-3.8/api4/brand.go#L19

#### Checklist

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -150,6 +150,10 @@ export default class Client4 {
         return `${this.getBaseRoute()}/emoji/${emojiId}`;
     }
 
+    getBrandRoute() {
+        return `${this.getBaseRoute()}/brand`;
+    }
+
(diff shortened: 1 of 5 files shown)


src/client/client4.js
```diff
@@ -1097,6 +1101,19 @@ export default class Client4 {
         );
     };
 
+    uploadBrandImage = async (imageData) => {
```
enahum: don't you need to include the `formBoundry` as with regular file upload?

jwilander: I don't think so since there's only the one attachment, not multiple fields like in the regular file upload. It works for the webapp as is

enahum: ok we can change it later if needed

PR #134, opened 2017-06-04
Add actions for LDAP and SAML

jwilander:
#### Summary
Add actions for LDAP and SAML


jwilander:
Need to add remove cert actions

PR #140, opened 2017-06-08
User actions

jwilander:
#### Summary
Add some user actions. @enahum note I also added the form boundary to upload profile and brand image. I think that should take care of it for RN as well

#### Checklist

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -79,7 +79,7 @@ export function createUser(user, data, hash, inviteId) {
     };
 }
 
-export function login(loginId, password, mfaToken = '') {
+export function login(loginId, password, mfaToken = '', ldapOnly = false) {
     return async (dispatch, getState) => {
         dispatch({type: UserTypes.LOGIN_REQUEST}, getState);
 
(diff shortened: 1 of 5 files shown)


src/actions/users.js
```diff
@@ -79,15 +79,15 @@ export function createUser(user, data, hash, inviteId) {
...
+export function login(loginId, password, mfaToken = '', ldapOnly = false) {
     return async (dispatch, getState) => {
         dispatch({type: UserTypes.LOGIN_REQUEST}, getState);
 
         const deviceId = getState().entities.general.deviceToken;
 
         let data;
         try {
-            data = await Client4.login(loginId, password, mfaToken, deviceId);
+            data = await Client4.login(loginId, password, mfaToken, deviceId, ldapOnly);
```
enahum: this is great and all, but how do you decide is `ldapOnly`? asking to see if RN needs to do something to set that value

jwilander: It's something the webapp has an action for, not exactly sure where it's used. I don't think you need to worry about it on RN. I just wanted to add the support here since the server allows it

PR #142, opened 2017-06-09
Add an internal reducer to store the last time a channel fetched posts

enahum:
#### Summary
Instead of using the last update_at from the posts that we have in the store, this creates an internal store that has the last time a channel fetched posts and that timestamp will be used to `getPostSince` that way we can avoid to miss messages.

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -5,7 +5,7 @@ import {batchActions} from 'redux-batched-actions';
 
 import {Client, Client4} from 'client';
 import {Preferences, Posts} from 'constants';
-import {PostTypes, FileTypes} from 'action_types';
+import {InternalTypes, PostTypes, FileTypes} from 'action_types';
 
 import * as Selectors from 'selectors/entities/posts';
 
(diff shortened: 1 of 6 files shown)


src/actions/posts.js
```diff
@@ -360,7 +360,18 @@ export function getPosts(channelId, page = 0, perPage = Posts.POST_CHUNK_SIZE) {
             return null;
         }
 
-        dispatch(batchActions([
+        const actions = [];
+        if (Object.keys(posts.posts).length === Posts.POST_CHUNK_SIZE) {
```
jwilander: Won't this set the timestamp if you scroll up and load more posts in the channel? I think we only want to set it if you don't have any posts in the channel yet unless I'm mistaken

enahum:
If you scroll up it will use getPostBefore, right?

jwilander:
On the webapp we only use that for permalink view, and use the paging here to get more posts

enahum:
You are right, I'll fix it by checking that the page number is 0, but I'll do that over the weekend as I'm not at my desk

jwilander:
Ok cool, that should work

enahum:
@jwilander done

src/actions/posts.js
```diff
@@ -360,7 +360,18 @@ export function getPosts(channelId, page = 0, perPage = Posts.POST_CHUNK_SIZE) {
             return null;
         }
 
-        dispatch(batchActions([
+        const actions = [];
+        if (page === 0 && Object.keys(posts.posts).length === Posts.POST_CHUNK_SIZE) {
```
jwilander: I don't think we need the length check do we? What happens in a newer channel with only a few posts?

enahum: Well at leadt in RN when that happens we call getPosts again

jwilander: Ok, up to you if you want to change it or not

PR #149, opened 2017-06-13
Add brand image URL function to client

jwilander:
#### Summary
Add brand image URL function to client.


enahum:
@jwilander looks good but you have some build failures

PR #155, opened 2017-06-14
Only load custom emojis when they're enabled on the server

hmhealey:


enahum:
@hmhealey eslint errors

PR #159, opened 2017-06-15
Only use version response header when cache control not set

jwilander:
#### Summary
On the webapp, if someone changes something in the system console then the `X-Version-Id` header will change to let the client know it needs to reload. Responses cached by the browser might still hold the old version header. If we use an old version header to update the current version, and then a non-cached request gives us the new version then we can get stuck in a loop of changing the version (and causing a redirect loop on the webapp).

There is apparently no good way of telling if a response given to you from a browser is from the browser cache or actually from the server. The workaround I've added here is to ignore the version header from any response that has caching control. Most of our requests do not so this shouldn't affect much but it should prevent any cached responses updating the version header.

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -1406,7 +1406,9 @@ export default class Client4 {
             };
         }
 
-        if (headers.has(HEADER_X_VERSION_ID)) {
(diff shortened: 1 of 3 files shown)


enahum:
@jwilander build failures

PR #160, opened 2017-06-15
Add ability to add client middleware to redux lib

csduarte:
This PR adds the ability to add middleware from the client using the `additionalMiddleware` property of the `clientOptions` argument.

@jarredwitt

### DIFF
--- a/src/store/configureStore.dev.js
+++ b/src/store/configureStore.dev.js
@@ -15,11 +15,21 @@ import {General} from 'constants';
 import serviceReducer from 'reducers';
 import deepFreezeAndThrowOnMutation from 'utils/deep_freeze';
 
-import {offlineConfig} from './helpers';
+import {defaultOptions, offlineConfig} from './helpers';
 
-export default function configureServiceStore(preloadedState, appReducer, userOfflineConfig, getAppReducer, enableBuffer = true) {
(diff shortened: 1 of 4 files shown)


jarredwitt:
Reopening to trigger tests.

jwilander:
Something is up with the build server, the redux builds seem to be failing although they pass locally

jwilander:
Alright fixed it, the command tests weren't friendly to being repeated on the same database. If you rebase onto latest master the tests should pass

jarredwitt:
@jwilander done.

src/store/configureStore.dev.js
```diff
@@ -15,11 +15,21 @@ import {General} from 'constants';
 import serviceReducer from 'reducers';
 import deepFreezeAndThrowOnMutation from 'utils/deep_freeze';
 
-import {offlineConfig} from './helpers';
+import {defaultOptions, offlineConfig} from './helpers';
 
-export default function configureServiceStore(preloadedState, appReducer, userOfflineConfig, getAppReducer, enableBuffer = true) {
+export default function configureServiceStore(preloadedState, appReducer, userOfflineConfig, getAppReducer, clientOptions = {}) {
```
jwilander: Can you add a comment here and in the prod file explaining the fields of `clientOptions`? This is arguably the most important function in the library and if the arguments are going to get obscured into an object I think we should document them at least a bit

jarredwitt: Will do

PR #169, opened 2017-06-20
Dispatch received post when create post action commits

jwilander:
#### Summary
Dispatch received post when create post action commits.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6887


jwilander:
That's correct, it'll just make sure the pending post gets overwritten by the actual post from the server

PR #171, opened 2017-06-21
Remove redundant commands selector and test

94117nl:
Remove redundant getCommandsAsArray selector and test. Same result could be achieved by using Object.values in container component.


jwilander:
We weren't using this on the webapp anywhere so should have no backwards compatibility problems

PR #174, opened 2017-06-23
RN-183 Simplified 'user is typing' code to ensure stop events are fired

hmhealey:
Instead of having any separate global state and having to deal with cancelling timeouts, for each typing event it now:

1. Fires a TYPING event with a new timestamp
2. After a delay, fires a STOP_TYPING event with the same timestamp

If the reducer receives a STOP_TYPING with a timestamp >= the stored one, it removes it from the store, but if it receives one with an outdated timestamp, it just ignores it (removing the need to cancel anything).

#### Ticket Link
https://mattermost.atlassian.net/browse/RN-183

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/websocket.js
+++ b/src/actions/websocket.js
@@ -522,41 +522,29 @@ function handleHelloEvent(msg) {
     }
 }
 
-const typingUsers = {};
 function handleUserTypingEvent(msg, dispatch, getState) {
     const state = getState();
     const {currentUserId, profiles, statuses} = state.entities.users;
     const {config} = state.entities.general;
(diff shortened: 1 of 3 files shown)


enahum:
@hmhealey you have some eslint errors

hmhealey:
@enahum There's a blacklist? That sould also be a good idea since I was actually thinking persisting this store might've been another possible cause of this issue

enahum:
@hmhealey oh yes there is https://github.com/mattermost/mattermost-mobile/blob/master/app/store/index.js#L120

PR #180, opened 2017-06-26
Log errors to state and send them to MM

enahum:
Guys take a look at this, before we weren't doing anything with the errors caught by the actions in terms of logging

### DIFF
--- a/src/action_types/errors.js
+++ b/src/action_types/errors.js
@@ -6,5 +6,6 @@ import keyMirror from 'utils/key_mirror';
 export default keyMirror({
     DISMISS_ERROR: null,
     LOG_ERROR: null,
-    CLEAR_ERRORS: null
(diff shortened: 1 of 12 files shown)


jarredwitt:
Is there a way that we can make this configurable? On the shift side we're using Sentry for error reporting and it would be nice to allow clients to pass in or in inject some kind of custom error reporting function that receives the raw error object. Looking at the code maybe we can place the client error handling function check here:

https://github.com/mattermost/mattermost-redux/pull/180/files#diff-72581fece2c9724f8b0810620832f46bR29

That way mattermost reporting is the default, but the client can override that without forking the redux lib and modifying the logError action. Thoughts?

hmhealey:
We could add a couple of settings to the mobile build config like LogErrorsToServer (true/false) and LogErrorsToUrl (a string)

PR #183, opened 2017-06-29
Add action for testElasticsearch.

grundleborg:
#### Summary
Adds TestElasticsearch action.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6474

#### Checklist


src/client/client4.js
```diff
@@ -1489,6 +1489,13 @@ export default class Client4 {
         );
     };
 
+    testElasticsearch = async () => {
```
jwilander: Minor but is Elasticsearch really one word? Looks so odd to me, 0/5 if you want to change it

grundleborg: Yup. It really is one word!

jwilander: Haha alright

PR #189, opened 2017-07-04
Merge webapp-master into master

jwilander:
#### Summary
Merge webapp-master into master.


package.json
```diff
@@ -6,6 +6,7 @@
     "deep-equal": "1.0.1",
     "harmony-reflect": "1.5.1",
     "isomorphic-fetch": "2.2.1",
+    "key-mirror": "1.0.1",
```
enahum: if we are going to use this dependency instead of the `utils` key_mirror, shouldn't we update all the references to it in the constant and action types files?

src/actions/admin.js
```diff
@@ -6,7 +6,9 @@ import {General} from 'constants';
 
 import {Client4} from 'client';
 
-import {bindClientFunc} from './helpers';
+import {getLogErrorAction} from './errors';
```
enahum: this should be changed for `logError` instead of `getLogErrorAction`

src/actions/general.js
```diff
@@ -40,12 +42,33 @@ export function resetPing() {
...
+                {
+                    type: GeneralTypes.CLIENT_CONFIG_FAILURE,
+                    error
+                },
+                logError(error)(dispatch)
+            ]), getState);
+            return null;
+        }
+
+        Client4.setEnableLogging(data.EnableDeveloper === 'true');
```
enahum: this for the alerts rights?

src/actions/teams.js
```diff
@@ -442,8 +442,14 @@ export function joinTeam(inviteId, teamId) {
     return async (dispatch, getState) => {
         dispatch({type: TeamTypes.JOIN_TEAM_REQUEST}, getState);
 
+        const serverVersion = getState().entities.general.serverVersion;
+
         try {
-            await Client.joinTeamFromInvite(inviteId);
+            if (serverVersion.charAt(0) === '3') {
+                await Client.joinTeamFromInvite(inviteId);
```
enahum: 👍

src/client/client4.js
```diff
@@ -184,7 +201,8 @@ export default class Client4 {
         const newOptions = Object.assign({}, options);
 
         const headers = {
-            [HEADER_REQUESTED_WITH]: 'XMLHttpRequest'
+            [HEADER_REQUESTED_WITH]: 'XMLHttpRequest',
+            ...this.defaultHeaders
```
enahum: 👍

src/client/client4.js
```diff
@@ -331,6 +361,12 @@ export default class Client4 {
     }
 
     login = async (loginId, password, token = '', deviceId = '', ldapOnly = false) => {
+        this.trackEvent('api', 'api_users_login');
+
+        if (ldapOnly) {
+            this.trackEvent('api', 'api_users_login_ldap');
+        }
```
enahum: Why are we tracking both events? shouldn't this be one or the other?

src/client/client4.js
```diff
@@ -440,7 +503,7 @@ export default class Client4 {
 
     getMe = async () => {
         return this.doFetch(
-            `${this.getUserRoute('me')}`,
+            `${this.getUserRoute('me')}?text=stuff`,
```
enahum: what is `stuff` here?

src/client/client4.js
```diff
@@ -891,27 +1052,39 @@ export default class Client4 {
 
     // Post Routes
     createPost = async (post) => {
+        this.trackEvent('api', 'api_posts_create', {channel_id: post.channel_id});
+
+        if (post.root_id != null && post.root_id !== '') {
+            this.trackEvent('api', 'api_posts_replied', {channel_id: post.channel_id});
```
enahum: it seems that we are tracking both events, is that ok? well I think you are just bringing this from what is currently in the webapp

src/client/client4.js
```diff
@@ -1095,14 +1302,15 @@ export default class Client4 {
     };
 
     logClientError = async (message, level = 'ERROR') => {
-        const body = {
-            message,
-            level
-        };
+        if (!this.enableLogging) {
```
enahum: hmmm, interesting we are using this in RN to log any error being caught, should we force this in the RN app?

package.json
jwilander: Yep, didn't realize that existed. I'll replace it

src/actions/general.js
jwilander: This is for logging client errors to the server. APIv4 only allows non-admin users to log if developer mode is enabled

src/client/client4.js
jwilander: This is how it's done in the webapp, just copied it over

src/client/client4.js
jwilander: Whoops, that's a mistake from testing

src/client/client4.js
jwilander: Yeah this is what the webapp was doing

src/client/client4.js
jwilander: You're logging the RN errors to the server? It will work if it's in developer mode

enahum: Well not that big of a deal, it won't make it to the server but it will be on the store for when the user reports a problem

src/client/client4.js
```diff
@@ -207,10 +207,10 @@ export default class Client4 {
 
         if (this.token) {
             headers[HEADER_AUTH] = `${HEADER_BEARER} ${this.token}`;
-        } else {
-            newOptions.credentials = 'include';
         }
 
+        newOptions.credentials = 'include';
```
enahum: why?

src/actions/admin.js
enahum: @jwilander you still need to change this here

src/client/client4.js
jwilander: To make sure the cookies are included for the webapp, even if the token is set in the client

package.json
enahum: are we going to keep this?

src/actions/admin.js
jwilander: I know, I haven't even looked at this PR in days. I'll mention you when it's ready for re-review

jwilander:
@enahum ready for review

src/actions/admin.js
```diff
@@ -229,3 +231,67 @@ export function testElasticsearch(config) {
...
+        dispatch({type: AdminTypes.GET_ANALYTICS_REQUEST}, getState);
+
+        let data;
+        try {
+            data = await Client4.getAnalytics(name, teamId);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch);
+            dispatch(batchActions([
+                {type: AdminTypes.GET_ANALYTICS_FAILURE, error},
+                logError(error)
```
enahum: I see everywhere else this as `logError(error)(dispatch)` and I'm thinking that we should either remove it from the batchActions or if it works like this we should make the change everywhere else, thoughts?

src/actions/general.js
```diff
@@ -40,12 +42,33 @@ export function resetPing() {
...
-        GeneralTypes.CLIENT_CONFIG_FAILURE
-    );
+    return async (dispatch, getState) => {
+        dispatch({type: GeneralTypes.CLIENT_CONFIG_REQUEST}, getState);
+
+        let data;
+        try {
+            data = await Client4.getClientConfigOld();
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch);
```
enahum: I think we don't need this here as the config can be requested without being an authorized user, im i right?

jwilander: No, but it doesn't hurt to have anyway. `bindClientFunc` does this anyways right and unless a 401 is received it won't matter

PR #196, opened 2017-07-11
PLT-7092 Fixed getProfilesAndStatusesForPosts when passed null

hmhealey:
When the server returns a `PostList` struct with no posts, it returns null instead of an empty object for posts

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7092

#### Test Information
This PR was tested on: Chrome/Win 10

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -487,6 +487,11 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
 export async function getProfilesAndStatusesForPosts(posts, dispatch, getState) {
+    if (!posts) {
+        // Some API methods return null for no results
+        return Promise.resolve();
+    }
+
     const state = getState();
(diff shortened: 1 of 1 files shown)


hmhealey:
@enahum I'll queue it for the dev meeting. I know there's also a few places where we have the json encoding skip empty fields which could potentially cause the same problem later on

src/actions/posts.js
```diff
@@ -487,6 +487,11 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
 export async function getProfilesAndStatusesForPosts(posts, dispatch, getState) {
+    if (!posts) {
+        // Some API methods return null for no results
+        return Promise.resolve();
```
jwilander: Not a big deal, but can't we just return here and the promise will resolve automatically?

hmhealey: That's how it does work, but ESLint was complaining about an inconsistent return when I did that originally

PR #203, opened 2017-07-19
Include search actions and reducers

enahum:


src/reducers/entities/search.js
```diff
@@ -0,0 +1,83 @@
...
+
+    default:
+        return state;
+    }
+}
+
+export default combineReducers({
+
+    // An ordered array with posts ids from the search results
+    results,
```
jwilander: No action needed, but in the future we could probably expand this to cache results for different search terms

enahum: makes sense ;)

src/actions/search.js
```diff
@@ -0,0 +1,74 @@
...
+import {batchActions} from 'redux-batched-actions';
+
+import {Client4} from 'client';
+import {SearchTypes} from 'action_types';
+
+import {forceLogoutIfNecessary} from './helpers';
+import {logError} from './errors';
+import {getProfilesAndStatusesForPosts} from './posts';
+
+export function searchPosts(teamId, terms, isOrSearch = 'false') {
```
hmhealey: `isOrSearch` should be a boolean value, not a string

src/actions/websocket.js
```diff
@@ -262,23 +263,7 @@ async function handleNewPostEvent(msg, dispatch, getState) {
...
-                getStatusesByIds([rootUserId])(dispatch, getState);
-            }
-
-            dispatch({
-                type: PostTypes.RECEIVED_POSTS,
-                data,
-                channelId: post.channel_id
-            }, getState);
-        });
+        getPostThread(post.root_id);
```
hmhealey: This doesn't work exactly the same as the previous code. It doesn't get the status of offline users, only the statuses of users that it doesn't have already.

src/reducers/entities/posts.js
```diff
@@ -90,6 +90,35 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, action) {
...
 
+function handlePostsFromSearch(posts = {}, postsInChannel = {}, action) {
+    const newPosts = action.data.posts;
+    let info = {posts, postsInChannel};
+    const postsForChannel = new Map();
+
+    for (const id in newPosts) {
+        if (newPosts.hasOwnProperty(id)) {
+            const nextPost = newPosts[id];
+            if (nextPost) {
```
hmhealey: Why are we checking if this is falsey? Shouldn't we know it exists because we just `id` is a key in `newPosts`?

src/reducers/entities/posts.js
```diff
@@ -90,6 +90,35 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, action) {
...
+                    );
+                } else {
+                    postsForChannel.set(channelId, {[id]: nextPost});
+                }
+            }
+        }
+    }
+
+    postsForChannel.forEach((postList, channelId) => {
+        const result = handleReceivedPosts(info.posts, postsInChannel, {channelId, data: {posts: postList}});
```
hmhealey: Can you change this line to just `info = handleReceivedPosts...`? `info` and `result` always have the exact same two fields, so the destructuring to merge them is just replacing info with result

src/reducers/entities/posts.js
```diff
@@ -90,6 +90,35 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, action) {
...
+    let info = {posts, postsInChannel};
+    const postsForChannel = new Map();
+
+    for (const id in newPosts) {
+        if (newPosts.hasOwnProperty(id)) {
+            const nextPost = newPosts[id];
+            if (nextPost) {
+                const channelId = nextPost.channel_id;
+                if (postsForChannel.has(channelId)) {
+                    postsForChannel.set(channelId,
```
hmhealey: Maybe change this to `postsForChannel.get(channelId)[id] = nextPost`? We don't need to worry about mutating the contents of `postsForChannel` since they're just used within this function

src/reducers/entities/posts.js
```diff
@@ -90,6 +90,35 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, action) {
     return {posts: nextPosts, postsInChannel: nextPostsForChannel};
 }
 
+function handlePostsFromSearch(posts = {}, postsInChannel = {}, action) {
+    const newPosts = action.data.posts;
+    let info = {posts, postsInChannel};
+    const postsForChannel = new Map();
+
+    for (const id in newPosts) {
```
hmhealey: This could be `for (const id of Object.keys(newPosts)) {` to avoid the `newPosts.hasOwnProperty` in the next line

src/reducers/entities/search.js
```diff
@@ -0,0 +1,83 @@
...
+
+import {combineReducers} from 'redux';
+import {SearchTypes, UserTypes} from 'action_types';
+
+function handleReceivedSearch(data) {
+    const newPosts = data.posts;
+    const postsForSearch = Object.values(newPosts);
+
+    // Sort to ensure that the most recent posts are first
+    postsForSearch.sort((a, b) => {
```
hmhealey: Shouldn't this have already been done by the server?

src/reducers/entities/search.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+function recent(state = {}, action) {
+    const nextState = {...state};
+    const {data, type} = action;
+    switch (type) {
+    case SearchTypes.RECEIVED_SEARCH_TERM: {
+        const team = {
+            ...nextState[data.teamId] || {},
```
hmhealey: Does this work right with the operator precedence? We use `...(nextState[data.teamId] || {})` elsewhere

src/reducers/entities/search.js
```diff
@@ -0,0 +1,83 @@
...
+        return state;
+    }
+}
+
+function recent(state = {}, action) {
+    const nextState = {...state};
+    const {data, type} = action;
+    switch (type) {
+    case SearchTypes.RECEIVED_SEARCH_TERM: {
+        const team = {
```
hmhealey: Isn't order important here? Why not store these as an array?

Also, should we put something in place to cap the number of search results stored?

src/reducers/entities/search.js
```diff
@@ -0,0 +1,83 @@
...
+            [data.terms]: data.isOrSearch
+        };
+
+        return {
+            ...nextState,
+            [data.teamId]: team
+        };
+    }
+    case SearchTypes.REMOVE_SEARCH_TERM: {
+        const team = {...nextState[data.teamId]} || {};
```
hmhealey: Same comment as above with the order of operations

src/actions/search.js
enahum: it has to be a string, the bool values aren't recognized by the REST API in go, so.. I can pass a boolean and then convert that to a string, should I do that?

src/reducers/entities/search.js
enahum: the order is created when we add the keys to the object so no need for that, now regarding the cap I think that will depend on the client and not defined in the store

src/reducers/entities/search.js
enahum: is the same thing

src/reducers/entities/posts.js
enahum: not really cause this is for every channel meaning that we need to concat each result into info, does it make sense?

enahum:
@hmhealey done

src/actions/search.js
hmhealey: Are you sure? It looks like it pulls that out as a boolean [here](https://github.com/mattermost/platform/blob/master/api4/post.go#L289)

src/reducers/entities/search.js
hmhealey: Javascript [doesn't guarantee key ordering](https://stackoverflow.com/a/5525820), so I don't think we should rely on that.

And for a cap, I'm more meaning a hard cap so that we don't store unbounded amounts of search results here, but this particular case isn't much of a problem since we don't limit that elsewhere in the redux store anyway, and that is storing way more data :p

src/reducers/entities/posts.js
```diff
@@ -95,25 +95,23 @@ function handlePostsFromSearch(posts = {}, postsInChannel = {}, action) {
...
-                    postsForChannel.set(channelId,
-                        {...postsForChannel.get(channelId), [id]: nextPost}
-                    );
-                } else {
-                    postsForChannel.set(channelId, {[id]: nextPost});
-                }
+    const postIds = Object.keys(newPosts);
+    for (const id of postIds) {
+        const nextPost = newPosts[id];
+        if (nextPost) {
```
hmhealey: I'm still not sure why we need this null check here, but I guess it's fine

PR #208, opened 2017-07-27
Getposts optimistic

csduarte:
#### Summary
Adds retry actions for all get posts methods as well as retry attempt tracking.

#### Checklist


src/reducers/requests/posts.js
```diff
@@ -46,6 +46,17 @@ function getPostThread(state = initialRequestState(), action) {
     );
 }
 
+function getPostThreadWithRetryAttempt(state = 0, action) {
+    switch (action.type) {
+    case PostTypes.GET_POST_THREAD_WITH_RETRY_ATTEMPT:
+        return state + 1;
```
jwilander: Just wondering what this is used for?

enahum: @jarredwitt ^

but if I understood correctly is to set the number of retry attempts so far. I'm I right?

jarredwitt: Yep, that's right. We need to track retry attempts client side so that we know if we need to display some kind of UI element if a request is retrying.

jwilander: I just don't see it being used anywhere (not sure if it's going to be used in RN) and it's going to be very racy if you're opening and closing a bunch of different threads rapidly

enahum: actually now that @jwilander mentions that, I wonder the same thing about a race condition

jarredwitt: yeah good catch, probably a good idea to reset the attempt on each request since that will only be dispatched once.

jarredwitt: Or do we even need to be able to retry getting posts for threads?

enahum: yes I would say we should retry getting the thread posts, as you might be replying to a post in the conversation

jarredwitt: Ok I'll add in a reset for the initial request. Those actions also have `cancel:true` in the offline meta which tells the middleware to remove any duplicate actions with the same type from the outbox.

jarredwitt: @enahum @jwilander added reset for when request occurs.

src/store/helpers.js
```diff
@@ -21,6 +21,6 @@ export const offlineConfig = {
             return retries === action.meta.offline.maxRetry;
```
hmhealey: While you're fixing the default return value, do you mind changing this to a `>=`? Using `==` when checking the maximum always gets me a bit worried

jarredwitt: Done.

PR #209, opened 2017-07-27
[PLT-7231] Set language to en when user.locale is empty

saturninoabril:
#### Summary
See report here: https://pre-release.mattermost.com/core/pl/p7nzhnqr9pdf887kwronq3kqzh

The repro steps from the ticket were not reproduced. However, when `user.locale` is empty (for whatever reason), JS errors occurred during login. As a result, user hanged up with loading screen.

screen shot 2017-07-27 at 7 24 32 pm

#### Ticket Link
Jira ticket: [PLT-7231](https://mattermost.atlassian.net/browse/PLT-7231)

#### Checklist

#### Test Information
This PR was tested on: [Mac/Chrome/FF]

### DIFF
--- a/src/utils/channel_utils.js
+++ b/src/utils/channel_utils.js
@@ -22,7 +22,7 @@ export function buildDisplayableChannelList(usersState, allChannels, myPreferenc
     const missingDirectChannels = createMissingDirectChannels(usersState.currentUserId, allChannels, myPreferences);
 
     const {currentUserId, profiles} = usersState;
-    const locale = profiles && profiles[currentUserId] ? profiles[currentUserId].locale : 'en';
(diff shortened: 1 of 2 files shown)


src/utils/channel_utils.js
```diff
@@ -22,7 +22,7 @@ export function buildDisplayableChannelList(usersState, allChannels, myPreferenc
     const missingDirectChannels = createMissingDirectChannels(usersState.currentUserId, allChannels, myPreferences);
 
     const {currentUserId, profiles} = usersState;
-    const locale = profiles && profiles[currentUserId] ? profiles[currentUserId].locale : 'en';
+    const locale = profiles && profiles[currentUserId] && profiles[currentUserId].locale ? profiles[currentUserId].locale : 'en';
```
enahum: looks like this code is used in three different places in this PR, can you make this a function?

PR #211, opened 2017-07-27
Changed getGroupDisplayNameFromUserIds to take the current user

hmhealey:
I need this because my map of profiles doesn't contain the current user in the one case I exported this function for

### DIFF
--- a/src/utils/channel_utils.js
+++ b/src/utils/channel_utils.js
@@ -261,16 +261,16 @@ export function getChannelsIdForTeam(state, teamId) {
     }, []);
 }
 
-export function getGroupDisplayNameFromUserIds(userIds, profiles, currentUserId, teammateNameDisplay) {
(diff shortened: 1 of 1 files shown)


src/utils/channel_utils.js
```diff
@@ -261,16 +261,16 @@ export function getChannelsIdForTeam(state, teamId) {
...
     userIds.forEach((id) => {
-        if (id !== currentUserId) {
+        if (id !== currentUser.id) {
             names.push(displayUsername(profiles[id], teammateNameDisplay));
         }
     });
 
     function sortUsernames(a, b) {
-        const locale = profiles[currentUserId].locale;
+        const locale = currentUser.locale;
```
enahum: Can you use the function added in #209 ?

hmhealey:
Closing this because I realized I do have what I need with the current signature. I can make that change you mentioned though, Elias

hmhealey:
Actually, I can't do that yet since it's part of the webapp-4.0 branch, not master

PR #213, opened 2017-07-28
PLT-6924 Added selectors for mobile file uploads/downloads

hmhealey:
These are for the new settings I'm adding to enable/disable file uploads from the mobile apps. This is needed in the Mattermost Classic app so that's why this isn't in the RN repo

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6924

#### Checklist

### DIFF
--- a/src/selectors/entities/general.js
+++ b/src/selectors/entities/general.js
@@ -1,6 +1,8 @@
 // Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
 // See License.txt for license information.
 
+import {createSelector} from 'reselect';
+
 export function getConfig(state) {
     return state.entities.general.config;
 }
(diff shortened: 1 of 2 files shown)


src/selectors/entities/general.js
```diff
@@ -12,3 +14,19 @@ export function getLicense(state) {
...
+
+export const canUploadFilesOnMobile = createSelector(
+    getConfig,
+    (config) => {
+        // Defaults to true if either setting doesn't exist
+        return config.EnableMobileFileUpload !== 'false' && config.EnableFileAttachments !== 'false';
+    }
+);
+
+export const canDownloadFilesOnMobile = createSelector(
```
enahum: What's the difference between this two selectors?

hmhealey: Different settings. `EnableMobileFileUpload` vs `EnableMobileFileDownload`

PR #222, opened 2017-08-07
Remove Post when error returns for TownSquareIsReadOnly=true

dmeza:
#### Summary
Added a config value to make TownSquare read only for regular users. That means that when TownSquareIsReadOnly=true only admins will be able to post in Town Square. If a basic user posts to town square an error will be displayed: "Town Square is a read-only channel." and the post will be removed.

NOTE: PR for changes to platform: https://github.com/mattermost/platform/pull/7140 

#### Checklist

#### Test Information
Mac OSX El Capitan 10.11.6: Google Chrome 59, Firefox 54, Safari 10.1.2

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -99,8 +99,11 @@ export function createPost(post, files = []) {
                             {type: PostTypes.CREATE_POST_FAILURE, error}
                         ];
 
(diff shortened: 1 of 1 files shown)


enahum:
@hmhealey do we really need https://github.com/mattermost/platform/pull/7140 to be merged before merging this?

PR #224, opened 2017-08-11
PLT-7316 Don't freeze in dev mode on IE11

hmhealey:
Some object didn't have a `hasOwnProperty` method and I think it was some field of a field of a `Set`, but I don't think we care enough about debugging the store being mutated on IE11 to fix it

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7316

### DIFF
--- a/src/utils/deep_freeze.js
+++ b/src/utils/deep_freeze.js
@@ -8,6 +8,10 @@
  *
  */
 
+ const disableOnIe11 =
+    window && window.navigator && window.navigator.userAgent &&
+    window.navigator.userAgent.indexOf('Trident') !== -1;
+
 /**
  * If your application is accepting different values for the same field over
  * time and is doing a diff on them, you can either (1) create a copy or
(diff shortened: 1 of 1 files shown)


src/utils/deep_freeze.js
```diff
@@ -26,6 +26,11 @@
  * only be used in DEV.
  */
 export default function deepFreezeAndThrowOnMutation(object) {
+    // Some objects in IE11 don't have a hasOwnProperty method so don't even bother trying to freeze them
+    if (window.navigator.userAgent.indexOf('Trident') !== -1) {
```
enahum: hmmm, can we test for `window`, `navigator` and `userAgent` don't know if any of those is undefined with RN

hmhealey: Oh yeah, that'll almost certainly break on RN. I'll add a bunch of null checking

enahum:
Build failures

hmhealey:
Hmm, not sure where that extra space came from...

PR #229, opened 2017-08-16
Add theme selectors

jwilander:
#### Summary
Need these for the webapp plugins. Implementation mainly copied from RN.

#### Checklist


src/selectors/entities/preferences.js
```diff
@@ -77,3 +78,55 @@ export const getTeammateNameDisplaySetting = createSelector(
...
+            theme = Preferences.THEMES.default;
+        }
+
+        if (typeof theme === 'string') {
+            try {
+                // A custom theme will be a JSON-serialized object stored in a preference
+                theme = JSON.parse(theme);
+            } catch (e) {
+                // But if it's not a JSON object, it must be the name of a default theme
+                theme = Preferences.THEMES[theme];
```
hmhealey: Can you remove this part and the theme constants from the redux repo? I need to think more about how this'll work since I kind of jammed it in here half-finished, and it won't work if/when we allow servers to change the default themes (there's a Shift PR in platform to allow just that). Maybe it can still have the default theme here for the error case, but the ability to not store just the string key isn't working yet

src/utils/theme_utils.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    const r = parseInt(color.substring(0, 2), 16);
+    const g = parseInt(color.substring(2, 4), 16);
+    const b = parseInt(color.substring(4, 6), 16);
+
+    return 'rgba(' + r + ',' + g + ',' + b + ',' + opacity + ')';
+}
+
+export function concatStyles(...styles) {
```
hmhealey: This function isn't needed here since it only works on RN

src/selectors/entities/preferences.js
jwilander: Yep, will remove all but the default one

jwilander:
@hmhealey updates made

hmhealey:
@enahum Yep, I'll take care of that today

PR #230, opened 2017-08-17
PLT-6377: also filter add / remove messages when join / leave filtering is enabled

ccbrown:
I'm making all add/remove messages hidden when the setting is enabled. The server doesn't send enough information to reliably show only messages for the added / removed user (It doesn't send a user id for them.). This won't have any impact on notifications though since it only filters the display of the messages client-side (The user will still get an email or whatever type of notification they normally get.).

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6377

#### Checklist
N/A

#### Test Information
This PR was tested on: Chrome v60

### DIFF
--- a/src/utils/post_utils.js
+++ b/src/utils/post_utils.js
@@ -119,7 +119,7 @@ export function shouldFilterPost(post, options = {}) {
 
     if (options.filterJoinLeave) {
         const joinLeaveTypes = [postTypes.JOIN_LEAVE, postTypes.JOIN_CHANNEL, postTypes.LEAVE_CHANNEL, postTypes.ADD_REMOVE, postTypes.ADD_TO_CHANNEL, postTypes.REMOVE_FROM_CHANNEL];
-        if (joinLeaveTypes.indexOf(post.type) >= 0) {
+        if (joinLeaveTypes.includes(post.type)) {
             return true;
         }
     }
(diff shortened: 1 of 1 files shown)


src/utils/post_utils.js
```diff
@@ -117,8 +117,11 @@ export function shouldFilterPost(post, options = {}) {
     // Add as much filters as needed here, if you want to filter the post return true
     const postTypes = Posts.POST_TYPES;
 
-    if (options.filterJoinLeave && (post.type === postTypes.JOIN_LEAVE || post.type === postTypes.JOIN_CHANNEL || post.type === postTypes.LEAVE_CHANNEL)) {
-        return true;
+    if (options.filterJoinLeave) {
+        const joinLeaveTypes = [postTypes.JOIN_LEAVE, postTypes.JOIN_CHANNEL, postTypes.LEAVE_CHANNEL, postTypes.ADD_REMOVE, postTypes.ADD_TO_CHANNEL, postTypes.REMOVE_FROM_CHANNEL];
+        if (joinLeaveTypes.indexOf(post.type) >= 0) {
```
enahum: Can we change indexOf with

`if (joinLeaveTypes.includes(post.type))` I think is way easier to read and is more consistent with what we do in other places

ccbrown: Good call. That's definitely better.

PR #232, opened 2017-08-18
Nodejs support

brettmc:
Summary
* add babel-plugin-transform-runtime to allow build to run in node.js
* add websocket to index.js for ease of access
* update readme with example for using the client from node.js

Test Information

This PR was tested on:
* docker node:alpine (node 8.2.1 + 8.4.0)
* docker node:latest (debian/jessie node 8.4.0)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,6 @@
   "plugins": [
     ["module-resolver", {
       "root": ["./src", "."]
-    }],
-    "transform-runtime"
+    }]
   ]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -6,6 +6,7 @@
   "plugins": [
     ["module-resolver", {
       "root": ["./src", "."]
-    }]
+    }],
+    "transform-runtime"
   ]
 }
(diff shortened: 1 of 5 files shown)


brettmc:
FYI the unknown with this PR is whether the use of the `transform-runtime` plugin has any adverse affects downstream (due to the code being transpiled differently). I don't know how to test for that....

.babelrc
```diff
@@ -6,6 +6,7 @@
   "plugins": [
     ["module-resolver", {
       "root": ["./src", "."]
-    }]
+    }],
+    "transform-runtime"
```
enahum: Thia should be taken care from the app that is going to implement and use this library if the sources want to be used

src/client/index.js
```diff
@@ -3,11 +3,13 @@
...
+import WebSocketClient from './websocket_client.js';
 
 const Client = new ClientClass();
 const Client4 = new ClientClass4();
 
 export {
     Client,
-    Client4
+    Client4,
+    WebSocketClient
```
enahum: Why do we want this here?

brettmc:
I got this working with no code changes - `babel-polyfill` was the missing ingredient.

jwilander:
@enahum if it looks good to you LGTM

PR #234, opened 2017-08-21
only set fetch's credentials:include option if no access token

brettmc:
#### Summary
Fixes the following error when making API requests via client4.js from a browser:
```
Fetch API cannot load http://127.0.0.1:8065/api/v4/users/me. Response to preflight request doesn't pass access control
check: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the
request's credentials mode is 'include'. Origin 'http://127.0.0.1:8002' is therefore not allowed access.
```

Per `client.js`, don't set credentials to "include" if a bearer token has been provided.

#### Ticket Link
[None]

#### Test Information
This PR was tested on: Chromium 60 + Firefox 55 (client bundled for browser via webpack)

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -30,6 +30,7 @@ export default class Client4 {
         this.enableLogging = false;
         this.defaultHeaders = {};
         this.userId = '';
+        this.includeCookies = true;
 
         this.translations = {
             connectionError: 'There appears to be a problem with your internet connection.',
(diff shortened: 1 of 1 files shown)


jwilander:
The webapp uses other cookies that need to be included even if the token is included in the header and I'm sure there are other clients doing something similar. I think this might be better as a setting, include cookies/credentials by default and provide the ability to turn them off. Thoughts?

enahum:
Yup I totally agree with @jwilander

brettmc:
Understood. Other options in `client4.js` are all configured via setter functions, how does `setSendHeaderCookies` sound for this case?

jwilander:
How about `setIncludeCookies`?

PR #235, opened 2017-08-21
Ping using v4 if v3 is disabled

enahum:


enahum:
@jwilander why does it look messy? can we make it better? if so happy to hear suggestions ;)

enahum:
We could but then if we get a 404 it might be cause of a wrong url like google.com

I don't mind changing this again once we fully remove APIv3

hmhealey:
Is it really a problem if the app hits `google.ca/api/v4/ping` and then `google.ca/api/v3/ping` in the case of someone entering the wrong URL? And I think we should make sure this is future proof since if removing v3 causes a 404, any app built with this code will just fail to work with a v4-only server. Maybe @jwilander will have some input since he might have an idea how he wants to remove v3

enahum:
Ok then makes sense, I'll move it to use APIv4 as default

enahum:
@hmhealey @jwilander switched to use APIv4 as default

PR #236, opened 2017-08-21
set websocket protocols to empty array instead of null

brettmc:
#### Summary
this fixes an issue in some (all?) browser websocket implementations where null protocol
is cast to the string "null", which results in the following header being sent:
Sec-WebSocket-Protocol:null
The websocket spec says that an empty string should be treated as an empty array, and passing
an empty array seems to make both browser and node's 'ws' websocket implementation happy.

#### Ticket Link
[none]

#### Test Information
This PR was tested on: [node 8.4.0, chromium 60, firefox 55]

### DIFF
--- a/src/client/websocket_client.js
+++ b/src/client/websocket_client.js
@@ -83,7 +83,7 @@ class WebSocketClient {
                 }
             }
 
-            this.conn = new Socket(connectionUrl, null, {origin});
+            this.conn = new Socket(connectionUrl, [], {origin});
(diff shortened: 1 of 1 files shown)


brettmc:
no, hasn't been tested on RN or webapp. I'll try to get a build environment running to see what I can do (but if somebody else is able to do this quickly, I'd appreciate the assistance!)

enahum:
I can test it in about 12 or 13 hours

PR #238, opened 2017-08-23
build for browser

brettmc:
#### Summary
This adds webpack and uglifyjs dev dependencies, webpack config, and a new script, "npm run webpack"
which transpiles the v4 client and websocket client with webpack.
Updated README on how to build and use (including the use of setIncludeCookies(false) to avoid preflight
failure from fetch API).
Lastly, update .gitignore and Makefile to exclude/cleanup new /lib directory.

#### Ticket Link
[None]

#### Test Information
This PR was tested on: [node 8.4.0 (build), chromium 60 + firefox 55 (run)]


brettmc:
sorry, I'll look at those build failures in the next 24 hours...

main.browser.js
```diff
@@ -0,0 +1,7 @@
+require('babel-polyfill');
+const Client4 = require('./client/client4.js').default;
+const client = new Client4;
+const wsClient = require('./client/websocket_client.js').default;
+
+window.MatterMostClient = client;
+window.MatterMostWebSocketClient = wsClient;
```
jwilander: Can you use `Mattermost` instead of `MatterMost`?

main.browser.js
```diff
@@ -0,0 +1,7 @@
+require('babel-polyfill');
+const Client4 = require('./client/client4.js').default;
```
crspeller: Why do this instead of using webpack's library target? Then we can import using import {} from

Makefile
```diff
@@ -37,8 +37,13 @@ pre-run:
...
 	@rm -rf store
 	@rm -rf utils
+	@rm -rf lib
 
 test: check-style
 	npm test
 
 install: .npminstall
+
+bundle:
```
crspeller: Doesn't this require the install target?

main.browser.js
```diff
@@ -0,0 +1,7 @@
+require('babel-polyfill');
+import {Client4} from './client/client4.js';
+import {WebSocketClient} from './client/websocket_client.js';
+const client = new Client4();
+
+window.MattermostClient = client;
```
crspeller: We should just build this as a library instead of loading the clients onto the window. Not sure why this file exists. You would just import the clients into your project.

Makefile
brettmc: I don't think it does. It works, ie `make bundle` does what I expect it would...

brettmc:
@crspeller does this PR require more changes, while it's still somewhat fresh in my mind?

PR #243, opened 2017-08-30
Update README usage to use try catchs

jwilander:
#### Summary
Update README usage to use try catchs

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ const store = configureStore();
 
 If you're only looking to use the v4 JavaScript web client for the Mattermost server:
 
+With async/await:
 ```
 import {Client4} from 'mattermost-redux/client';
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,8 +38,22 @@ import {Client4} from 'mattermost-redux/client';
 Client4.setUrl('https://your-mattermost-url.com');
 
 async function loginAndGetUser(username, password) {
-    await Client4.login(username, password);
-    return await Client4.getMe();
+    try {
+        await Client4.login(username, password);
+    } catch (error) {
+        console.error(error);
+        return null;
```
enahum: This should be returning {error}

README.md
```diff
@@ -38,8 +38,22 @@ import {Client4} from 'mattermost-redux/client';
...
+
+    let user;
+    try {
+        user = await Client4.getMe();
+    } catch (error) {
+        console.error(error);
+        return null;
+    }
+
+    return user;
```
enahum: And this should be {data: user} as we talked

jwilander: These aren't actions?

enahum: Oh is just using the client...

Can you add another example using promises instead of async/await?

jwilander: Yeah sure

PR #245, opened 2017-09-01
PLT-7498 Fix reply threads not showing up unless loaded in to center channel

jwilander:
#### Summary
Does two things:
* Check all posts in thread selector (also created a utils function for sorting to support this)
* Check all posts when counting replies in post selector

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7498

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -3,7 +3,7 @@
 
 import {PostTypes, SearchTypes, UserTypes} from 'action_types';
 import {Posts} from 'constants';
-import {isPostPendingOrFailed} from 'utils/post_utils';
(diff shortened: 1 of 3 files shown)


jwilander:
I'm not too worried about having to go through all the posts instead of just those in the channel because multiple other selectors were already doing that.

However, I think we need to add a reducer to start tracking `postsInThread` so that we don't have to loop through all the posts trying to find replies. I didn't do that here because that would be a lot of work close to a release but I think it's worthwhile

enahum:
@jwilander build failures
```
home/ubuntu/workspace/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux/src/reducers/entities/posts.js
  6:9  error  'isPostPendingOrFailed' is defined but never used  no-unused-vars

```

PR #253, opened 2017-09-18
PLT-7544 Add various actions, selectors and utils to support channel header

jwilander:
#### Summary
Add various actions, selectors and utils to support channel header.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7544

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -6,7 +6,7 @@ import {batchActions} from 'redux-batched-actions';
 import {Client4} from 'client';
 import {General, Preferences} from 'constants';
 import {ChannelTypes, PreferenceTypes, TeamTypes, UserTypes} from 'action_types';
-import {savePreferences} from 'actions/preferences';
+import {savePreferences, deletePreferences} from 'actions/preferences';
 import {getChannelsIdForTeam} from 'utils/channel_utils';
 
 import {logError} from './errors';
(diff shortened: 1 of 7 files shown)


src/actions/channels.js
```diff
@@ -1060,6 +1060,37 @@ export function getMyChannelMember(channelId) {
     );
 }
 
+export function favoriteChannel(channelId) {
```
enahum: we'll be moving the actions that we have in RN to use these

src/selectors/entities/users.js
```diff
@@ -177,14 +177,18 @@ export const getProfileSetNotInCurrentTeam = createSelector(
     }
 );
 
-function sortAndInjectProfiles(profiles, profileSet) {
+function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
```
enahum: what is this used for? should we do anything in RN for this?

jwilander: To skip inactive users, the default is the previous behaviour so you shouldn't need to change anything

enahum: Hmm, I wonder if we should turn this around as I don't really see any reasons to show inactive users other than in the system console

jwilander: Maybe, but that would need to be done because that would require more changes in the webapp that I don't have time to do now

PR #254, opened 2017-09-19
Add channel_updated websocket action

enahum:
#### Summary
Handle channel updated WebSocket events

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7677

#### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 
 	@echo Getting dependencies using npm
 
-	npm install --ignore-scripts
+	npm install --ignore-scripts --no-package-lock
 
 	touch $@
 
@@ -22,7 +22,7 @@ check-style: | pre-run .npminstall
 clean:
 	@echo Cleaning app
 
-	npm cache clean
+	npm cache clean --force
 	rm -rf node_modules
 	rm -f .npminstall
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ check-style: | pre-run .npminstall
 clean:
 	@echo Cleaning app
 
-	npm cache clean
+	npm cache clean --force
```
jwilander: What are the makefile changes needed for?

src/actions/websocket.js
```diff
@@ -457,6 +460,28 @@ function handleChannelDeletedEvent(msg, dispatch, getState) {
...
+
+            if (currentChannelId === channel.id) {
+                // Emit an event with the channel received as we need to handle
+                // the changes without listening to the store
+                EventEmitter.emit(WebsocketEvents.CHANNEL_UPDATED, channel);
+            }
+        }
+    } catch (err) {
+        // do nothing
+    }
```
jwilander: Is the try catch just for the JSON parsing? If so, maybe just put it around that and return in the catch

Makefile
enahum: npm 5 complains, that's why I added it

src/actions/websocket.js
enahum: will do

PR #256, opened 2017-09-21
GH-7236: Add history actions/selectors/reducers

jespino:
Migration of the flux message history store from the webapp into redux


src/reducers/entities/posts.js
```diff
@@ -314,6 +314,65 @@ function openGraph(state = {}, action) {
     }
 }
 
+function messagesHistory(state = {}, action) {
+    const index = {};
+    index[Posts.MESSAGE_TYPES.POST] = -1;
+    index[Posts.MESSAGE_TYPES.COMMENT] = -1;
+
+    const nextIndex = state.index ? {...state.index} : index;
+    let nextMessages = state.messages ? [...state.messages] : [];
```
jwilander: Just a small note as I don't think this is an issue in this case, but every time any action dispatches this code will run even if the reducer doesn't need to do anything. It's best to keep the setup outside the switch statement to a minimum

I don't think it's a problem here because the setup should have negligible effect but just something to keep in mind

src/selectors/entities/posts.js
```diff
@@ -263,3 +263,17 @@ export const getSearchResults = createSelector(
         return postIds.map((id) => posts[id]);
     }
 );
+
+export function makeGetMessageInHistoryItem(index) {
+    return createSelector(
+      (state) => state.entities.posts.messagesHistory,
+      (messagesHistory) => {
```
jwilander: Can you give me a bit of context on how this is to be used? Is it common that a selector will be created for a specific index and used over and over? I don't think a selector makes much sense when the index being asked for is going to change a lot

jespino: Sorry, probably i have to change the parameter name, is not really the index, is the type.

The idea of usage is create a selector for a specific type ("post" or "comment") and always select the current selectect history entry.

src/reducers/entities/posts.js
jespino: The approach here was to reduce the quantity of repeated code, but I can change it.

jwilander: No change required just wanted to give a heads-up

src/selectors/entities/posts.js
jwilander: Ahh ok, that makes more sense

jespino:
Ready!

Other thing, this is only part of the solution to the mattermost/mattermost-server#7236 issue, i have to migrate browser storage to redux too, and then, migrate the component to use the redux approach.

enahum:
@jespino looks good but before I approve can you please rebase the PR?

jespino:
Sure!

jespino:
Rebased!

enahum:
@jespino I think you have a bad rebase

/home/ubuntu/workspace/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux/test/selectors/posts.test.js
  11:1  error  'constants' import is duplicated  no-duplicate-imports
  11:9  error  'Posts' is already defined        no-redeclare

jespino:
Yes, Git didn't give me a conflict in this part. Now fixed!

enahum:
Thanks @jespino

PR #258, opened 2017-09-25
[PLT-7703] Add/distinguish telemetry for api channels add/join/add/remove member events

saturninoabril:
#### Summary
Add/distinguish telemetry for api channels events
- `api_channels_leave`
- `api_channels_join`
- `api_channels_add_member`
- `api_channels_remove_member`

#### Ticket Link
Jira ticket: [PLT-7703](https://mattermost.atlassian.net/browse/PLT-7703)

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -512,6 +512,8 @@ export function leaveChannel(channelId) {
         const channel = channels[channelId];
         const member = myMembers[channelId];
 
+        Client4.trackEvent('api', 'api_channels_leave', {channel_id: channelId});
+
         dispatch({
             type: ChannelTypes.LEAVE_CHANNEL,
             data: {
@@ -564,6 +566,8 @@ export function joinChannel(userId, teamId, channelId, channelName) {
             ]), getState);
             return {error};
         }
+
+        Client4.trackEvent('api', 'api_channels_join', {channel_id: channelId});
 
         dispatch(batchActions([
             {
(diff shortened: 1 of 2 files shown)


src/actions/channels.js
```diff
@@ -512,6 +512,8 @@ export function leaveChannel(channelId) {
         const channel = channels[channelId];
         const member = myMembers[channelId];
 
+        Client4.trackEvent('api', 'api_channels_leave', {channel_id: channelId});
```
jwilander: These should use the type `action` and the names should be prefixed with `action_` instead of `api_`

Like so:
https://github.com/mattermost/mattermost-redux/blob/master/src/actions/posts.js#L334

src/client/client4.js
```diff
@@ -1071,8 +1069,6 @@ export default class Client4 {
     };
 
     removeFromChannel = async (userId, channelId) => {
-        this.trackEvent('api', 'api_channels_remove_member', {channel_id: channelId});
```
jwilander: I don't think we should remove these, as this is tracking API usage and the new ones are tracking specific actions

jwilander:
@enahum this is for adding extra tracking when actions don't align with the APIs directly but we should mark them properly and not remove the API tracking

enahum:
Ok that is different then, I’m 100% ok with that approach

saturninoabril:
Change type to `action` and modify prefix into:
- action_channels_leave
- action_channels_join
- action_channels_add_member
- action_channels_remove_member

Thanks all!

PR #268, opened 2017-10-16
[WIP][Gh 7237] migrate search store

Zapix:
#### Summary
Move `SearchStore` to redux search reducer

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/7489

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -910,7 +910,7 @@ export function getFlaggedPosts() {
                     }
                 },
                 {
-                    type: SearchTypes.RECEIVED_SEARCH,
+                    type: SearchTypes.RECEIVED_FLAGGED_POSTS,
                     data: {
                         results: data,
                         is_flagged_posts: true,
@@ -951,11 +951,14 @@ export function getPinnedPosts(channelId) {
                     }
                 },
                 {
-                    type: SearchTypes.RECEIVED_SEARCH,
+                    data,
+                    channelId: requestedChannelId,
+                    type: PostTypes.RECEIVED_POSTS
+                },
+                {
+                    type: SearchTypes.RECEIVED_PINNED_POSTS,
                     data: {
-                        results: data,
-                        is_flagged_posts: false,
-                        is_pinned_posts: true,
+                        ...data,
                         channelId: requestedChannelId,
                     }
                 }
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -910,12 +910,12 @@ export function getFlaggedPosts() {
                     }
                 },
                 {
-                    type: SearchTypes.RECEIVED_FLAGGED_POSTS,
-                    data: {
-                        results: data,
-                        is_flagged_posts: true,
-                        is_mention_search: false,
-                    }
+                    data,
+                    type: SearchTypes.RECEIVED_FLAGGED_POSTS
+                },
+                {
+                    data,
+                    type: SearchTypes.RECEIVED_SEARCH_POSTS
                 }
             ]));
 
@@ -929,7 +929,7 @@ export function getFlaggedPosts() {
             ]), getState);
             return null;
         }
-    }
+    };
 }
 
 export function getPinnedPosts(channelId) {
@@ -959,7 +959,7 @@ export function getPinnedPosts(channelId) {
                     type: SearchTypes.RECEIVED_PINNED_POSTS,
                     data: {
                         ...data,
-                        channelId: requestedChannelId,
+                        channelId: requestedChannelId
                     }
                 }
             ]));
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -5,8 +5,9 @@ import {batchActions} from 'redux-batched-actions';
 
 import {Client4} from 'client';
 import {General, Preferences, Posts} from 'constants';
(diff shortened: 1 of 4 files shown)


src/reducers/entities/search.js
```diff
@@ -61,12 +61,122 @@ function recent(state = {}, action) {
...
+ * Structure is:
+ * {
+ *     term: string on null - searching terms,
+ *     searchResults: array or null - list of searched posts
+ *     isMentionSearch: boolean - is searched be mentions
+ *     isFlaggedPosts: boolean - is searched flagged posts
+ *     isPinnedPosts: boolean - is searched boolean posts
+ * }
+ */
+function search(
```
jwilander: Getting there but we should split out the flagged posts and pinned posts into their own reducers and instead of adding a new search reducer, we'd have:

* `results` - This already exists and we should re-use it for storing the results of a regular search or a mention search. No need to differentiate between the two here since they're pretty much the same. We can track the difference in the webapp reducer if we need to
* `flaggedPosts` - This will be new and just store the post ids of the flagged posts, the actual posts should be stored in the `posts` reducer. E.g. `Set[postid1, postid2, etc...]`
* `pinnedPosts` - This will be new and store the pinned post ids for each channel E.g. `{channelid1: Set[postid1, postid2, etc...], channelid2: Set[], etc...}`

The `isMentionSearch`, `isFlaggedPosts`, `isPinnedPosts` and `term` fields don't need to be tracked in mattermost-redux because they're all controlling view state. The reducer that gets added to the webapp should take care of that view state (preferably using an enum var instead of individual booleans). The idea here is that post searches, flagged posts and pinned posts aren't really related so we shouldn't make them so in the reducers. We should be able to store them all in parallel without affecting the other and then the client (the webapp in our case) can choose how to show them

Zapix: @jwilander Thank you for your feedback. Will update PR soon

jwilander:
Hey @Zapix, just checking how it's going and if you need any help with this?

Zapix:
@jwilander Sorry. for to late response. I'll continue work under pr in weekend

Zapix:
@jwilander Updated pr. Please check. Am I in right way? If yes. I'll write tests and sync with master branch.

jwilander:
@Zapix sorry, I must have missed your last message. I'm taking a look now

src/actions/posts.js
```diff
@@ -888,6 +890,93 @@ export function getOpenGraphMetadata(url) {
     };
 }
 
+export function getFlaggedPosts() {
+    return async (dispatch, getState) => {
+        const state = getState();
+        const userId = getCurrentUserId(state);
+        const teamId = getCurrentTeamId(state);
+
+        dispatch({type: SearchTypes.SEARCH_POSTS_REQUEST});
```
jwilander: Can we create a request type just for getting flagged posts?

src/actions/posts.js
```diff
@@ -888,6 +890,93 @@ export function getOpenGraphMetadata(url) {
...
+                    type: SearchTypes.RECEIVED_FLAGGED_POSTS
+                },
+                {
+                    data,
+                    type: SearchTypes.RECEIVED_SEARCH_POSTS
+                }
+            ]));
+
+            getProfilesAndStatusesForPosts(data.posts, dispatch, getState);
+            return data;
```
jwilander: Looks good, but can we move the dispatch, other actions and return out of the try block so it matches other actions?

Also we switched to using `return {data}` and `return {error}` [per this post](https://pre-release.mattermost.com/core/pl/ap1ue3xtobb6uxssdypq1rudyr), so if you could update the returns for this action that'd be great

src/actions/posts.js
```diff
@@ -888,6 +890,93 @@ export function getOpenGraphMetadata(url) {
...
+            dispatch(batchActions([
+                {type: SearchTypes.SEARCH_POSTS_FAILURE},
+                logError(error)(dispatch)
+            ]), getState);
+            return null;
+        }
+    };
+}
+
+export function getPinnedPosts(channelId) {
```
jwilander: Can we do the same for this action as my comments for the `getFlaggedPosts` one?

src/actions/posts.js
```diff
@@ -888,6 +890,93 @@ export function getOpenGraphMetadata(url) {
...
+                        is_mention_search: false
+                    }
+                },
+                {
+                    data,
+                    type: SearchTypes.RECEIVED_FLAGGED_POSTS
+                },
+                {
+                    data,
+                    type: SearchTypes.RECEIVED_SEARCH_POSTS
```
jwilander: I don't think we need this `RECEIVED_SEARCH_POSTS` action dispatched here, since we want to separate the search results for posts and the getting flagged posts

Zapix:
@jwilander Thank you for your review. Will fix as soon as possible.

jwilander:
@Zapix let me know if you're interested in continuing work on this and if so, please let me know if you need any help

Zapix:
@jwilander Hi. I'll continue work on this PR on weekends

jasonblais:
Hey @Zapix, let us know if you need any help with this PR. We really appreciate your help

jwilander:
Closing due to inactivity, feel free to re-open if warranted

PR #269, opened 2017-10-16
Selectors to improve performance for listing channels

enahum:
#### Summary
In order to increase the performance unnecessary re-renders I've created other types of selectors that the regular `createSelector` those are:
* `createIdsSelector` this on is used to shallow compare all the arguments and memoize the result, try and use it only when the selector returns an array of ids.
* `createShallowSelector` shallow compare all the arguments but does not memoize the result.
* `createDeepSelector` use it when you need to deep compare all the arguments, this one won't memoize the result.

Also I've created a few selectors that make use of the above creators to get sorted team Ids, different type of channels, mentions count, teams count, etc..

#### Ticket Link
https://mattermost.atlassian.net/browse/RN-415

#### Checklist

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -4,16 +4,26 @@
 import {createSelector} from 'reselect';
 
 import {getConfig, getLicense} from 'selectors/entities/general';
-import {getMyPreferences, getTeammateNameDisplaySetting} from 'selectors/entities/preferences';
+import {
+    getFavoritesPreferences,
+    getMyPreferences,
+    getTeammateNameDisplaySetting,
+    getVisibleTeammate,
+    getVisibleGroupIds
+} from 'selectors/entities/preferences';
 import {getCurrentTeamId, getCurrentTeamMembership} from 'selectors/entities/teams';
-import {getCurrentUser} from 'selectors/entities/users';
+import {getCurrentUser, getUsers} from 'selectors/entities/users';
 import {
     buildDisplayableChannelList,
     buildDisplayableChannelListWithUnreadSection,
     canManageMembers,
     completeDirectChannelInfo,
-    sortChannelsByDisplayName
+    completeDirectChannelDisplayName,
+    sortChannelsByDisplayName,
+    getDirectChannelName
 } from 'utils/channel_utils';
+import {createIdsSelector} from 'utils/helpers';
+
 import {General} from 'constants';
 
 export function getAllChannels(state) {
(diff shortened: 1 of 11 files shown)


src/selectors/entities/preferences.js
```diff
@@ -98,7 +120,7 @@ const getThemePreference = createSelector(
     }
 );
 
-export const getTheme = createSelector(
+export const getTheme = createShallowSelector(
```
hmhealey: This doesn't need a shallow selector since the only way for a preference to change right now is if it's value changes or if a team-specific theme is added/removed

src/selectors/entities/teams.js
```diff
@@ -42,6 +43,14 @@ export const getCurrentTeam = createSelector(
     }
 );
 
+export const getTeam = createSelector(
```
hmhealey: This doesn't need to be a selector since it's just returning a part of the store

src/selectors/entities/teams.js
```diff
@@ -124,3 +133,84 @@ export const getJoinableTeams = createSelector(
...
+export const getMyTeamsCount = createSelector(
+    getTeamMemberships,
+    (teams) => {
+        return Object.keys(teams).length;
+    }
+);
+
+// Returns the number of mentions or -1 if it has unreads
+// for every team except the current one
+export const getTeamsMentions = createSelector(
```
hmhealey: Why not return the current one? It might be useful in the future if we change the UI, and the UI can handle not showing for the current channel easily enough

hmhealey: Oh I see, it's not returning a map of team ids to mention counts. That name doesn't make it at all clear what this does, but I can't come up with a better one right now. Maybe something super obvious like `getChannelDrawerBadgeCount`?

I don't know how I feel about this being here then since it's pretty specific to the channel drawer button, although I guess it'll get used eventually for the mobile web view, so I guess I'm 0/5 about having it here.

src/selectors/entities/teams.js
```diff
@@ -124,3 +133,84 @@ export const getJoinableTeams = createSelector(
...
+            badgeCount = -1;
+        }
+
+        return badgeCount;
+    }
+);
+
+// Returns the number of mentions or -1 if it has unreads
+// for every team except the current one
+export const getTeamMentions = createSelector(
```
hmhealey: This should be done as a selector factory (ie `makeGetTeamMentions`) or else the memoization does nothing since it'll constantly be called with different ids

hmhealey: Similarly, the name is a bit unclear. Maybe `makeGetBadgeCountForTeamId` to make its use clear and point to the fact that you need to pass in a team id?

src/utils/helpers.js
```diff
@@ -0,0 +1,35 @@
...
+        return lastResult;
+    };
+}
+
+// Use this selector when you want a shallow comparison of the arguments and you want to memoize the result
+// try and use this only when your selector returns an array of ids
+export const createIdsSelector = createSelectorCreator(memoizeResult);
+
+// Use this selector when you want a shallow comparison of the arguments and you don't need to memoize the result
+export const createShallowSelector = createSelectorCreator(defaultMemoize, shallowEqual);
```
hmhealey: Do we have a good use case yet for using createShallowSelector and createDeepSelector? I'd prefer not to add them yet until we know we need them since we otherwise might start using them when it's not necessary since they are more expensive

src/selectors/entities/preferences.js
enahum: Well actually the whole thing changes whenever the websocket re-connects as it replaces the whole `myPreferences` object

src/selectors/entities/teams.js
enahum: yup, I'll remove the selector

src/selectors/entities/teams.js
enahum: Will update the name

src/selectors/entities/teams.js
enahum: I'll create the factory and update the name here

src/utils/helpers.js
enahum: The ShallowSelector is being applied to the `getTheme` selector, I can remove the other one

enahum:
@hmhealey changes made

src/selectors/entities/preferences.js
hmhealey: Ah fair enough

src/selectors/entities/channels.js
```diff
@@ -321,3 +331,185 @@ export const canManageChannelMembers = createSelector(
...
+);
+
+export const getSortedUnreadChannelIds = createIdsSelector(
+    getCurrentUser,
+    getUsers,
+    getAllChannels,
+    getMyChannelMemberships,
+    getUnreadChannelIds,
+    getTeammateNameDisplaySetting,
+    (currentUser, profiles, channels, members, unreadIds, settings) => {
```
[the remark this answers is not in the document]
enahum: this will perform a shallowEquality with the arguments, if one of them is different it will re-run the function but if the result of the function is shallow equal to the last result then it will return the memoized result so it is strict equal

src/selectors/entities/teams.js
```diff
@@ -124,3 +130,86 @@ export const getJoinableTeams = createSelector(
...
+export const getMyTeamsCount = createSelector(
+    getTeamMemberships,
+    (teams) => {
+        return Object.keys(teams).length;
+    }
+);
+
+// Returns the number of mentions or -1 if it has unreads
+// for every team except the current one
+export const getChannelDrawerBadgeCount = createSelector(
```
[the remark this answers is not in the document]
enahum: yes, I'll make the comment more clear

enahum:
@jwilander fixed the comments

src/selectors/entities/channels.js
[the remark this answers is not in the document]
jwilander: Ok, so the savings isn't in not re-running the function it's in returning the memoized result so the components don't re-render if the result didn't change. Got it, thanks.

PR #271, opened 2017-10-17
Update location of mattermost-server config file in README

fraziern:


jwilander:
It should actually still be in config/config.json. If there is no config/config.json then the server just falls back to default.json. So I guess the updated steps would be to copy default.json to config.json if it doesn't exist and then update the values

fraziern:
Sounds good. Added some edits.

PR #275, opened 2017-10-18
Added better selectors to get lists of post IDs instead of posts

hmhealey:
These selectors now
1. Return post IDs instead of posts so that equality checking is easier for calling code
2. Are memoized based on results as well as inputs, so a selector will always return the same array whenever possible instead of sometimes returning a different, but shallowly equal array

I removed addDatesToPostList since it was only used by the mobile app, and it was no longer necessary. I left most of the old selectors around since they're still in use by the web app.

#### Checklist

#### Test Information


enahum:
I think you need to remove the tests that make use of `addDatesToPostList`

hmhealey:
@enahum Right, that's fixed now

PR #279, opened 2017-10-19
Added getPostsForIds selector

hmhealey:
Okay, this should be the final selector for the post changes. It's used to reduce recalculation when working with a subset of posts that we're not recalculating when unrelated posts are changed. For example, if we have a list of post IDs in a thread and we want to perform some action on them (eg filter them), we get only the posts in that thread so that we avoid recalculation when a post in another channel changes.

#### Checklist

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -201,11 +201,11 @@ export function makeGetPostsInChannel() {
             const posts = [];
 
             const joinLeavePref = myPreferences[getPreferenceKey(Preferences.CATEGORY_ADVANCED_SETTINGS, 'join_leave')];
-            const filterJoinLeave = joinLeavePref ? joinLeavePref.value === 'false' : false;
+            const showJoinLeave = joinLeavePref ? joinLeavePref.value !== 'false' : true;
 
             for (let i = 0; i < postIds.length; i++) {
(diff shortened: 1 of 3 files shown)


hmhealey:
Sorry, I know this has been reviewed already, but I snuck in a change to the shouldFilterPost helper function to change the "filterJoinLeave" option into "showJoinLeave". I spent an embarrassing amount of time trying to figure out how the preference for that is supposed to work (true or unset is to show the posts, false to hide them) and then that function takes its parameters completely differently (false or unset to show the posts, false to hide them), so I wanted to get that changed because I just wasn't able to get my unit tests working and not break that feature with the way it worked.

jwilander:
Got a few eslint errors:

```
/home/ubuntu/workspace/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux/test/selectors/posts.test.js
  1358:19  error  'posts' is already declared in the upper scope  no-shadow
  1385:19  error  'posts' is already declared in the upper scope  no-shadow
  1453:42  error  Unnecessarily quoted property 'message' found   quote-props

✖ 3 problems (3 errors, 0 warnings)
```

hmhealey:
Fixed the style errors and re-enabled my git hook that usually checks those

PR #281, opened 2017-10-20
PLT-7819 Split RECEIVED_POST into RECEIVED_NEW_POST and RECEIVED_POST

jwilander:
#### Summary
This allows us to add a post to the posts entity without affecting the postsInChannel. This fixes a bug previously fixed in the webapp where another user editing a post that your client doesn't have would cause it to appear at the bottom of the channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-7819

#### Test Information
This PR was tested on: webapp


jwilander:
Nope, the behaviour should be the same unless you are manually dispatching POST_RECEIVED actions anywhere where you want it to add to postsInChannel

PR #282, opened 2017-10-23
WIP: [PLT-7364] Ability to Direct Message yourself

cpanato:
I need some help to test and find if this is good in terms of code.
@jasonblais can help me here?

Server PR: https://github.com/mattermost/mattermost-server/pull/7696
Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/186

#### Summary
 Ability to Direct Message yourself

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/PLT-7364
GitHub: https://github.com/mattermost/mattermost-server/issues/7618

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -232,7 +232,7 @@ export function logout() {
     );
 }
 
-export function getProfiles(page = 0, perPage = General.PROFILE_CHUNK_SIZE) {
+export function getProfiles(page = 0, perPage = General.PROFILE_CHUNK_SIZE, skipCurrent = true) {
     return async (dispatch, getState) => {
         dispatch({type: UserTypes.PROFILES_REQUEST}, getState);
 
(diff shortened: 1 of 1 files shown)


src/actions/users.js
```diff
@@ -241,7 +241,9 @@ export function getProfiles(page = 0, perPage = General.PROFILE_CHUNK_SIZE) {
         let profiles;
         try {
             profiles = await Client4.getProfiles(page, perPage);
-            removeUserFromList(currentUserId, profiles);
+            if (skipCurrent) {
```
enahum: why is this conditional?

The problem by not skipping the current user all the time is that you will replace the currentUser with whatever getProfiles returns and then cause of the profile being sanitize you'll remove the `notifiy_props` needed by the currentUser

src/actions/users.js
```diff
@@ -241,7 +241,9 @@ export function getProfiles(page = 0, perPage = General.PROFILE_CHUNK_SIZE) {
...
+                removeUserFromList(currentUserId, profiles);
+            }
```
jwilander: Looking at your webapp code, is this change really needed? The redux store should already have the current user and you're not using the direct action result anywhere you're just selectors to pull from the store, which is guaranteed to already have the current user.

Let me know if I'm missing something

cpanato: tested, dont need this change here, closing...

src/actions/users.js
cpanato: tested, dont need this change here, closing...

cpanato:
tested, dont need this change here, closing...

PR #283, opened 2017-10-23
Fix crash caused by selectors

enahum:


src/selectors/entities/posts.js
```diff
@@ -313,6 +313,10 @@ export const getSearchResults = createSelector(
     getAllPosts,
     (state) => state.entities.search.results,
     (posts, postIds) => {
+        if (!postIds) {
+            return null;
```
jwilander: Should this an empty array?

enahum:
@jwilander updated to return an empty array

PR #286, opened 2017-10-27
[PLT-7362] Add post root ID to client4 addToChannel to render added user (as system post) at RHS

saturninoabril:
#### Summary
- Add post' root ID to APIv4 `addChannelMember` to render added user (as system post) at RHS. Feature is requested while testing `webapp` https://github.com/mattermost/mattermost-webapp/pull/149#pullrequestreview-72349879

Note: Only client4 was updated.

#### Ticket Link
Jira ticket: [PLT-7362](https://mattermost.atlassian.net/browse/PLT-7362)

#### Checklist
- [x] Corresponding server side change (https://github.com/mattermost/mattermost-server/pull/7730)
- [x] Corresponding client side change (https://github.com/mattermost/mattermost-webapp/pull/149)

#### Test Information
This PR was tested on: Chrome/FF(s), Mac OS Sierra]

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -1081,10 +1081,10 @@ export default class Client4 {
         );
     };
 
-    addToChannel = async (userId, channelId) => {
+    addToChannel = async (userId, channelId, postRootId) => {
         this.trackEvent('api', 'api_channels_add_member', {channel_id: channelId});
 
-        const member = {user_id: userId, channel_id: channelId};
+        const member = {user_id: userId, channel_id: channelId, post_root_id: postRootId};
         return this.doFetch(
             `${this.getChannelMembersRoute(channelId)}`,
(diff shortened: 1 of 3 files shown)


jwilander:
Per my comments on the server PR, I'm not sure this is needed

jasonblais:
@saturninoabril Looks like the build is failing

saturninoabril:
test updated, should be ready for review...

src/client/client4.js
```diff
@@ -1081,10 +1081,10 @@ export default class Client4 {
         );
     };
 
-    addToChannel = async (userId, channelId) => {
+    addToChannel = async (userId, channelId, postRootId) => {
```
jwilander: Minor but can we default this to `''`? e.g. `... postRootId = '')` That way if it's not used we're not sending null value to the server

PR #291, opened 2017-11-01
Added dev and devwatch scripts to easy the webapp+redux development

jespino:


package.json
```diff
@@ -33,6 +33,7 @@
     "form-data": "2.3.1",
     "jsdom": "9.12.0",
     "jsdom-global": "2.1.1",
+    "jquery": "3.2.1",
```
enahum: Why would you need jquery here? can we please remove it

package.json
```diff
@@ -43,6 +44,8 @@
   },
   "scripts": {
     "build": "babel src --out-dir .",
+    "dev": "babel src --out-dir ../mattermost-webapp/node_modules/mattermost-redux",
+    "dev:watch": "babel --watch src --out-dir ../mattermost-webapp/node_modules/mattermost-redux",
```
jwilander: Is it possible to have the location of the webapp repo configurable via an environment variable?

package.json
jespino: Sorry, fixed, was part of an afternoon of a lot of tries with different approaches.

jespino:
@enahum added the README paragraph about compiling.

PR #292, opened 2017-11-01
Migration of browserStore to redux

jespino:


jespino:
The login/logout actions is better managed as specific actions with special side effects, because, the objective and the behavior is completely different. This actions are for communication between browsers, and not for store data. Here is where the important logic of login/logout actions really happens: https://github.com/mattermost/mattermost-webapp/blob/master/components/root.jsx#L59

jespino:
I think is ready for get comments (probably the setLandingPageSeen related stuff can be leaved in mattermost-webapp). The tests are failing, but not my tests are other tests unrelated and failing only sometimes.

enahum:
I’ll take a closer look to this on monday but seems to me that everything related to the persistance of data should be done in the webapp in a similar way as it is done in the mobile apps as it can be very specific to each client

jespino:
Hi @enahum, I'm not agree, the data stored in the localStorage is part of the global data of the application, and the read/write actions are side-effects. I think that encapsulating it behind the flux pattern is a good approach (at least for the webapp).

The set of actions/selectors are really generic to any storage (SET_ITEM/GET_ITEM/REMOVE_ITEM...), so I understand is interesting for any application.

enahum:
@jespino After looking at this PR I'm totally convinced this should be done in redux but in the webapp and not with this redux library

The data stored in localStore as you say is only needed for the webapp, I don't really see the need for this if not in the webapp along with all the read/write actions and the selectors

jespino:
Ok, I hadn't noticed mattermost-webapp has its own actions/reducers/selectors set. So I agree. I going to migrate this to the mattermost-webapp redux part. Thanks for your feedback :smiley:

PR #298, opened 2017-11-03
Fix .npminstall Makefile task in linux

jespino:
The command "command", doesn't exist in Linux (at least in Arch Linux), and
which works here for both operating systems. Anyway, this check is needed?
mattermost-webapp haven't this check, I'm 5/5 about it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 
 .npminstall: package.json
-	@if ! [ $(shell command -v npm) ]; then \
+	@if ! [ $(shell which npm) ]; then \
 		echo "npm is not installed"; \
 		exit 1; \
 	fi


Makefile
```diff
@@ -2,7 +2,7 @@
 
 
 .npminstall: package.json
-	@if ! [ $(shell command -v npm) ]; then \
+	@if ! [ $(shell which npm) ]; then \
```
enahum: but when not found it will return something like `bash: type: npms: not found` and then the if statement will always be true, isn't it?

jespino: At least in Arch Linux works correctly, because the return value (not the string printed) is 1. As far as I know, this must works in every Linux system in the same way (I'm 4/5 about it). I'm 2/5 sure about it must works in MacOS.

enahum: I just tested and it works fine

PR #299, opened 2017-11-03
Add initial state for dev env

sudheerDev:
#### Summary
Add initial state for dev environment.

This PR was tested on: [Mac, Sierra]

### DIFF
--- a/src/store/initial_state.js
+++ b/src/store/initial_state.js
@@ -45,7 +45,10 @@ const state = {
             currentFocusedPostId: '',
             messagesHistory: {
                 messages: [],
-                index: {}
+                index: {
+                    post: -1,
+                    comment: -1
+                }
             }
         },
         preferences: {
--- a/src/store/configureStore.dev.js
+++ b/src/store/configureStore.dev.js
@@ -23,6 +23,7 @@ import {General} from 'constants';
 import serviceReducer from 'reducers';
 import deepFreezeAndThrowOnMutation from 'utils/deep_freeze';
 
+import initialState from './initial_state';
 import {defaultOptions, offlineConfig} from './helpers';
 
 /***
@@ -34,6 +35,7 @@ additionalMiddleware - func | array - Allows for single or multiple additional m
 export default function configureServiceStore(preloadedState, appReducer, userOfflineConfig, getAppReducer, clientOptions = {}) {
     const baseOfflineConfig = Object.assign({}, defaultOfflineConfig, offlineConfig, userOfflineConfig);
     const options = Object.assign({}, defaultOptions, clientOptions);
+    const baseState = Object.assign({}, initialState, preloadedState);
 
     const {additionalMiddleware, enableBuffer} = options;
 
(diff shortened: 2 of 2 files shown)


enahum:
@jarredwitt can you confirm this doesn't break anything with redux offline please

enahum:
@sudheerDev this PR has some build failures and I'm not sure this is needed, but deferring to @jarredwitt

sudheerDev:
@enahum I just added it locally for simulating prod initial state for dev env as it caused few issues before for me. one such is -https://pre-release.mattermost.com/core/pl/hj8ip4hphj8hmc83pqrtmdb1dh. Your call though.

jwilander:
@enahum it's matching what the prod config does

jwilander:
Looks like a couple related tests are failing:
```
 354 passing (13s)
  2 failing

  1) Actions.Posts moveHistoryIndexBack:

      AssertionError: false == true
      + expected - actual

      -false
      +true
      
      at _callee34$ (test/actions/posts.test.js:1465:16)
      at tryCatch (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
      at step (test/actions/posts.test.js:41:191)
      at test/actions/posts.test.js:41:361

  2) Actions.Posts moveHistoryIndexForward:

      AssertionError: false == true
      + expected - actual

      -false
      +true
      
      at _callee35$ (test/actions/posts.test.js:1511:16)
      at tryCatch (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
      at step (test/actions/posts.test.js:41:191)
      at test/actions/posts.test.js:41:361
```
Mind taking a look?

sudheerDev:
I will fix them.

src/store/initial_state.js
```diff
@@ -45,7 +45,10 @@ const state = {
             currentFocusedPostId: '',
             messagesHistory: {
                 messages: [],
-                index: {}
+                index: {
+                    post: -1,
+                    comment: -1
+                }
             }
```
sudheerDev: I Changed the initial state here to reflect reset_state.

Reason, why the tests fails, is because of - this action https://github.com/mattermost/mattermost-redux/blob/ae530a32d0d4c5b0f3c8200f7d7890439e6cf9ca/test/actions/posts.test.js#L1462

Which reduces https://github.com/mattermost/mattermost-redux/blob/ae530a32d0d4c5b0f3c8200f7d7890439e6cf9ca/src/reducers/entities/posts.js#L367 to

 returns existing state as `state.index` exists so, the assertion would not have a value defined. 

I could change assertion as well instead of changing initial state.  Let me know if you want me to change.

sudheerDev:
BTW @jwilander if this works then i think 
https://github.com/mattermost/mattermost-redux/blob/ae530a32d0d4c5b0f3c8200f7d7890439e6cf9ca/src/reducers/entities/posts.js#L367

Few of these in this file probably are not needed anymore.

jwilander:
Probably not, but a bit of defensive coding is never a bad idea in case some bad state gets into the store

PR #302, opened 2017-11-05
Add editingPost information in the redux store

jespino:
Needed for mattermost/mattersmot-webapp#173

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -19,6 +19,17 @@ export function getPost(state, postId) {
     return getAllPosts(state)[postId];
 }
 
+export const getEditingPost = createIdsSelector(
+    (state) => state,
+    (state) => state.entities.posts.editingPost,
+    (state, editingPost) => {
+        return {
+            ...editingPost,
+            post: getPost(state, editingPost.postId)
+        };
+    }
+);
+
 export function getReactionsForPosts(state) {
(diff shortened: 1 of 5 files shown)


src/selectors/entities/posts.js
```diff
@@ -19,6 +19,17 @@ export function getPost(state, postId) {
     return getAllPosts(state)[postId];
 }
 
+export const getEditingPost = createIdsSelector(
```
enahum: wooow! if we are going to keep this PR which I want to discuss further this selector is going to be a performance killer.
1. The createIdsSelector will memoize the arguments and the result of the selector and we normally use it when returning an array of ids in a way that if after the calculations the result has the same values then it returns the same array to avoid unnecessary re-renders and stuff like that.
2. createIdsSelector uses a shallow comparison for every argument and in this case the argument is the whole state (killer)
3. The selector will not memoize anything here as the reality is that the first argument `(state) => state,` will always return a different object thus executing the result function every time.

In this case I'd advice to use a regular function that receives the state as the argument

jespino:
Ups, I'll move it to mattermost-webapp redux part. Sorry.

src/selectors/entities/posts.js
jespino: The use of createIdsSelector was a copy and paste error, I wanted to use createSelector, sorry.

PR #305, opened 2017-11-08
Clean up channel ordering logic and export some helpers

hmhealey:
This goes along with my PR to reduce duplication between the web app and redux versions of ChannelUtils since I need these helper functions in the web app, and the sorting function was just bugging me since it was overly complicated

### DIFF
--- a/src/utils/channel_utils.js
+++ b/src/utils/channel_utils.js
@@ -6,10 +6,10 @@ import {displayUsername} from './user_utils';
 import {getPreferencesByCategory} from './preference_utils';
 
 const channelTypeOrder = {
-    [Constants.OPEN_CHANNEL]: 0,
-    [Constants.PRIVATE_CHANNEL]: 1,
-    [Constants.DM_CHANNEL]: 2,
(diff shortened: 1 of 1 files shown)


src/utils/channel_utils.js
```diff
@@ -5,8 +5,12 @@ import {General, Preferences} from 'constants';
 import {displayUsername} from './user_utils';
 import {getPreferencesByCategory} from './preference_utils';
 
-const defaultPrefix = 'D'; // fallback for future types
-const typeToPrefixMap = {[General.OPEN_CHANNEL]: 'A', [General.PRIVATE_CHANNEL]: 'B', [General.DM_CHANNEL]: 'C', [General.GM_CHANNEL]: 'C'};
+const channelTypeOrder = {
+    [Constants.OPEN_CHANNEL]: 0,
```
enahum: Constants is not defined, I think it has to be `General`

hmhealey: Ah, that's what I get for copying and pasting from the webapp repo

PR #307, opened 2017-11-08
Improve new posts received through the WS

enahum:
#### Summary
* When creating a post from the same client we already have the post so there is no need to replace the current post in the store nor we need to make the channel visible as the post was created from the same client meaning that the channel is already visible.
* If the post was made by the same user we don't need to fetch profiles or status for the current.
* Finally we want to make sure that if you receive a post and you are on the same channel marking the channel as read also sets the last_viewed_at of the channel that way we prevent the `new message indicator` to appear or to flash if later on you receive a `CHANNEL_VIEWED` websocket event

### DIFF
--- a/src/action_types/channels.js
+++ b/src/action_types/channels.js
@@ -77,6 +77,7 @@ export default keyMirror({
     RECEIVED_CHANNEL_PROPS: null,
     RECEIVED_CHANNEL_DELETED: null,
     RECEIVED_MSG_AND_MENTION_COUNT: null,
+    RECEIVED_LAST_VIEWED_AT: null,
     UPDATE_CHANNEL_HEADER: null,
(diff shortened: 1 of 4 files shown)


enahum:
actually @hmhealey we have another one but If we do that there is a small flicker that will show the `new messages` indicator for a small amount of time while it sets the last_viewed_at in another action

hmhealey:
I mean just add an extra action to the batch like
```
actions.push({
    type: RECEIVED_MSG_AND_MENTION_COUNT,
    ...
});

if (updateLastViewedAt) {
    actions.push({
        type: RECEIVED_LAST_VIEWED_AT,
        ...
    });
}
```

Since it's all in one batch, the UI won't re-render between them

enahum:
@hmhealey done

PR #308, opened 2017-11-09
Change isFavoriteChannel to not modify channel object

csduarte:
#### Summary
Changes`isFavoriteChannel` to not modify channel object and return boolean.


enahum:
@jarredwitt build failures, looks like there are two tests that were using the `channel.isFavorite` prop

jarredwitt:
@enahum ready for review

PR #321, opened 2017-11-14
Add new selector getLatestReplyablePost

sudheerDev:
#### Summary
Selector for the latest post to reply.
 
#### Checklist

#### Test Information
This PR was tested on: [Mac, Seirra]

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -391,3 +391,15 @@ export const getMostRecentPostIdInChannel = createSelector(
         return postIdsInChannel[0];
     }
 );
+
+export const getLatestReplyablePostId = createSelector(
+    getPostsInCurrentChannel,
+    (posts) => {
+        for (const post of posts) {
+            if (post.state !== Posts.POST_DELETED && !isSystemMessage(post) && !isPostEphemeral(post)) {
+                return post.id;
+            }
+        }
+        return null;
+    }
+);
--- a/test/selectors/posts.test.js
+++ b/test/selectors/posts.test.js
@@ -1540,4 +1540,50 @@ describe('Selectors.Posts', () => {
             assert.equal(postId, '1002');
         });
     });
+
+    describe('getLatestReplyablePostId', () => {
+        it('no posts', () => {
+            const noPosts = {};
+            const state = {
+                entities: {
+                    posts: {
+                        posts: noPosts,
+                        postsInChannel: []
+                    },
+                    channels: {
+                        currentChannelId: 'abcd'
+                    }
+                }
+            };
+            const actual = Selectors.getLatestReplyablePostId(state);
+
+            assert.equal(actual, null);
+        });
+
+        it('return first post which dosent have POST_DELETED state', () => {
+            const postsAny = {
+                a: {id: 'a', channel_id: 'a', create_at: 1, user_id: 'a'},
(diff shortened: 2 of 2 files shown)


sudheerDev:
Selector needed for create_post component

https://github.com/mattermost/mattermost-webapp/blob/f1fd8469c1d9243c4068fe56ec8d67fd9dcbac39/components/create_post.jsx#L498

src/selectors/entities/posts.js
```diff
@@ -391,3 +391,15 @@ export const getMostRecentPostIdInChannel = createSelector(
         return postIdsInChannel[0];
     }
 );
+
+export const getLatestReplyablePost = createSelector(
+  getPostsInCurrentChannel,
+  (posts) => {
+      for (const post of posts) {
+          if (post.state !== Posts.POST_DELETED && !isSystemMessage(post)) {
```
enahum: This one should also check if the post is ephemeral

enahum: Oh and you may want to return just the post id instead of the whole post, that way you can avoid a re-render if the last post has been edited

sudheerDev: @enahum Makes sense. I will change this to return only ID. 

BTW I copied the logic from here 
https://github.com/mattermost/mattermost-webapp/blob/f1fd8469c1d9243c4068fe56ec8d67fd9dcbac39/stores/post_store.jsx#L67.  Just wanted to check if i need to add `ephemeral` check as well?

sudheerDev:
@enahum Not sure if this selector would be useful in mobile app. Let me know if this has to be pushed to webapp instead.

src/selectors/entities/posts.js
enahum: Well I don't really se myself replying to an ephemeral post, probably that was a bug in the webapp post store.

enahum:
I don't think this will be useful, as this is used with a shortcut key right? Anyway is not that big of a deal but we need to make sure to include in this library things that might be of benefit for every client and not just one.

sudheerDev:
@enahum Fixed review comments.

src/selectors/entities/posts.js
```diff
@@ -391,3 +391,15 @@ export const getMostRecentPostIdInChannel = createSelector(
         return postIdsInChannel[0];
     }
 );
+
+export const getLatestReplyablePost = createSelector(
```
hmhealey: Can you rename this to getLatestReplyablePostId so that callers know it's returning an id instead of the whole object?

src/selectors/entities/posts.js
```diff
@@ -391,3 +391,15 @@ export const getMostRecentPostIdInChannel = createSelector(
...
+  getPostsInCurrentChannel,
```
hmhealey: It's hard to tell here, but can you make sure this is indented with 4 spaces? It looks like it might only have 2. I'm also not sure why Jenkins wouldn't be complaining if that was the case though

src/selectors/entities/posts.js
sudheerDev: Done

src/selectors/entities/posts.js
sudheerDev: Ya, true it had just two spaces. Fixed it. Gave a PR for fixing the eslint rule as well. 
https://github.com/mattermost/mattermost-redux/pull/325/files

PR #323, opened 2017-11-15
Improve marking the channel as read

enahum:
#### Summary
Currently we were using `markAsUnread` to just reset the *msg_count* and *mention_count*  from the channel and then subtract it from the team, but that was causing the channel to be marked as read only in the same client, and the `viewChannel` action was used to set the channel `last_viewed_at` in the client and to inform the server that the channel was actually viewed.

Now `markAsUnread` does reset the counts but also informs the server that the channel has been read and `viewChannel` will only update the `last_viewed_at` locally for the client that way we have better control for displaying or not the **new messages** indicator.

This PR also adds a missing WebSocket event to mark the channel as read when it was read in another device/window/tap/app/etc

#### Ticket Link
https://mattermost.atlassian.net/browse/RN-511

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -658,20 +658,7 @@ export function deleteChannel(channelId) {
 
 export function viewChannel(channelId, prevChannelId = '') {
     return async (dispatch, getState) => {
-        dispatch({type: ChannelTypes.UPDATE_LAST_VIEWED_REQUEST}, getState);
-
-        try {
-            await Client4.viewMyChannel(channelId, prevChannelId);
-        } catch (error) {
-            forceLogoutIfNecessary(error, dispatch);
-            dispatch(batchActions([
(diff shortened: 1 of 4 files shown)


jwilander:
I'm OK with this but can we rename the `viewChannel` action to something like `viewChannelLocal`? Right now it lines up in naming with REST API `/channels/view` so someone calling it might expect it to hit that endpoint

Also do the tests need to be updated?

jwilander:
This will also require a change in the webapp

enahum:
@jwilander done

jwilander:
Sorry, I made a mistake. We need to keep `viewChannel` around in it's previous form too. We don't want to break anyone using mattermost-redux as a driver

enahum:
Oh ok sure thing makes sense, I’ll do that tomorrow then

enahum:
@jwilander done

PR #326, opened 2017-11-16
Use remote-redux-devtools only when browser extension is not available

jespino:


jwilander:
@enahum I don't think this should cause you any issue on RN but maybe give it a try first?

enahum:
This should not cause any issues

PR #329, opened 2017-11-21
PLT-2068: Bases the custom theme off of the default system theme in c…

mkraft:
#### Summary
Ensures object returned by the `getTheme` selector has all of the properties present in the `default` system themes.

#### Ticket Link
[PLT-2068](https://mattermost.atlassian.net/browse/PLT-2068)

#### Checklist

#### Test Information
* Chrome on MacOS v62.0.3202.94

### DIFF
--- a/src/selectors/entities/preferences.js
+++ b/src/selectors/entities/preferences.js
@@ -152,7 +152,7 @@ export const getTheme = createShallowSelector(
             }
         }
 
-        return Object.assign({}, Preferences.THEMES.organization, theme);
+        return Object.assign({}, Preferences.THEMES.default, theme);
     }
 );
 
--- a/src/selectors/entities/preferences.js
+++ b/src/selectors/entities/preferences.js
@@ -138,27 +138,21 @@ export const getTheme = createShallowSelector(
         // At this point, the theme should be a plain object
 
         // If this is a system theme, find it in case the user's theme is missing any fields
-        let baseTheme = Preferences.THEMES.default;
-        if (theme.type) {
-            for (const key of Object.keys(Preferences.THEMES)) {
-                if (Preferences.THEMES[key].type === theme.type) {
-                    baseTheme = Preferences.THEMES[key];
-                    break;
-                }
+        if (theme.type && theme.type !== 'custom') {
+            const match = Object.entries(Preferences.THEMES).find(([, v]) => v.type === theme.type);
+            if (match) {
+                return match[1];
             }
         }
 
-        for (const key of Object.keys(baseTheme)) {
+        for (const key of Object.keys(Preferences.THEMES.default)) {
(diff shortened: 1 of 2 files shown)


src/selectors/entities/preferences.js
```diff
@@ -138,27 +138,19 @@ export const getTheme = createShallowSelector(
...
                 // Fix a case where upper case theme colours are rendered as black
                 theme[key] = theme[key].toLowerCase();
-            } else {
-                // This theme is missing some colours, so fall back to the default theme when necessary
-                theme[key] = baseTheme[key];
             }
         }
 
-        return theme;
+        return Object.assign({}, Preferences.THEMES.organization, theme);
```
hmhealey: I think this ticket is out of date. The default theme should be "default", not "organization"

src/selectors/entities/preferences.js
```diff
@@ -138,27 +138,19 @@ export const getTheme = createShallowSelector(
...
-        if (theme.type) {
-            for (const key of Object.keys(Preferences.THEMES)) {
-                if (Preferences.THEMES[key].type === theme.type) {
-                    baseTheme = Preferences.THEMES[key];
-                    break;
-                }
-            }
+        const baseTheme = Preferences.THEMES.default;
+        if (theme.type && theme.type !== 'custom') {
+            return baseTheme;
```
hmhealey: This should use the matching system theme like it was doing before, but it should return the whole theme instead of using that to cherry-pick missing fields

src/selectors/entities/preferences.js
mkraft: @asaadmahmood Can you confirm please? Did you intend for missing values to be overwritten with the `default` team as it was before or with `organization` theme as indicated in the ticket? Thanks.

test/selectors/preferences.test.js
```diff
@@ -329,6 +329,28 @@ describe('Selectors.Preferences', () => {
...
+                    preferences: {
+                        myPreferences: {
+                            [getPreferenceKey(Preferences.CATEGORY_THEME, '')]: {
+                                category: Preferences.CATEGORY_THEME, name: '', value: JSON.stringify(theme)
+                            }
+                        }
+                    }
+                }
+            }).codeTheme, Preferences.THEMES.windows10.codeTheme);
+        });
```
mkraft: @hmhealey You're right! I have added this test case to cover that miss.

src/selectors/entities/preferences.js
asaadmahmood: Harrison is correct @mkraft. It should be the Mattermost default theme.

mkraft:
@hmhealey @asaadmahmood Updated code (and Jira ticket) to use `default` system theme rather than `organization`.

PR #330, opened 2017-11-22
Add autocomplete command support

csduarte:
#### Summary
Adds autocomplete command support for supporting slash command execution.

#### Test Information
This PR was tested on:
iPhone 8 iOS 11


enahum:
@jarredwitt 

1) Actions.Integrations getCustomTeamCommands:
     AssertionError: undefined == true
      at _callee15$ (test/actions/integrations.test.js:409:16)
      at tryCatch (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
      at step (test/actions/integrations.test.js:37:191)
      at test/actions/integrations.test.js:37:361
      at process._tickCallback (internal/process/next_tick.js:103:7)

src/client/client4.js
```diff
@@ -1550,6 +1550,13 @@ export default class Client4 {
         );
     };
 
+    getAutocompleteCommandsList = async (teamId, page = 0, perPage = PER_PAGE_DEFAULT) => {
+        return this.doFetch(
+            `${this.getTeamRoute(teamId)}/commands/autocomplete${buildQueryString({page, per_page: perPage})}`,
```
enahum: This endpoint as it is now won't do anything with the paging but we can leave it there for when the endpoint does the implementation for it

src/reducers/entities/integrations.js
```diff
@@ -89,11 +89,16 @@ function outgoingHooks(state = {}, action) {
...
+        const nextCommands = {};
         for (const command of action.data) {
-            nextState[command.id] = command;
+            if (command.id) {
+                const id = command.id;
+                nextCommands[id] = command;
+            }
         }
-        return nextState;
+        return nextCommands;
```
enahum: @jwilander are you ok with this changes here?

src/reducers/entities/integrations.js
```diff
@@ -122,6 +127,34 @@ function commands(state = {}, action) {
     }
 }
 
+function executableCommands(state = {}, action) {
```
enahum: I'm wondering if this would be best to call it `builtInCommands` instead of `executableCommands` but I'm 0/5, definitely non-blocking for me

src/reducers/entities/integrations.js
jarredwitt: I'm actually going to revert this. This is what is causing the current tests to fail and it's not longer needed since we're filtering for team commands at the selector now.

src/reducers/entities/integrations.js
jwilander: Maybe even `systemCommands` because really they're just commands without teams right? And "system" kind of means "built-in" too

jarredwitt: I'm good with systemCommands. @enahum work for you?

enahum: I don't know if you saw my reaction, but just in case the answer was YES

jarredwitt:
@enahum @jwilander changes made and tests added

src/reducers/entities/integrations.js
```diff
@@ -89,17 +89,26 @@ function outgoingHooks(state = {}, action) {
 function commands(state = {}, action) {
     const nextState = {...state};
     switch (action.type) {
+    case IntegrationTypes.RECEIVED_COMMANDS:
     case IntegrationTypes.RECEIVED_CUSTOM_TEAM_COMMANDS: {
         for (const command of action.data) {
-            nextState[command.id] = command;
+            if (command.id) {
+                const id = command.id;
```
jwilander: Not a big deal and non-blocker, but weird to allocate a variable then immediately use it

PR #331, opened 2017-11-23
Ensure serverId (diagnosticId) is always sent to segment

enahum:
Need this to update https://github.com/mattermost/mattermost-mobile/pull/1192

Note: This will also ensure that the right diagnosticId aka `serverId`  is sent with every track event to segment on the webapp which was not the case.

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -30,6 +30,7 @@ export default class Client4 {
         this.enableLogging = false;
         this.defaultHeaders = {};
         this.userId = '';
+        this.diagnosticId = '';
         this.includeCookies = true;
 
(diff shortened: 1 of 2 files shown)


src/client/client4.js
```diff
@@ -2133,7 +2138,8 @@ export default class Client4 {
                 options.context = global.analytics_context;
             }
             global.analytics.track(Object.assign({
-                event: 'event'
+                event: 'event',
+                userId: this.diagnosticId
```
jwilander: It's supposed to be under `userId`?

enahum: yes!

PR #333, opened 2017-11-23
Fix selectors for initial load of create_post redux container 

sudheerDev:
#### Summary
Fix selectors for initial load of create_post redux container as posts will not be available on the first call

#### Test Information
This PR was tested on: [Mac, Sierra]


jwilander:
I'm always for some defensive JS coding 👍

PR #335, opened 2017-11-24
getPostIdsInChannel selector

enahum:


enahum:
@jwilander no idea why this is failing the tests, my code should not affect that at all.

jwilander:
Agreed, I'll force merge it and we can look into the test later

PR #337, opened 2017-11-26
Allow to join group and direct messages

jespino:
#### Summary
This is only part of the solution, the other part is in mattermost/mattermost-webapp#365

#### Summary
This is part of the fix for enter directly in the message when `/groupmsg` is executed (solve the same problem for `/msg` command)

#### Ticket Link
[PLT-7111](https://mattermost.atlassian.net/browse/PLT-7111)

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -582,7 +582,11 @@ export function joinChannel(userId, teamId, channelId, channelName) {
                 channel = await Client4.getChannel(channelId);
             } else if (channelName) {
                 channel = await Client4.getChannelByName(teamId, channelName);
-                member = await Client4.addToChannel(userId, channel.id);
+                if ((channel.type === General.GM_CHANNEL) || (channel.type === General.DM_CHANNEL)) {
+                    member = await Client4.getChannelMember(channel.id, userId);
+                } else {
+                    member = await Client4.addToChannel(userId, channel.id);
+                }
             }
         } catch (error) {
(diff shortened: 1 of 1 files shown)


src/actions/channels.js
```diff
@@ -582,7 +582,11 @@ export function joinChannel(userId, teamId, channelId, channelName) {
                 channel = await Client4.getChannel(channelId);
             } else if (channelName) {
                 channel = await Client4.getChannelByName(teamId, channelName);
-                member = await Client4.addToChannel(userId, channel.id);
+                if ((channel.type === General.GM_CHANNEL) || (channel.type === General.DM_CHANNEL)) {
+                    member = await Client4.getChannelMember(channel.id, userId);
```
enahum: If the group or dm does not exist then the channel member does not exist either, im not sure this works

jespino: Exactly. Currently the group or channel must exist, if not it raise an error, which is captured later.

The current behavior is raise an error (in the addToChannel) always when you try to join an existing or not existing DM/GM channel.

The full workflow is, send a `/groupmsg` or `/msg` which creates the DM or GM, and return a GotoLocation to redirect to the channel. Because the channel, doesn't exist in the store, It try to join to the channel, and because is a group or msg, it fails. With this PR it joins because expect when you join to a DM o GM you must already be a member, if not, fails anyway.

enahum: Makes sense

PR #339, opened 2017-11-28
Add source maps generation to build/dev/dev:watch

jespino:
This is a proposal based on this conversation: https://pre-release.mattermost.com/core/pl/ke489sx953f89jsmupittcwgrw

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
   "scripts": {
     "build": "babel src --out-dir .",
     "webpack": "webpack",
-    "dev": "babel src --out-dir ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -47,10 +47,10 @@
     "ws": "2.3.1"
   },
   "scripts": {
-    "build": "babel src --out-dir .",
+    "build": "babel src --out-dir . --source-maps",
```
jwilander: Do we want them for the regular build? Wondering if there's a benefit to including them

jespino: Yes, the reason to include it in regular build is for get better sentry (or sentry-like) reports (but of course, its not necesary).

enahum: please don't do this, you could create another script if you want, but this will cause the RN app to have more than it actually needs when building for release

jespino: Completely agree, was a proposal based on the comments in the redux channel, but of course, is better to remove it :)

jespino:
Changed! :)

PR #340, opened 2017-11-29
Reexporting redux-persist for webapp and mobile usage

jespino:


jwilander:
I don't think so, I'm pretty sure webpack doesn't take into account dependencies of dependencies and will complain

hmhealey:
I just tried and it works fine with webpack. I think it just cares that the folder exists in node_modules, not that it exists in package.json. If we wanted to be extra sure, we could include it in the web app's peerDepdendencies, but I think that might ruin the point of this

jespino:
Yes, actually, it works. Anyway I don't love the idea of use directly a dependency of a dependency.

jwilander:
No me neither, I'd prefer to be explicit about it

hmhealey:
If we're wanting to be explicit, then why remove it from their package.json in the first place? The only thing this would change is that when we update it, we don't have to update the package.json in the webapp/mobile repos. We'll still have to update the version of mattermost-redux and do any other required changes in the other repo, so why not also update redux-persist there at the same time?

Or like I mentioned above, we could make it a peerDependency in one of the repositories with looser version restrictions so that it's still included, but we won't need to make sure the version numbers are exact

jespino:
I don't know too much about peer dependencies, but i have read about it, and if I understood it correctly, we can force a exact redux-persist peer dependency in mattermost-redux and if any user of the mattermost-redux library use other version, mpm will complain about it. So we can keep the dependency in package.json of the other repositories ensuring that we always are synced.

It's that correct? If so, then it solves the problem of libraries out of sync and we can have the dependency explicitly declared in all the apps.

enahum:
I really don't see why using it this way is not good enough

hmhealey:
@jespino Yeah, it sounds like that would work. I think they recommend not requiring an exact version from the peer dependency, but I think it just prints a warning if it has a different version that doesn't quite match, so I think we can use an exact one for now.

Can you also see if there's any other libraries that could be moved to peer dependencies? It looks like at least `redux`, `redux-batched-actions`, and `reselect` are required by both mattermost-webapp and mattermost-redux

@elias Re-exporting the one library would fix this case, but it wouldn't help for any of the other libraries that are included by both packages. It would also require that we keep the RN and web app completely in sync which may not always happenI think peer dependencies are the intended way to fix a problem like this.

enahum:
@hmhealey agree!!

jespino:
Ok, I will take a look to it and try to solve the problem with peerDependencies :)

PR #341, opened 2017-12-01
Make requests run in parallel wherever possible

hmhealey:
I came a couple places where we're awaiting requests in sequence instead of doing them in parallel, so I took a pass over all of the actions to fix wherever I could find that we were doing it unnecessarily. There were also a few unnecessary awaits that I got rid of where we were awaiting on a value that had already been resolved (which is a noop)

#### Test Information

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -197,7 +197,7 @@
         "sort-imports": 0,
         "sort-keys": 0,
         "space-before-blocks": [2, "always"],
-        "space-before-function-paren": [2, "never"],
+        "space-before-function-paren": [2, {"anonymous": "never", "named": "never", "asyncArrow": "always"}],
         "space-in-parens": [2, "never"],
         "space-infix-ops": 2,
         "space-unary-ops": [2, { "words": true, "nonwords": false }],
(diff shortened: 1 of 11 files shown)


enahum:
You have some build failures

enahum:
@hmhealey eslint errors

/home/ubuntu/workspace/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux/src/actions/channels.js
  411:62  error  Missing semicolon  semi

jwilander:
There's 15 failing tests in the build

enahum:
@hmhealey what happened with this one?

hmhealey:
It's blocked on issues with the tests not working in parallel. My changes exposed some tests that were throwing exceptions but still passed, so I was looking at if there was a good way to catch those.

I wasn't finding anything good that we could do, so I'll probably just fix the ones I found and move on.

hmhealey:
Marking this as Awaiting PR since we don't have a DO NOT MERGE label here, and I don't want this merged until 4.5 is cut

jwilander:
Alright I branched for 4.5, this is safe to merge when you want

PR #342, opened 2017-12-01
Add canTimeout flag for mobile

csduarte:
#### Summary
Adds 2 properties that will be utilized by the mobile application for timing out API calls.

Here's the PR on the mobile side for more context.: https://github.com/mattermost/mattermost-mobile/pull/1242

@miguelespinoza

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -58,6 +58,8 @@ export function createPost(post, files = []) {
             },
             meta: {
                 offline: {
+                    canTimeout: true,
+                    timeout: 100,
                     effect: () => Client4.createPost({...newPost, create_at: 0}),
                     commit: (success, payload) => {
                         // Use RECEIVED_POSTS to clear pending posts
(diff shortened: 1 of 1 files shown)


enahum:
@miguelespinoza build failures

miguelespinoza:
Aw man, I was debugging short timeouts and left 100ms. I meant for 10 secs

enahum:
so this is only for creating a post?

I saw that you have a PR in `mattermost-mobile` that actually makes sense to have it along side this, actually if I get it right this is going to be used when overriding the `effect` in  `redux-offline`, right?

@jarredwitt thoughts?

miguelespinoza:
@enahum correct `mattermost-mobile` will depend on this change
We override the `effect` reconciler in mobile so it only affects the mobile clients. 

Mobile has been a pain for our users. Reasons have been described in this PR: https://github.com/mattermost/mattermost-mobile/pull/1242

If we want to add this solution to all clients that's something we can consider.  It wouldn't hurt for the web app to utilize this solution but from the users' experience it's an issue directly on the mobile side

enahum:
Ok Undestood

PR #346, opened 2017-12-05
Fix GM/DM sidebar ordering

jwilander:
#### Summary
There was a bug in the sort, it shouldn't compare what the "types" are, only what the "order types" are. Caused by #305 

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-8262


hmhealey:
@enahum This does not affect RN. It uses a different selector for each type of channel, so it uses a different comparison function that ignores the type

PR #347, opened 2017-12-06
Do not modify WS protocol

enahum:
#### Summary
Currently we are modifying the url to connect the WebSocket, we replace http(s) with ws(s) protocols, and it seems that is causing some issues were the server cannot validate the origin.

This PR removes the replacement and leave the url as is and lets the client underlying WebSocket library to handle it.

#### Ticket Link
Closes mattermost/mattermost-mobile/issues/1263

#### Checklist

#### Test Information
This PR was tested on: Mobile app Android and iOS

### DIFF
--- a/src/actions/websocket.js
+++ b/src/actions/websocket.js
@@ -50,25 +50,8 @@ import EventEmitter from 'utils/event_emitter';
 
 export function init(platform, siteUrl, token, optionalWebSocket) {
     return async (dispatch, getState) => {
-        const config = getState().entities.general.config;
         let connUrl = siteUrl || Client4.getUrl();
         const authToken = token || Client4.getToken();
-
-        // replace the protocol with a websocket one
-        if (connUrl.startsWith('https:')) {
-            connUrl = connUrl.replace(/^https:/, 'wss:');
-        } else {
-            connUrl = connUrl.replace(/^http:/, 'ws:');
-        }
-
-        // append a port number if one isn't already specified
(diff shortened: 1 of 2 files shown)


jwilander:
Do the native browser implementations of websockets handle the protocol replacement?

enahum:
By looking at https://i.imgur.com/xrvGyhQ.png where the one at the bottom is Firefox I would say yes.

jwilander:
That looks like it's still using `https` to me?

jwilander:
This just worries me as all the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications) I see shows passing in `ws` or `wss`

jwilander:
It would make me feel better to do this as an option, instead of by default

enahum:
right, I know that.. but somehow that is messing with the authorized origins and by using http or https as the protocol works fine and we don't get the origin issue.

jwilander:
I'm pretty sure it's just a bug in the WebSocket implementation used by RN (it works fine on the webapp). I'll be OK with it if you it make it an option, I just don't want to change the behavior for the webapp and break on someone's browser

enahum:
alright I'll make it optional

enahum:
@jwilander made it so only applies when the platform is mobile

alecisec:
Does this mean that the 'SiteURL' should then be set to wss://foo.com instead of https://foo.com ? My understanding was that this is what's used to validate the Origin for upgrade requests.

PR #348, opened 2017-12-06
Share channel selectors between rn and webapp

jespino:


enahum:
I did some RN testing and everything seems to be working fine

hmhealey:
I'm not merging this yet, since I don't know if we should be waiting for 4.5 to be cut

jespino:
I will need it for the LHS to redux/pureComponent migration, and that wont be ready for the 4.5, so we can (and in my opinion, we should) wait to merge it after the 4.5 cut.

jwilander:
It's safe to merge now, I've tagged the redux commit for 4.5 and can branch from there for any needed changes

jwilander:
Ah, I didn't realize @ccbrown's latest PR was for 4.5 but it should be fine I'll revert it if there's any issues

PR #349, opened 2017-12-06
PLT-8302: figure out the channelId in RECEIVED_POSTS handler

ccbrown:
#### Summary
There are a few cases where the `RECEIVED_POSTS` event doesn't come with a `channelId`. This causes the `postsInChannel` map to get an "undefined" channel inserted into it. This patch makes the handler pick up the channel id from one of the posts if it's not part of the action.

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-8302

#### Checklist
N/A

#### Test Information
This PR was tested on: Chrome, macOS

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -33,13 +33,17 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, action) {
 
 function handleReceivedPosts(posts = {}, postsInChannel = {}, action) {
     const newPosts = action.data.posts;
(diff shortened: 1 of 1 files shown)


ccbrown:
@enahum Actually, it looks like there was just one place: https://github.com/mattermost/mattermost-webapp/pull/424

I'm gonna close this in favor of that for consistency. Maybe at some point it would make sense to just remove the channelId from the action in all cases since it seems to be totally redundant.

PR #351, opened 2017-12-07
PLT-8321: show channels if unread or recently viewed

ccbrown:
#### Summary
This makes channels always show up in the side-bar if they have unread messages.

To prevent them from disappearing immediately once you view those unread messages, it also makes channels stay around for approximately 7 days after you last viewed them.

See conversation here: https://pre-release.mattermost.com/core/pl/e4fjurdmcpd8783qx9q4t9573r

#### Ticket Link
https://mattermost.atlassian.net/browse/PLT-8321

#### Checklist
N/A

#### Test Information
This PR was tested on: macOS, Chrome

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -662,6 +662,18 @@ export function deleteChannel(channelId) {
 
 export function viewChannel(channelId, prevChannelId = '') {
     return async (dispatch, getState) => {
+        const {currentUserId} = getState().entities.users;
+
+        const {myPreferences} = getState().entities.preferences;
+        const viewTimePref = myPreferences[`${Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME}--${channelId}`];
+        const viewTime = viewTimePref ? parseInt(viewTimePref.value, 10) : 0;
+        if (viewTime < new Date().getTime() - (3 * 60 * 60 * 1000)) {
+            const preferences = [
+                {user_id: currentUserId, category: Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME, name: channelId, value: new Date().getTime().toString()}
+            ];
+            savePreferences(currentUserId, preferences)(dispatch, getState);
+        }
+
         dispatch({type: ChannelTypes.UPDATE_LAST_VIEWED_REQUEST}, getState);
 
         try {
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -32,7 +32,8 @@ import {
     getDirectChannelName,
     isAutoClosed,
     isDirectChannelVisible,
-    isGroupChannelVisible
+    isGroupChannelVisible,
+    isGroupOrDirectChannelVisible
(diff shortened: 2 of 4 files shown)


src/actions/channels.js
```diff
@@ -662,6 +662,18 @@ export function deleteChannel(channelId) {
...
+
+        const {myPreferences} = getState().entities.preferences;
+        const viewTimePref = myPreferences[`${Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME}--${channelId}`];
+        const viewTime = viewTimePref ? parseInt(viewTimePref.value, 10) : 0;
+        if (viewTime < new Date().getTime() - (3 * 60 * 60 * 1000)) {
+            const preferences = [
+                {user_id: currentUserId, category: Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME, name: channelId, value: new Date().getTime().toString()}
+            ];
+            savePreferences(currentUserId, preferences)(dispatch, getState);
+        }
```
ccbrown: Unfortunately the `last_viewed_at` time that the server keeps track of doesn't work. That time is actually the time of the last post in the channel when you view it. So I'm storing another preference here with an approximate view time. (Approximate because there's no need to add an additional request to every channel view.)

src/selectors/entities/channels.js
```diff
@@ -575,3 +577,16 @@ export const getSortedDirectChannelIds = createIdsSelector(
...
+    getMyChannelMemberships,
+    getConfig,
+    getMyPreferences,
+    getCurrentUser,
+    getUsers,
+    getLastPostPerChannel,
+    (channel, memberships, config, myPreferences, currentUser, users, lastPosts) => {
+        return isGroupOrDirectChannelVisible(channel, memberships, config, myPreferences, currentUser.id, users, lastPosts);
+    }
+);
```
ccbrown: Updating the webapp to use this instead of calling `isXVisible` directly. Simplifies those call sites a lot.

src/actions/channels.js
ccbrown: I'm not particularly happy with so I'm open to alternatives. I think ideally the members table would probably have another column with the actual wall time, and the /view API would keep those up-to-date.

src/selectors/entities/channels.js
hmhealey: Actually, can you either make this into a `makeGetGroupOrDirectChannelVisibility` that returns a selector or just make it a plain function? Since it takes an argument besides the state, memoization is impossible with a single instance if there's multiple places calling this, so I'd rather either make it a factory (so it can be memoized if necessary) or not have it be a selector at all so there's no illusion that it would be memoized well.

ccbrown:
@hmhealey Updated. Plain function seemed simpler.

src/actions/channels.js
jwilander: Hmmm yeah, I think this is fine for now but you're right ideally this is handled in the backend by a change to the members table

jwilander:
Needs a rebase

PR #356, opened 2018-01-07
[PLT-7793] Rename userAccessTokens + fetch all userAccessTokens

watadarkstar:
#### Summary

Renames `userAccessTokens` to `userAccessTokensByUser` and now `userAccessTokens` fetches all `UserAccessTokens` in the system (see notes for more details).

#### PR Links

- [Documentation PR](https://github.com/mattermost/mattermost-api-reference/pull/318)
- [Webapp PR](https://github.com/mattermost/mattermost-webapp/pull/496)
- [Server PR](https://github.com/mattermost/mattermost-server/pull/8038)
- [Redux PR](https://github.com/mattermost/mattermost-redux/pull/356)

#### Ticket Link

- [Jira](https://mattermost.atlassian.net/browse/PLT-7793)
- [Github](https://github.com/mattermost/mattermost-server/issues/7584#issuecomment-352808489)

#### Notes

- To improve scalability I renamed `userAccessTokens` to be `userAccessTokensByUser` which can be used like this `userAccessTokensByUser[userId][tokenId]` and then there is another object called `userAccessTokens` which can be used like this: `userAccessTokens[tokenId]`. They both should stay updated with each other.

#### Checklist
- [x] Redux: `mapStateToProps` to get tokens
- [x] Redux: rename `userAccessTokens` to `userAccessTokensByUser`
- [x] Redux: make `userAccessTokens` now fetch all the `UserAccessTokens`
- [x] Redux: fix failing tests + new tests


watadarkstar:
@jwilander A few notes on how this should get merged:

- Please add a mobile dev reviewer since they may be using `userAccessTokens` and this will break the mobile app if its not up to date. i.e. they should be using `userAccessTokensByUser` instead of `userAccessTokens`.
- We need to merge this PR first, so I can change whats in the webapp's `package.json` file:

```
"mattermost-redux": "mattermost/mattermost-redux#767db254e463e3ee99be69b7eec870345daedaf8",
```

watadarkstar:
@jwilander I found there is inconsistency in `mattermost-redux` regarding `userAccessToken(s)`. Can we agree that `userAccessToken(s)` **never** belong in `admin.js` files? 

Sometimes I find `userAccessToken` related code in `admin.js` files and sometimes I find it in `users.js` files. So far I have been writing code to be consistent with the chaos. For example, you will find `RECEIVED_USER_ACCESS_TOKENS_FOR_USER` inside `action_types/admin.js`, yet you will find `getUserAccessTokensForUser` inside `actions/users.js`.

I kind of understand why this might be the case: `RECEIVED_USER_ACCESS_TOKENS_FOR_USER` is something only the admin can do (unless of course the user is receiving his/her own tokens), but the code organization is unclear sometimes when you have entities that pertain to both the user and the admin.

So, I'd like to make this more consistent in my PR if possible, i.e. I'd like to put everything regarding `userAccessToken(s)` inside `users.js`. What do you think?

watadarkstar:
@jwilander This is no longer a WIP and is ready to be reviewed.

jwilander:
Whoops, I was typing out a response earlier but got distracted and didn't finish it. I'm not at my computer right now but the short version is that I wouldn't really consider it an inconsistency. IIRC the only place there should be accessToken references outside of user related files is in the admin reducer. To me that one makes perfect sense to be in there as it's storing/reducing the token objects that only admins will care about. So I think it fits the model fairly well

As it is everything inside the user reducer will be the same for all users, no matter who's logged in, and the admin reducer stores anything used by just the system console/admin. I do see your point about an entity that's mostly concerned with users being in a reducer that's not the users reducer but I really don't think it's that unusual. For example the channels reducer has references to teams

If we were designing this from scratch I'd probably be more open to doing it the way you suggest but since it's already implemented the way it is, I'd prefer leaving it

Sorry if that response is a bit scatter brained, writing this from my phone

jwilander:
@enahum I don't think this will affect RN but please confirm

watadarkstar:
@jwilander Sounds good Joram, let's leave it as is then. 👍

jwilander:
Sounds good! I'll take a look at the code in the morning

src/reducers/entities/admin.js
```diff
@@ -221,6 +264,16 @@ function userAccessTokens(state = {}, action) {
 
         return {...state, [action.userId]: nextUserState};
     }
+    case AdminTypes.RECEIVED_USER_ACCESS_TOKENS: {
+        const nextUserState = {};
+
+        for (const uat of action.data) {
+            nextUserState[uat.user_id] = {...nextUserState[uat.user_id] || {}};
```
jwilander: Unless I'm missing something (and this isn't really a big deal), but I think it would be a little more efficient to do something like:

```javascript
if (!nextUserState[uat.user_id]) {
  nextUserState[uat.user_id] = {};
}
```
or if you don't want the if statement
```javascript
nextUserState[uat.user_id] = nextUserState[uat.user_id] || {};
```

I don't think it's necessary to make a new object every time with `...`

watadarkstar: I think its okay we do this. The example I can think of where this **may** fail would be this one:

```jsx
class Wrapper = (props) => {
    <InnerWrapper tokens={props.userAccessTokensByUser[props.userId]}>
}
```

However since we use the spread operator on `...nextUserState`, I think `<Wrapper>` will get re-rendered and therefore `<InnerWrapper>` will also get re-rendered.  So it should be okay. I made the change.

PR #359, opened 2018-01-15
[PLT-8494] Remove team object from entities.teams.myMembers when team is deleted

saturninoabril:
#### Summary
- Remove team object from entities.teams.myMembers when team is deleted

#### Ticket Link
Jira ticket: [PLT-8494](https://mattermost.atlassian.net/browse/PLT-8494)

#### Checklist
- [x] has mattermost-webapp change (https://github.com/mattermost/mattermost-webapp/pull/611)

#### Test Information
This PR was tested on: [Chrome/FF browsers, MacOS]

### DIFF
--- a/src/reducers/entities/teams.js
+++ b/src/reducers/entities/teams.js
@@ -51,10 +51,10 @@ function teams(state = {}, action) {
 }
 
 function myMembers(state = {}, action) {
-    function updateState(receivedTeams = {}, currenState = {}) {
+    function updateState(receivedTeams = {}, currentState = {}) {
         return Object.keys(receivedTeams).forEach((teamId) => {
-            if (receivedTeams[teamId].delete_at > 0 && currenState[teamId]) {
-                Reflect.deleteProperty(currenState, teamId);
+            if (receivedTeams[teamId].delete_at > 0 && currentState[teamId]) {
+                Reflect.deleteProperty(currentState, teamId);
             }
         });
     }
--- a/src/reducers/entities/teams.js
+++ b/src/reducers/entities/teams.js
@@ -51,6 +51,14 @@ function teams(state = {}, action) {
 }
 
 function myMembers(state = {}, action) {
+    function updateState(receivedTeams = {}, currenState = {}) {
+        return Object.keys(receivedTeams).forEach((teamId) => {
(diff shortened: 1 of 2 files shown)


src/reducers/entities/teams.js
```diff
@@ -64,7 +64,28 @@ function myMembers(state = {}, action) {
...
+            if (teamsList[myMemberTeamId] && teamsList[myMemberTeamId].delete_at) {
+                Reflect.deleteProperty(nextState, myMemberTeamId);
+            }
+        });
 
+        return nextState;
+    }
+    case TeamTypes.RECEIVED_TEAMS: {
+        const nextState = {...state};
+        const team = action.data;
```
jwilander: Can this be `teams` instead of `team`?

src/reducers/entities/teams.js
```diff
@@ -64,7 +64,28 @@ function myMembers(state = {}, action) {
...
+        Object.keys(team).forEach((teamId) => {
```
jwilander: Here you're looping through the teams you've received but in the above case you're looping through all of state. Is there a reason for doing it differently? I'd prefer to be consistent, leaning towards looping through the incoming data rather than the state as the incoming data is likely to be smaller

saturninoabril: For the RECEIVED_TEAMS_LIST, I just thought the incoming data would be bigger than the state that's why I did looping through state. In addition to your comment, the difference between the two might not really be significant, so consistency is the real deal.

saturninoabril:
Add `updateState` function for RECEIVED_TEAMS_LIST and RECEIVED_TEAMS_LIST action types, and then added tests

enahum:
@jwilander please re-review when you have a chance

src/reducers/entities/teams.js
```diff
@@ -51,6 +51,14 @@ function teams(state = {}, action) {
 }
 
 function myMembers(state = {}, action) {
+    function updateState(receivedTeams = {}, currenState = {}) {
```
jwilander: Typo here `currenState` ->`currentState` then it looks good

This function's also a bit too much abstraction in my opinion but 0/5

saturninoabril:
Done fixing typo. Thanks!

PR #360, opened 2018-01-15
XYZ-21: new permissions roles management

jespino:
#### Summary
Add api client, actions, reducers and selectors to manage advanced permissions roles.

**DO NOT MERGE**: This depends on [XYZ-37](https://mattermost.atlassian.net/browse/XYZ-37).

#### Ticket Link
[XYZ-21](https://mattermost.atlassian.net/browse/XYZ-21)

#### Checklist

#### Test Information
This PR was tested on: Firfox 57, on Arch Linux


src/reducers/entities/roles.js
```diff
@@ -0,0 +1,45 @@
...
+    const nextState = {...state};
+    switch (action.type) {
+    case RoleTypes.RECEIVED_ROLES: {
+        if (action.data) {
+            for (const role of action.data) {
+                nextState[role.name] = role;
+            }
+        }
+
+        return nextState;
```
jwilander: Minor but if you're not modifying `nextState` it's probably better to just return `state` so it's not a new object

src/selectors/entities/roles.js
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+import {createSelector} from 'reselect';
+
+import {getCurrentUser} from 'selectors/entities/common';
+import {getTeamMemberships} from 'selectors/entities/teams';
+import {getMyChannelMemberships} from 'selectors/entities/channels';
+
+export const getMyRoles = createSelector(
```
jwilander: Would it possibly be better to split this into `getMySystemRoles`, `getMyTeamRoles`, `getMyChannelRoles`? Just wondering because it's going to be returning new objects every time for `system`, `team` and `channel` fields even if they don't change (e.g the current user changes but not the team or channel members)

Non-blocking comment, and really I'm 0/5. Just something to think about

enahum: Let me jump here saying that i’m actually 3/5 on splitting the roles as much as possible to avoid a new object cause i’m assuming those objects are going to be used everywhere unless i’m totally wrong and is a one time thing

jespino: Completely agree, actually, the getMyXPerms and haveIXPerm selectors depends on that, so will improve that too using only the required ones.

src/reducers/entities/roles.js
jespino: Fixed

src/selectors/entities/roles.js
jespino: Fixed too

jespino:
Anyway! DO NOT MERGE! It must be merge with the permissions backend feature.

src/utils/channel_utils.js
```diff
@@ -240,11 +251,20 @@ export function showCreateOption(config, license, channelType, isAdmin, isSystem
     return true;
 }
 
-export function showManagementOptions(config, license, channel, isAdmin, isSystemAdmin, isChannelAdmin) {
+export function showManagementOptions(state, config, license, channel, isAdmin, isSystemAdmin, isChannelAdmin) {
     if (license.IsLicensed !== 'true') {
```
grundleborg: I think these license checks should only be here if we do *not* have "new permissions system".

grundleborg:
@jespino I changed the target branch for this to `advanced-permissions-phase-1`, and I re-reviewed since you pushed some more commits since before. There are a few places where you are doing a license check where I think it should only be done if we *don't* have "new permissions". But otherwise looks great.

src/utils/channel_utils.js
jespino: Now must be fine :)

PR #361, opened 2018-01-17
Only logout once if status code is 401

enahum:
#### Summary
Ensure that logout is run only once if we get a 401 response, meaning that the user has to be logged in before performing the `forceLogoutIfNecessary` action

### DIFF
--- a/src/actions/helpers.js
+++ b/src/actions/helpers.js
@@ -7,16 +7,20 @@ import {UserTypes} from 'action_types';
 import {logError} from './errors';
 const HTTP_UNAUTHORIZED = 401;
 
-export async function forceLogoutIfNecessary(err, dispatch) {
-    if (err.status_code === HTTP_UNAUTHORIZED && err.url.indexOf('/login') === -1) {
-        dispatch({type: UserTypes.LOGOUT_REQUEST});
-        try {
-            await Client4.logout();
-        } catch (error) {
-            logError(error)(dispatch);
+export function forceLogoutIfNecessary(err, dispatch, getState) {
+    // Wrap around a setTimeout so it gets executed once the store updates
+    setTimeout(() => {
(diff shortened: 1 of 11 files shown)


src/actions/helpers.js
```diff
@@ -7,16 +7,20 @@ import {UserTypes} from 'action_types';
...
 
-export async function forceLogoutIfNecessary(err, dispatch) {
-    if (err.status_code === HTTP_UNAUTHORIZED && err.url.indexOf('/login') === -1) {
-        dispatch({type: UserTypes.LOGOUT_REQUEST});
-        try {
-            await Client4.logout();
-        } catch (error) {
-            logError(error)(dispatch);
+export function forceLogoutIfNecessary(err, dispatch, getState) {
+    // Wrap around a setTimeout so it gets executed once the store updates
```
hmhealey: Why do we need to wait for the store to update? I really don't like the look of a setTimeout here since it's not just a UI delay that we're compensating for.

jwilander: I'm with HH on this

enahum: If we don’t currentUserId will still be set for a batch of actions thus logout will be performed more than once.

SetTineout is not UI related. Remember that JS is single threaded and with this i’m just saying that this function should be ran after the current message queue

hmhealey: What does logging out with the client actually do in this case? The session has already expired, so do we even need to call `Client4.logout()` call, or is clearing the token/serverVersion enough?

enahum: Actually I think you are totally right, we shouldn’t be calling Client4.logout as the session already expired, we just need to clear the data in the store by dispatching the right type, I’ll do that ;)

enahum:
@hmhealey @jwilander done, I've removed the `Client4.logout` request as is actually not needed here

PR #362, opened 2018-01-17
Logout even if we cannot reach the server

enahum:
#### Summary
If we cannot reach the server when the user is trying to logout, log the user out anyway.

Also in this PR I've added a couple of actions to be able to clear the config and license data.

#### Ticket Link
https://mattermost.atlassian.net/browse/ICU-388


src/actions/users.js
```diff
@@ -230,12 +230,18 @@ export function loadMe() {
...
-        UserTypes.LOGOUT_REQUEST,
-        UserTypes.LOGOUT_SUCCESS,
-        UserTypes.LOGOUT_FAILURE,
-    );
+    return async (dispatch, getState) => {
+        dispatch({type: UserTypes.LOGOUT_REQUEST}, getState);
+
+        try {
+            Client4.logout();
+            Client.setToken('');
```
hmhealey: This could be in a finally statement or just go after the try/catch

src/actions/users.js
```diff
@@ -230,12 +230,18 @@ export function loadMe() {
...
-        Client4.logout,
-        UserTypes.LOGOUT_REQUEST,
-        UserTypes.LOGOUT_SUCCESS,
-        UserTypes.LOGOUT_FAILURE,
-    );
+    return async (dispatch, getState) => {
+        dispatch({type: UserTypes.LOGOUT_REQUEST}, getState);
+
+        try {
+            Client4.logout();
```
hmhealey: This needs an await since it's asynchronous

enahum: Actually I don’t think we need to wait for it to complete, we really don’t care do we?

PR #366, opened 2018-01-19
More robust favorite channel selection

jespino:
#### Summary
Not add favorite channel when the channel is not accessible (It happens time to time because a race condition).


src/selectors/entities/channels.js
```diff
@@ -463,6 +463,9 @@ export const getSortedFavoriteChannelWithUnreadsIds = createIdsSelector(
             if (!myMembers[id]) {
                 return false;
             }
+            if (!channels[id]) {
```
enahum: perhaps you should do
```js
if (!channels[id] || !myMembers[id]) {
    return false;
}
```

jespino: Done!

PR #368, opened 2018-01-20
ICU-457 Made unread message changes relative to prevent race conditions

hmhealey:
As I discussed with Elias earlier, I'm almost certain now that the issue with the teams not getting marked unread on mobile was because the action we were dispatching to mark the channel as read was saying "set the message count to the current total", but by the time it finished, the total had changed, leaving the team marked as unread while the channel was not. By using relative adjustments, we're making sure that the channelMember unread count is always synced up with the teamMember unread count.

#### Ticket Link
https://mattermost.atlassian.net/browse/ICU-457

#### Checklist


src/actions/channels.js
```diff
@@ -980,99 +980,73 @@ export function updateChannelPurpose(channelId, purpose) {
...
-        const prevChannelMember = channelState.myMembers[prevChannelId]; // May also be null
-
+    return (dispatch, getState) => {
         // Send channel last viewed at to the server
         if (updateLastViewedAt) {
             dispatch({type: ChannelTypes.UPDATE_LAST_VIEWED_REQUEST}, getState);
 
-            Client4.viewMyChannel(channelId, prevChannelId).catch((error) => {
+            Client4.viewMyChannel(channelId, prevChannelId).then(() => {
+                dispatch({type: ChannelTypes.UPDATE_LAST_VIEWED_SUCCESS}, getState);
```
enahum: why are we dispatching this here? won't this mess with the `NEW_MESSAGE_INDICATOR`?

hmhealey: I don't know. I didn't test it with that.

I don't think this ever worked correctly though. It didn't actually wait on `Client4.viewMyChannel` before, so it would always report that the request succeeded, and it would dispatch that as soon as this function ended. The new version just changes it so that the request is set successful at the proper time

enahum: I got confused with `markChannelAsViewed` sorry, this looks good to me

PR #377, opened 2018-01-25
Added script for testing redux changes in mobile project

stephenkiers:
This is a simple addition of a script so that I can modify redux files and see the changes immediately in the mobile project.


stephenkiers:
And should I document this anywhere specific? Should I update the Readme.md? https://github.com/mattermost/mattermost-redux#how-to-build-mattermost-redux

package.json
```diff
@@ -51,6 +51,7 @@
     "webpack": "webpack",
     "dev": "babel src --out-dir ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux --source-maps",
     "dev:watch": "babel --watch src --out-dir ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux --source-maps",
+    "dev-mobile:watch": "babel --watch src --out-dir ${WEBAPP_DIR:-../mattermost-mobile}/node_modules/mattermost-redux --source-maps",
```
enahum: can we make this `MOBILE_DIR` | `RN_DIR` ? I'm 0/5 on the name but it should definitely be different than `WEBAPP_DIR`

stephenkiers: right. I copy/pasted and didn't realize that it shouldn't use the same var. Good catch

stephenkiers:
@enahum how do we test package.json scripts?

README.md
```diff
@@ -169,13 +169,32 @@ client.login(username, password)
...
+If you have mattermost-mobile in other directory or you are developing your own 
+application, you can define the environment variable `MOBILE_DIR` to change the 
+destination app
+(e. g. `MOBILE_DIR=/tmp/mattermost-mobile`).
+
+#### Resetting apps to use package redux
+If you want to go back to using the package specified redux in your web or mobile
+app you can stop the server and run `rm -rf node_modules/mattermost-redux` to force
+your project to reset to the specified package version on next server start.  
+
```
stephenkiers: Is this the best way to reset the webapp/mobile apps? Does it read clearly?

enahum: I don’t think that part is actually true, maybe it’ll be best to also delete .yarninstall

enahum:
I would say we test them by running them :p

stephenkiers:
They work for me...

PR #378, opened 2018-01-25
Migrate from use npm to yarn

jespino:


Makefile
```diff
@@ -1,30 +1,30 @@
...
+	@if ! [ $(shell which yarn) ]; then \
+		echo "yarn is not installed"; \
 		exit 1; \
 	fi
 
-	@echo Getting dependencies using npm
+	@echo Getting dependencies using yarn
 
-	npm install --ignore-scripts --no-package-lock
+	yarn install --ignore-scripts --no-package-lock
```
enahum: does yarn has a `--ignore-scripts` flag?

do we really need to set `--no-package-lock` ?

jespino: Yes, and I don't know.

enahum: can you please test it and then report back?

jespino: Yes sorry, I tested look over the help too fast and my mind made the match --no-lockfile with --no-package-lock automatically :(

`--pure-lockfile` sounds good to me

PR #379, opened 2018-01-26
ABC-93 Load needed custom emojis for posts

jwilander:
#### Summary
When receiving posts, request custom emojis we might be missing, similar to how at-mentions are handled.

Two slightly odd aspects I should point out:
* mattermost-redux needs to know what system emojis are, so this exposes an action to set those. It's not being stored in state or a reducer because they're essentially a constant
* A new reducer `nonExistentEmoji` tracks any emojis we've tried to request but got a 404 response for.  This is needed to by the UI components to discover whether or not to show the original text instead of a placeholder but it also helps cut down on repeating requests to the server for emoji we know don't exist

#### Ticket Link
https://mattermost.atlassian.net/browse/ABC-93

#### Checklist

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -947,7 +947,7 @@ export function getNeededCustomEmojis(state, posts) {
             customEmojisByName = selectCustomEmojisByName(state);
         }
 
-        const pattern = /\B:([A-Za-z0-9_-]+):/gi;
+        const pattern = /\B:([A-Za-z0-9_-]+):\B/gi;
 
         let match;
         while ((match = pattern.exec(post.message)) !== null) {
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -7,13 +7,16 @@ import {Client4} from 'client';
 import {General, Preferences, Posts} from 'constants';
 import {PostTypes, FileTypes} from 'action_types';
 import {getUsersByUsername} from 'selectors/entities/users';
+import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
 
 import * as Selectors from 'selectors/entities/posts';
+import {getConfig} from 'selectors/entities/general';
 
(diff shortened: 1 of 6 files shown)


src/actions/posts.js
```diff
@@ -920,6 +932,47 @@ export function getNeededAtMentionedUsernames(state, posts) {
...
+    Object.values(posts).forEach((post) => {
+        if (!post.message.includes(':')) {
+            return;
+        }
+
+        if (!customEmojisByName) {
+            customEmojisByName = selectCustomEmojisByName(state);
+        }
+
+        const pattern = /\B:([A-Za-z0-9_-]+):/gi;
```
hmhealey: This needs another `\B` at the end to be consistent with the markdown parser so that it doesn't catch things like `:strawberry:sundae`

`/\B:([A-Za-z0-9_-]+):\B/gi`

jwilander:
@hmhealey updated

src/actions/posts.js
```diff
@@ -875,10 +878,19 @@ export async function getProfilesAndStatusesForPosts(posts, dispatch, getState)
...
+    let emojisToLoad;
+    if (getConfig(state).EnableCustomEmoji === 'true') {
+        emojisToLoad = getNeededCustomEmojis(state, posts);
+    }
+
     if (usernamesToLoad.size > 0) {
         promises.push(getProfilesByUsernames(Array.from(usernamesToLoad))(dispatch, getState));
     }
 
+    if (emojisToLoad && emojisToLoad.size > 0) {
```
stephenkiers: We shouldn't need the `emojisToLoad` check in the if statement, as it is declared to exist 6 lines up.

jwilander: It might not get set though, such as when custom emojis are disabled, so we only want to push the promise when it is set

stephenkiers: Of course. My bad, it is declared, not defined. Just testing you. ;)

PR #380, opened 2018-01-27
Initial flow usage

jespino:
#### Summary
Initial flow usage

#### Ticket
[XYZ-65](https://mattermost.atlassian.net/browse/XYZ-65)


CometKim:
It's nice.

Actually I've worked at here: https://github.com/CometKim/mattermost-typed
But I have to consider to deprecate my repo if it mattermost-redux would support flow itself.

jespino:
Interesting. I think we can use a lot of things already defined in your repo, to complete the mattermost-redux flow migration.

jespino:
Rebased to last commit and ready to review.

src/actions/emojis.js
```diff
@@ -43,14 +47,14 @@ export function getCustomEmoji(emojiId) {
...
-export function getCustomEmojiByName(name) {
+export function getCustomEmojiByName(name: string): ActionFunc {
     return async (dispatch, getState) => {
         const serverVersion = Client4.getServerVersion();
         if (!isMinimumServerVersion(serverVersion, 4, 7)) {
             return {data: {}};
         }
 
-        dispatch({type: EmojiTypes.GET_CUSTOM_EMOJI_REQUEST}, getState);
+        dispatch({type: EmojiTypes.GET_CUSTOM_EMOJI_REQUEST, data: null}, getState);
```
lieut-data: I suppose the `data: null` here is required to satisfy Flow? This seems potentially cumbersome if required everywhere. Is it not possible to mark the parameter as optional and allow `undefined`?

src/actions/emojis.js
```diff
@@ -79,13 +83,13 @@ export function getCustomEmojiByName(name) {
             {
                 type: EmojiTypes.GET_CUSTOM_EMOJI_SUCCESS
             }
-        ]));
+        ]), getState);
```
lieut-data: Flow again, I suppose? Does this basically mean optional parameters are always required?

src/actions/emojis.js
```diff
@@ -94,17 +98,18 @@ export function getCustomEmojisByName(names) {
         const promises = [];
         names.forEach((name) => promises.push(getCustomEmojiByName(name)(dispatch, getState)));
 
-        return Promise.all(promises);
+        await Promise.all(promises);
+        return {data: true};
```
lieut-data: Can you elaborate on what's changing here exactly -- I don't immediately follow.

src/actions/emojis.js
```diff
@@ -94,17 +98,18 @@ export function getCustomEmojisByName(names) {
...
 
-export function getCustomEmojisInText(text) {
+export function getCustomEmojisInText(text: string): ActionFunc {
     return async (dispatch, getState) => {
         if (!text) {
             return {data: true};
         }
 
-        const state = getState();
+        const state: GlobalState = getState();
```
lieut-data: Why `state: GlobalState` vs. declaring the parameter type on line 107?

src/actions/emojis.js
```diff
@@ -133,7 +143,8 @@ export function getCustomEmojis(page = 0, perPage = General.PAGE_SIZE_DEFAULT, s
 
         if (loadUsers) {
             const usersToLoad = {};
-            Object.values(data).forEach((emoji) => {
+            Object.keys(data).forEach((key: string) => {
```
lieut-data: Is this a performance-related change or something Flow imposes/makes harder?

src/actions/emojis.js
jespino: Yes, we can support optional data. But I'm not sure about it, I like the idea of explicitly data equal null, declared. But Flow supports optional fields.

src/actions/emojis.js
jespino: The reason is the expected return. of the function, defined in ActionFunc.

src/actions/emojis.js
jespino: No, we can make parameters optional, I didn't done it, but this is a good catch. :)

src/actions/emojis.js
jespino: I think this is not needed, because getState is actually defined as part of the ActionFunc definition. So I can remove it :)

src/actions/emojis.js
jespino: In this case is flow related, flow have problem to infer Object.values result data type, so this is a workaround.

src/actions/emojis.js
jespino: Changed! Thanks! :)

src/actions/emojis.js
jespino: Changed! Thanks :)

src/actions/emojis.js
jespino: Anyway if we declare the data as optional here, we have to check in every single reducer the data existence before we use it. In the future, we can define for each action a type and make it very robust, and allow to only pass data when needed.

jespino:
Rebased

package.json
```diff
@@ -47,10 +52,12 @@
...
   "scripts": {
-    "build": "babel src --out-dir .",
+    "build": "babel src --out-dir . && flow-copy-source src .",
     "webpack": "webpack",
-    "dev": "babel src --out-dir ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux --source-maps",
-    "dev-mobile": "babel src --out-dir ${MOBILE_DIR:-../mattermost-mobile}/node_modules/mattermost-redux --source-maps",
+    "dev": "babel src --out-dir ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux --source-maps && yarn dev:flow-copy",
+    "dev:flow-copy": "flow-copy-source src ${WEBAPP_DIR:-../mattermost-webapp}/node_modules/mattermost-redux",
+    "dev-mobile": "babel src --out-dir ${MOBILE_DIR:-../mattermost-mobile}/node_modules/mattermost-redux --source-maps && yarn dev-mobile:flow-copy",
+    "dev-mobile:flow-copy": "flow-copy-source src ${MOBILE_DIR:-../mattermost-webapp}/node_modules/mattermost-redux",
```
jwilander: What's the purpose of the `:flow-copy` commands? Non-blocking comment, just curious

cometkim: the purpose is same with `.d.ts` of TypeScript.

flow type annotations will be removed after transpile. so it makes sense copying sources to another directory before transpile to use types in other projects which uses mattermost-redux.

src/actions/emojis.js
mkraft: @jespino Wouldn't we need to do a `null` check anyhow?

jespino:
I going to rebase this (@mkraft probably you will need to reset hard and cherry-pick your change)

mkraft:
@jespino Sounds good, thanks for the heads up.

PR #358, opened 2018-01-10
PLT-8532: Select correctly the last user post in a channel

jespino:
#### Summary
Select correctly the last user post (system and ephimeral messages are ignored)

#### Ticket Link
[PLT-8532](https://mattermost.atlassian.net/browse/PLT-8532)

#### Test Information
This PR was tested on: Linux Firefox 57.0.3

PR #381, opened 2018-01-28
PLT-8532: Select correctly the last user post in a channel (#358)

jespino:
#### Summary
This commit was applied into the webapp-4.6 branch (here #358), but never in the master
branch (maybe there are other commits with the same problem, generating
regressions).

#### Ticket Link
[PLT-8532](https://mattermost.atlassian.net/browse/PLT-8532)


jwilander:
I'll check and make sure any thing missed gets merged in

jespino:
I think this one is the only one not merged in the webapp-4.6 branch, so looks like it's right now :)

PR #384, opened 2018-01-30
ABC-79: add autocomplete channels endpoint

ccbrown:
#### Summary
See https://github.com/mattermost/mattermost-server/pull/8163

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -808,6 +808,37 @@ export function getChannels(teamId, page = 0, perPage = General.CHANNELS_CHUNK_S
     };
 }
 
+export function autocompleteChannels(teamId, term) {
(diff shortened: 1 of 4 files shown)


src/actions/channels.js
```diff
@@ -808,6 +808,37 @@ export function getChannels(teamId, page = 0, perPage = General.CHANNELS_CHUNK_S
     };
 }
 
+export function autocompleteChannels(teamId, term) {
```
jwilander: Can you add a unit test for this action?

ccbrown: 💩 You caught me. I went to copy and paste the "searchChannels" unit test, saw that there wasn't one, and thought "This looks hard...".

ccbrown:
@jwilander Added test.

PR #386, opened 2018-01-31
Hid join/leave team messages based on user settings

hmhealey:
I feel like there was a ticket for this, but I can't find it now. The messages for joining/leaving/being invited to/being removed from a team are always shown even if you turn off join/leave team messages in your settings.


hmhealey:
Fixes https://mattermost.atlassian.net/browse/ICU-658

PR #367, opened 2018-01-19
ABC-96 Add new functionality for custom emojis

jwilander:
#### Summary
* Add sorting for get custom emojis actions
* Add actions for searching and autocompleting custom emojis

DO NOT MERGE UNTIL AFTER mattermost/mattermost-server#8125

#### Ticket Link
Part of https://mattermost.atlassian.net/browse/ABC-96

#### Checklist
- [x] Added or updated unit tests (required for all new features)
- [x] All new/modified APIs include changes to the drivers

PR #388, opened 2018-02-03
MM-9274 Added sort=‘status’ to querystring

stephenkiers:
#### Summary
Added sort='status' to lookups for Users in Channel for listing first 100 users in a channel

#### Ticket Link
https://mattermost.atlassian.net/browse/ICU-668

#### Related PRs
https://github.com/mattermost/mattermost-server/pull/8181
https://github.com/mattermost/mattermost-api-reference/pull/329
https://github.com/mattermost/mattermost-webapp/pull/747

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -7,6 +7,7 @@ import {General} from 'constants';
 const FormData = require('form-data');
 
 import fetch from './fetch_etag';
+import {isMinimumServerVersion} from 'src/utils/helpers';
 
 const HEADER_TOKEN = 'Token';
 const HEADER_AUTH = 'Authorization';
@@ -523,8 +524,16 @@ export default class Client4 {
 
     getProfilesInChannel = async (channelId, page = 0, perPage = PER_PAGE_DEFAULT) => {
         this.trackEvent('api', 'api_profiles_get_in_channel', {channel_id: channelId});
+
+        const serverVersion = Client4.getServerVersion();
+        let queryStringObj;
+        if (isMinimumServerVersion(serverVersion, 4, 7)) {
+            queryStringObj = {in_channel: channelId, page, per_page: perPage, sort: 'status'};
+        } else {
+            queryStringObj = {in_channel: channelId, page, per_page: perPage};
+        }
         return this.doFetch(
-            `${this.getUsersRoute()}${buildQueryString({in_channel: channelId, page, per_page: perPage, sort: 'status'})}`,
+            `${this.getUsersRoute()}${buildQueryString(queryStringObj)}`,
             {method: 'get'}
         );
     };
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -17,7 +17,7 @@ const HEADER_USER_AGENT = 'User-Agent';
 const HEADER_X_VERSION_ID = 'X-Version-Id';
 const HEADER_X_CLUSTER_ID = 'X-Cluster-Id';
 
-const PER_PAGE_DEFAULT = 60;
+export const PER_PAGE_DEFAULT = 60;
 const LOGS_PER_PAGE_DEFAULT = 10000;
 
 export default class Client4 {
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -523,9 +523,8 @@ export default class Client4 {
 
     getProfilesInChannel = async (channelId, page = 0, perPage = PER_PAGE_DEFAULT) => {
         this.trackEvent('api', 'api_profiles_get_in_channel', {channel_id: channelId});
-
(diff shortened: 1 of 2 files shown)


src/client/client4.js
```diff
@@ -523,9 +523,8 @@ export default class Client4 {
 
     getProfilesInChannel = async (channelId, page = 0, perPage = PER_PAGE_DEFAULT) => {
         this.trackEvent('api', 'api_profiles_get_in_channel', {channel_id: channelId});
-
         return this.doFetch(
-            `${this.getUsersRoute()}${buildQueryString({in_channel: channelId, page, per_page: perPage})}`,
+            `${this.getUsersRoute()}${buildQueryString({in_channel: channelId, page, per_page: perPage, sort: 'status'})}`,
```
enahum: Why isn't this optional? what happens to those that are using an older server?

stephenkiers: What do you mean by optional?
Concerning supporting other server versions, I assume the server and the webapp would be up to date with each other; however I now realize that would screw up the mobile app.

So, how do I detect the server version to decide if this attribute is included or not? Or what is the strategy for backwards compatibility here?

src/client/client4.js
```diff
@@ -523,8 +524,16 @@ export default class Client4 {
 
     getProfilesInChannel = async (channelId, page = 0, perPage = PER_PAGE_DEFAULT) => {
         this.trackEvent('api', 'api_profiles_get_in_channel', {channel_id: channelId});
+
+        const serverVersion = Client4.getServerVersion();
+        let queryStringObj;
```
enahum: I think you want to do `this.getServerVersion()`

src/client/client4.js
```diff
@@ -524,8 +525,15 @@ export default class Client4 {
     getProfilesInChannel = async (channelId, page = 0, perPage = PER_PAGE_DEFAULT) => {
         this.trackEvent('api', 'api_profiles_get_in_channel', {channel_id: channelId});
 
+        const serverVersion = Client4.getServerVersion();
+        let queryStringObj;
+        if (isMinimumServerVersion(serverVersion, 4, 7)) {
+            queryStringObj = {in_channel: channelId, page, per_page: perPage, sort: 'status'};
+        } else {
```
enahum: You could create this object once and add the status property if needed

src/client/client4.js
jwilander: Hold up, we shouldn't be doing any version checking in the client. Couple things:
1. The client is a 'dumb' web utility and shouldn't do any 'smart' casing like this
2. This will break backwards compatibility the same way changing the default behaviour on the server would. If I upgrade to 4.7 server and start using this client in my integration, all of a sudden the sorting on the results when I use `getProfilesInChannel` is different. That's a breaking change and it's very important we keep those out of the client web utilities

The proper way to do this is to add another argument `sort` and default it to an empty string. Then in your action you do the same thing, add another argument `sort` and default it to whatever makes sense (pretty sure it should still be a blank string though as that's the 99% case and sorting by status is the 1% special case)

See how I added sort to getting emojis for example: 
https://github.com/mattermost/mattermost-redux/pull/367/files#diff-6655fe9b9bb92382c9e8084066aa895eR1702

Notice I didn't add the sort option to the action in that PR because I wanted the default case to sort by name because it made sense in that context

jwilander: P.S. I don't think you'll need to do any backwards compatibility version checking anywhere as even if sort is provided in the query string to an old server, it will ignore it and just do the old behaviour

stephenkiers: Talked this one over with @jwilander and decided we do need to keep this check here because the server currently explicitly rejects requests with the `sort` attribute. Therefore we can only add this attribute on updated servers.

src/client/client4.js
stephenkiers: I could, but I my thought was that, since this code should be deleted in the future, this would make it easier for whomever deletes the check in the future to immediately see it can simply be deleted.

stephenkiers: They really have to just switch from `let` => `const` and delete a code block.

jwilander:
@stephenkiers is this ready for review? I think it still needs the change to the action that calls the `Client4.getProfilesInChannel()` function if it's going to work with your webapp PR

(Sorry if you still need to make changes and I'm jumping the gun)

stephenkiers:
I am still waiting for server changes to be complete. I figured the merge order would be server => redux => webapp (after updating redux pkg) => documentation

hmhealey:
@jwilander Does this still need any changes or is it ready for review? I assume this has missed 4.8, so it would be nice to get it merged for 4.9 shortly

PR #389, opened 2018-02-05
Emoji backwards compatibility with pre 4.7 servers

jwilander:
#### Summary
Emoji backwards compatibility with pre 4.7 servers.

#### Checklist

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -40,24 +40,11 @@ export function isMinimumServerVersion(currentVersion, minMajorVersion = 0, minM
         return false;
     }
 
-    let firstDotIndex = currentVersion.indexOf('.');
-    if (firstDotIndex < 0) {
-        firstDotIndex = currentVersion.length;
-    }
+    const split = currentVersion.split('.');
 
-    let secondDotIndex = currentVersion.indexOf('.', firstDotIndex + 1);
-    if (secondDotIndex < 0) {
-        secondDotIndex = currentVersion.length;
-    }
-
-    let thirdDotIndex = currentVersion.indexOf('.', secondDotIndex + 1) || currentVersion.length;
-    if (thirdDotIndex < 0) {
-        thirdDotIndex = currentVersion.length;
-    }
-
-    const major = parseInt(currentVersion.slice(0, firstDotIndex), 10);
-    const minor = parseInt(currentVersion.slice(firstDotIndex + 1, secondDotIndex) || '0', 10);
-    const dot = parseInt(currentVersion.slice(secondDotIndex + 1, thirdDotIndex) || '0', 10);
+    const major = parseInt(split[0], 10);
+    const minor = parseInt(split[1] || '0', 10);
+    const dot = parseInt(split[2] || '0', 10);
 
     if (major > minMajorVersion) {
         return true;
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -3,7 +3,6 @@
 
 import {createSelectorCreator, defaultMemoize} from 'reselect';
 import shallowEqual from 'shallow-equals';
-import {Client4} from 'client';
 
 export function memoizeResult(func) {
     let lastArgs = null;
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -3,6 +3,7 @@
 
 import {createSelectorCreator, defaultMemoize} from 'reselect';
 import shallowEqual from 'shallow-equals';
+import {Client4} from 'client';
 
 export function memoizeResult(func) {
     let lastArgs = null;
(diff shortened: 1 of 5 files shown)


src/utils/helpers.js
```diff
@@ -29,3 +29,59 @@ export const createIdsSelector = createSelectorCreator(memoizeResult);
...
+// versions, and a non-equal minor version will ignore dot version.
+// currentVersion is a string, e.g '4.6.0'
+// minMajorVersion, minMinorVersion, minDotVersion are integers
+export function isMinimumServerVersion(currentVersion, minMajorVersion = 0, minMinorVersion = 0, minDotVersion = 0) {
+    if (!currentVersion || typeof currentVersion !== 'string') {
+        return false;
+    }
+
+    let firstDotIndex = currentVersion.indexOf('.');
+    if (firstDotIndex < 0) {
```
stephenkiers: I am curious why you don't use `=== -1` here. It is more explicit than <0, although they are pretty much the same thing. I am just used to seeing `=== -1`

src/utils/helpers.js
```diff
@@ -29,3 +29,59 @@ export const createIdsSelector = createSelectorCreator(memoizeResult);
...
+        firstDotIndex = currentVersion.length;
+    }
+
+    let secondDotIndex = currentVersion.indexOf('.', firstDotIndex + 1);
```
stephenkiers: If the firstDotIndex < 0, couldn't we skip doing this check. If there wasn't 1 '.', there won't be two.

src/utils/helpers.js
```diff
@@ -29,3 +29,59 @@ export const createIdsSelector = createSelectorCreator(memoizeResult);
...
+    if (secondDotIndex < 0) {
+        secondDotIndex = currentVersion.length;
+    }
+
+    let thirdDotIndex = currentVersion.indexOf('.', secondDotIndex + 1) || currentVersion.length;
```
stephenkiers: Or three

src/utils/helpers.js
stephenkiers: you could split based on '.' and take the length of the resulting array...? Could be less code?  
    currentVersion.split('.') // ["4","6","0"]

stephenkiers: You can decide, I simply feel like it would be less code and easier to read. Plus you wouldn't have to slice later.

src/utils/helpers.js
```diff
@@ -29,3 +29,59 @@ export const createIdsSelector = createSelectorCreator(memoizeResult);
 
 // Use this selector when you want a shallow comparison of the arguments and you don't need to memoize the result
 export const createShallowSelector = createSelectorCreator(defaultMemoize, shallowEqual);
+
+// isMinimumServerVersion will return true if currentVersion is equal to higher or than the
+// the provided minimum version. A non-equal major version will ignore minor and dot
+// versions, and a non-equal minor version will ignore dot version.
+// currentVersion is a string, e.g '4.6.0'
+// minMajorVersion, minMinorVersion, minDotVersion are integers
+export function isMinimumServerVersion(currentVersion, minMajorVersion = 0, minMinorVersion = 0, minDotVersion = 0) {
```
stephenkiers: we couldn't move getting the current server version into this function? I feel like it would be nice if the function got the server version itself, instead of it being passed in.

src/utils/helpers.js
jwilander: Good suggestion, that's probably a better way to do it

src/utils/helpers.js
jwilander: IMO utility functions shouldn't be pulling data but should have it all passed to them. Easier for testing and for use in cases where the store/whatever might not have the values you need set

jwilander: Essentially it completely decouples them from anything but the caller

src/utils/helpers.js
enahum: Quick reply here, i’m not reviewing the code but instead of indexOf you can use **includes**

src/utils/helpers.js
stephenkiers: okay. If that is the reason, I am good with that. I just think it isn't very DRY having `const serverVersion = Client4.getServerVersion();` so many times; especially since the server shouldn't change between checks. 
Also, is Client4.getServerVersion() memoized?

jwilander: No it's not memoized but it's just a string and there's no calculations going on there so I don't think memoizing it would be useful

jwilander:
@stephenkiers updated the version parsing with your suggestion

src/utils/helpers.js
jwilander: To get around having multiple `const serverVersion = Client4.getServerVersion()` we could have a helper function in the actions folder that wraps the utility but I don't think it's warranted just yet

src/utils/helpers.js
jwilander: I changed it completely anyways but I actually needed the indexes in my original implementation

src/utils/helpers.js
stephenkiers: @jwilander okay, I thought it may be round tripping to the server each time to check. 👍

jwilander: Oh, yeah it's just a string stored as a property on the Client4 class

hmhealey: We could treat this like a selector depending on where we're using it. Like you just pass in state and a version requirement, and it automatically returns whether or not it meets that requirement. 0/5 though

PR #391, opened 2018-02-06
XYZ-71: Uses DefaultTheme from server-side config in some cases.

mkraft:
#### Summary
Uses the `DefaultTheme` from `config.json` if required and present.

#### Ticket Link
[XYZ-71](https://mattermost.atlassian.net/browse/XYZ-71)

#### Checklist
- [x] Added or updated unit tests

#### Test Information
This PR was tested on:
* Chrome browser, macOS

### DIFF
--- a/src/selectors/entities/preferences.js
+++ b/src/selectors/entities/preferences.js
@@ -120,14 +120,23 @@ const getThemePreference = createSelector(
     }
 );
 
-const getDefaultThemeName = createSelector(getConfig, (config) => config.DefaultTheme || 'default');
+const getDefaultTheme = createSelector(getConfig, (config) => {
+    if (config.DefaultTheme) {
+        const theme = Preferences.THEMES[config.DefaultTheme];
+        if (theme) {
+            return theme;
+        }
+    }
+
+    // If no config.DefaultTheme or value doesn't refer to a valid theme name...
+    return Preferences.THEMES.default;
+});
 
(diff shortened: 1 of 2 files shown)


src/selectors/entities/preferences.js
```diff
@@ -120,14 +120,18 @@ const getThemePreference = createSelector(
     }
 );
 
+const getDefaultThemeName = createSelector(getConfig, (config) => config.DefaultTheme || 'default');
```
hmhealey: If you wanted, you could make this `getDefaultTheme` and just have it return `Preferences.THEMES[config.DefaultTheme || 'default']`

src/selectors/entities/preferences.js
```diff
@@ -120,14 +120,18 @@ const getThemePreference = createSelector(
...
+
 export const getTheme = createShallowSelector(
     getThemePreference,
-    (themePreference) => {
+    getDefaultThemeName,
+    (themePreference, defaultThemeName) => {
         let theme;
+        const defaultTheme = Preferences.THEMES[defaultThemeName];
```
hmhealey: We need to make sure this theme exists or else the app will crash on an invalid DefaultTheme. If you actually change the selector to `getDefaultTheme` like I mentioned above, you could do the check inside the selector so that it's safe if it gets used elsewhere

mkraft:
@hmhealey Good idea. Pushed a change and related test.

PR #393, opened 2018-02-06
added getTeamByName selector

stephenkiers:
#### Related PR

https://github.com/mattermost/mattermost-webapp/pull/669

### DIFF
--- a/src/selectors/entities/teams.js
+++ b/src/selectors/entities/teams.js
@@ -12,9 +12,13 @@ export function getCurrentTeamId(state) {
     return state.entities.teams.currentTeamId;
 }
 
-export function getTeamByName(state, name) {
-    return Object.values(getTeams(state)).find((team) => team.name === name);
-}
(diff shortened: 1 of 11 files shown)


src/selectors/entities/teams.js
```diff
@@ -12,6 +12,10 @@ export function getCurrentTeamId(state) {
     return state.entities.teams.currentTeamId;
 }
 
+export function getTeamByName(state, name) {
```
enahum: can you change this to be a selector??
```js
export const getTeamByName = createSelector(
    getTeams,
    (state, name) => name,
    (teams, name) => {
        return Object.values(teams).find((team) => team.name === name);
    }
);
```

that way it gets memoized

stephenkiers:
This needs to be merged and updated on webapp:master asap as webapp has a bug without it.

jwilander:
Maybe DM @enahum if you need a quick review

PR #396, opened 2018-02-07
Verify nonExistentEmoji is a set

enahum:
#### Summary
When using the mobile app without blacklisting the reducer for nonExistentEmoji it gets converted into an array instead of a Set and then when the action gets called it crashes the app, this PR makes sure that `nonExistentEmoji` is always a Set

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -401,7 +401,14 @@ export function getReactionsForPost(postId) {
         }
 
         if (reactions && reactions.length > 0) {
-            const nonExistentEmoji = getState().entities.emojis.nonExistentEmoji;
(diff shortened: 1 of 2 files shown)


enahum:
@hmhealey I could try and implement a transform for it in the mobile app to see if that works, but this way we ensure that even without a transform it will work in any client

hmhealey:
I'd prefer to fix the transform. The problem with doing it this way is that we should then theoretically add this check to anywhere we access one of the Sets which would add a lot of unnecessary code. The problem here is with the mobile app's persist code messing up the state, not something that needs to be fixed in the redux library

PR #397, opened 2018-02-08
Reducing temporary the version to use new permissions to 4.6.0

jespino:
This is temporary for development, when it code go into master must be using the 4.8.0 version as "cut" version for new permissions.


jespino:
Sure!

grundleborg:
@jespino it's an ESlint error that's making this fail on Jenkins.

PR #398, opened 2018-02-08
Fix roles loading

jespino:
#### Summary
Correctly store pending roles to ensure correctly loading after version is available.

#### Test Information
This PR was tested on: [Firefox 58, Arch Linux]

### DIFF
--- a/src/actions/roles.js
+++ b/src/actions/roles.js
@@ -65,12 +65,12 @@ export function loadRolesIfNeeded(roles) {
             pendingRoles.add(role);
         }
         if (!state.entities.general.serverVersion) {
-            setPendingRoles(pendingRoles);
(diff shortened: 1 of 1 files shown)


src/actions/roles.js
```diff
@@ -65,12 +65,12 @@ export function loadRolesIfNeeded(roles) {
             pendingRoles.add(role);
         }
         if (!state.entities.general.serverVersion) {
-            setPendingRoles(pendingRoles);
+            setPendingRoles(pendingRoles)(dispatch, getState);
```
enahum: I know we already merged this but as a reference for the future this can be done like

```js
dispatch(setPendingRoles(pendingRoles));
```

jespino: thanks

PR #401, opened 2018-02-12
Track pending posts for removal

enahum:
#### Summary
In the mobile app we were using the `canTimeout` to track the effect of the optimistic action when creating new posts but this was causing some false/positives and then the post was created twice.

I'm removing the option to handle the timeouts and letting it fail if it actually fails without introducing a race condition by tracking the pending post ids.

When a new post is created we will add the pending post id to the list until the post is actually created.

In case we do not have a response from the server while the post is in a pending state and the app is sent to the background the pending post is cleared from the store and if in fact the post was created on the server it will be received with the next get posts request.

#### Ticket Link
https://mattermost.atlassian.net/browse/ICU-680

mobile-PR: https://github.com/mattermost/mattermost-mobile/pull/1430

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -443,6 +443,7 @@ export default function(state = {}, action) {
         // Object mapping post ids to post objects
         posts,
 
+        // Array that contains the pending post ids for those messages that are in transition to being created
         pendingPostIds: handlePendingPosts(state.pendingPostIds, action),
 
         // Object mapping channel ids to an array of posts ids in that channel with the most recent post first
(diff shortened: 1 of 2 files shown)


src/reducers/entities/posts.js
```diff
@@ -411,7 +443,9 @@ export default function(state = {}, action) {
         // Object mapping post ids to post objects
         posts,
 
-        // Object mapping channel ids to an list of posts ids in that channel with the most recent post first
+        pendingPostIds: handlePendingPosts(state.pendingPostIds, action),
```
jwilander: Maybe just add a quick comment saying what this is like the other reducers

enahum: @jwilander done

PR #402, opened 2018-02-13
Advanced permissions phase 1

mkraft:
This is the PR to merge the "Advanced Permissions Phase 1" feature branch into master.

All the individual commits to this branch have already received a full PR review, and the branch as a whole has undergone a detailed QA pass on a spinmint, so only minimal review should be required at this stage.


src/selectors/entities/general.js
```diff
@@ -22,6 +23,13 @@ export function getCurrentUrl(state: GlobalState): string {
     return state.entities.general.credentials.url;
 }
 
+export function hasNewPermissions(state) {
+    const version = state.entities.general.serverVersion;
+
+    // FIXME This must be changed to 4, 8, 0 before we merge this branch into master
+    return isMinimumServerVersion(version, 4, 6, 0);
```
jwilander: This should become `4, 9, 0` right?

src/actions/channels.js
```diff
@@ -79,6 +80,7 @@ export function createChannel(channel, userId) {
 
         if (!myMembers[created.id]) {
             actions.push({type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER, data: member});
+            loadRolesIfNeeded(new Set(member.roles.split(' ')))(dispatch, getState);
```
enahum: Can we change these to 
`dispatch(loadRolesIfNeeded(new Set(member.roles.split(' '))));`

That way we don’t care if the action is a thunk or not

src/actions/roles.js
```diff
@@ -0,0 +1,93 @@
...
+            setPendingRoles(pendingRoles)(dispatch, getState);
+            return {data: []};
+        }
+        if (!hasNewPermissions(state)) {
+            if (state.entities.roles.pending) {
+                await setPendingRoles(new Set())(dispatch, getState);
+            }
+            return {data: []};
+        }
+        const loadedRoles = getRoles(state);
```
enahum: Don’t you need to await on getRoles?

src/actions/users.js
```diff
@@ -197,6 +198,19 @@ function completeLogin(data) {
             },
         ]), getState);
 
+        const roles = new Set();
```
enahum: I keep seeing this pattern over and over again, can we create an util function for this?

src/actions/websocket.js
```diff
@@ -475,6 +485,33 @@ function handleUserUpdatedEvent(msg, dispatch, getState) {
     }
 }
 
+function handleRoleAddedEvent(msg, dispatch) {
+    const role = JSON.parse(msg.data.role);
```
enahum: What if JSON.parse throws an error?

src/store/initial_state.js
```diff
@@ -89,6 +89,10 @@ const state: GlobalState = {
             recent: {},
         },
         typing: {},
+        roles: {
+            roles: [],
```
hmhealey: This should be an object, not an array

src/utils/channel_utils.js
```diff
@@ -218,7 +220,16 @@ export function isGroupOrDirectChannelVisible(channel, memberships, config, myPr
     return isDirectChannelVisible(users[otherUserId] || otherUserId, config, myPreferences, channel, lastPost, isUnreadChannel(memberships, channel));
 }
 
-export function showCreateOption(config, license, channelType, isAdmin, isSystemAdmin) {
+export function showCreateOption(state, config, license, teamId, channelType, isAdmin, isSystemAdmin) {
```
hmhealey: Do we need to pass in most of these when we just have the state?

hmhealey: 0/5 on that. Changing this will require rewriting all the unit tests that have been added, so it's not likely not worth it at this point

src/selectors/entities/roles.js
```diff
@@ -0,0 +1,212 @@
...
+                if (teamsMemberships.hasOwnProperty(key)) {
+                    roles[key] = new Set(teamsMemberships[key].roles.split(' '));
+                }
+            }
+        }
+        return roles;
+    }
+);
+
+export const getMyChannelRoles = createSelector(
```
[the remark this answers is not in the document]
hmhealey: That's a really big change though, so we can wait to see if it is an issue later on

src/selectors/entities/general.js
jespino: Yes

src/actions/roles.js
jespino: getRoles is not an action, is a selector.

src/actions/channels.js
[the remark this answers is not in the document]
jespino: We want to avoid dupplications to not ask multiple times for the same roles

src/selectors/entities/roles.js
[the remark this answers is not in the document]
jespino: Yes, it is a big change, and we can do that later.

src/store/initial_state.js
jespino: You are right.

src/utils/channel_utils.js
jespino: Yes, probably we can rewrite it later.

src/actions/channels.js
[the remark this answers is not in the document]
hmhealey: loadRolesIfNeeded already puts everything into a Set. We might want to make that clear with a comment or something though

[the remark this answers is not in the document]
jespino: Yeah, you are right, we can remove the Set creation before each loadRolesIfNeeded

src/store/initial_state.js
grundleborg: Ah, good catch, this explains a really minor but odd bug I was seeing on the spinmint!

jespino:
I going to add issues for each proposed change, but If you find any "stopper" to merge it, we can stop the merge, if not, we can merge, and fix the problems later.

src/selectors/entities/channels.js
```diff
@@ -363,7 +364,26 @@ export const canManageChannelMembers = createSelector(
     getMyCurrentChannelMembership,
     getConfig,
     getLicense,
-    canManageMembers
+    hasNewPermissions,
+    (state) => haveICurrentChannelPermission(state, {permission: Permissions.MANAGE_PRIVATE_CHANNEL_MEMBERS}),
```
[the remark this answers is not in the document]
jespino: I will create an issue to improve this.

src/selectors/entities/roles.js
```diff
@@ -0,0 +1,212 @@
+// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+import {createSelector} from 'reselect';
+
+import {getCurrentUser, getCurrentChannelId} from 'selectors/entities/common';
+import {getTeamMemberships, getCurrentTeamId} from 'selectors/entities/teams';
+
+export const getMySystemRoles = createSelector(
+    getCurrentUser,
```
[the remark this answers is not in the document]
jespino: I'll create an issue to fix this

src/store/initial_state.js
jespino: Fixed

src/actions/channels.js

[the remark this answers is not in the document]
jespino: Solved in #447

PR #408, opened 2018-02-19
Fix reducer for permalink and search

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-9374

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -508,7 +508,7 @@ export function getPostThread(postId, skipAddToChannel = true) {
             {
                 type: PostTypes.GET_POST_THREAD_SUCCESS
             }
-        ]), getState);
+        ], 'BATCH_GET_POST_THREAD'), getState);
 
         return {data: posts};
     };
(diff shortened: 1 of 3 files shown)


src/actions/posts.js
```diff
@@ -508,7 +508,7 @@ export function getPostThread(postId, skipAddToChannel = true) {
             {
                 type: PostTypes.GET_POST_THREAD_SUCCESS
             }
-        ]), getState);
+        ], 'BATCH_GET_POST_THREAD'), getState);
```
jespino: I don't know exactly why this change. Is simply for be more explicit about the action type generated, for redux-devtools and for code clarity, or is there any place where we use this information?

enahum: This is mainly to use while debugging as is easier to find the action with all the `BATCH` actions everywhere, it does not have any implications

jespino: Sounds good :)

saturninoabril: @enahum Thanks for the info. I also learned from this :)

PR #415, opened 2018-02-22
MM-9289 Stopped showing deleted teams as joinable

hmhealey:
The only really important change here is that we filter out teams with `delete_at === 0`

#### Checklist


jwilander:
Needs a rebase @hmhealey

PR #418, opened 2018-02-27
MM-8604: process config/license websocket events

lieut-data:
#### Summary
Respond to the new `config_changed` and `licensed_changed` websocket events to update the Redux store. Note that this only impacts the mobile app, since the webapp has its own websocket processing logic. (Corresponding changes to the webapp are are forthcoming.)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-8604

#### Checklist

#### Test Information
This PR was tested on: iPhone 6 (Simulator)

### DIFF
--- a/src/actions/general.js
+++ b/src/actions/general.js
@@ -4,10 +4,8 @@
 import {Client, Client4} from 'client';
 import {bindClientFunc, forceLogoutIfNecessary, FormattedError} from './helpers.js';
 import {GeneralTypes} from 'action_types';
-import {General} from 'constants';
 import {loadMe} from './users';
 import {logError} from './errors';
-import EventEmitter from 'utils/event_emitter';
 import {batchActions} from 'redux-batched-actions';
 
 export function getPing(useV3 = false) {
@@ -145,8 +143,6 @@ export function setDeviceToken(token) {
         return {data: true};
     };
 }
-
-EventEmitter.on(General.CONFIG_CHANGED, setServerVersion);
 
 export function setServerVersion(serverVersion) {
     return async (dispatch, getState) => {
--- a/src/actions/websocket.js
+++ b/src/actions/websocket.js
@@ -23,12 +23,15 @@ import {
     getProfilesAndStatusesForPosts,
     getCustomEmojiForReaction,
 } from './posts';
-
 import {
     getMyPreferences,
     makeDirectChannelVisibleIfNecessary,
     makeGroupMessageVisibleIfNecessary,
 } from './preferences';
+import {
+    getLicenseConfig,
+    getClientConfig,
+} from './general';
 
 import {getTeam, getTeams, getMyTeams, getMyTeamMembers, getMyTeamUnreads} from './teams';
 
@@ -123,6 +126,8 @@ async function handleReconnect(dispatch, getState) {
     const {currentChannelId} = entities.channels;
     const {currentUserId} = entities.users;
 
+    getLicenseConfig()(dispatch, getState);
+    getClientConfig()(dispatch, getState);
     getMyPreferences()(dispatch, getState);
 
     if (currentTeamId) {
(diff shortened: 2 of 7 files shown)


test/actions/websocket.test.js
```diff
@@ -595,4 +595,49 @@ describe('Actions.Websocket', () => {
 
         test();
     });
+
+    it('handle license changed', (done) => {
+        async function test() {
+            if (TestHelper.isLiveServer()) {
+                // No live server version implemented for this test case.
+                this.skip();
```
lieut-data: Running `yarn test-no-mock` is broken for all sorts of tests right now, so I figured I wouldn't spend a lot of time on this. Pinged the developers channel for discussion on this point.

src/actions/general.js
```diff
@@ -146,8 +146,6 @@ export function setDeviceToken(token) {
     };
 }
 
-EventEmitter.on(General.CONFIG_CHANGED, setServerVersion);
```
lieut-data: This never did anything, since calling `setServerVersion` just returned and discarded the thunk. The mobile app handles this event and dispatches on its own accord, and the webapp never used this part of the state tree. I contemplated wiring this up correctly and dispensing with the mobile handling, but deferred for now since not required.

src/actions/websocket.js
```diff
@@ -123,6 +126,8 @@ async function handleReconnect(dispatch, getState) {
     const {currentChannelId} = entities.channels;
     const {currentUserId} = entities.users;
 
+    getLicenseConfig()(dispatch, getState);
```
enahum: 1/5 on this one minor thing, this can be written as `dispatch(getLicenseConfig())` the other ones as well

src/actions/websocket.js
```diff
@@ -602,7 +613,7 @@ function handleHelloEvent(msg) {
     const serverVersion = msg.data.server_version;
     if (serverVersion && Client4.serverVersion !== serverVersion) {
         Client4.serverVersion = serverVersion;
-        EventEmitter.emit(General.CONFIG_CHANGED, serverVersion);
+        EventEmitter.emit(General.SERVER_VERSION_CHANGED, serverVersion);
```
enahum: if you do this change here, then you'll need to make some changes in [mobile](https://github.com/mattermost/mattermost-mobile/blob/master/app/mattermost.js#L82) as well as this was used to request the new config from the server, you have to assume that the websocket is not connected and you already have a config loaded into the offline store.

when the config or the license changed this version used to change as well so with any new request this would trigger that event thus we are requesting the new config from the server.

makes sense?

src/actions/websocket.js
```diff
@@ -671,6 +682,25 @@ function handleAddEmoji(msg, dispatch) {
     });
 }
 
+function handleLicenseChangedEvent(msg, dispatch) {
+    const data = msg.data.license;
+
+    dispatch({
```
enahum: what happens when the config does not change but the license does? same as above look at https://github.com/mattermost/mattermost-mobile/blob/master/app/mattermost.js#L82

src/actions/websocket.js
lieut-data: I had already made some changes to mobile, but was waiting for this PR to update the `yarn.lock`. Of course, reviewing this doesn't make sense without the context of that PR, so I've pushed it up here: https://github.com/mattermost/mattermost-mobile/pull/1469.

I'd love your feedback on the comments in that PR. Effectively, this means /no/ changes in the short term, since we can't assume we're connected to a server which will emit these websockets, at least not without server version parsing or some other signal. It didn't seem like a bad thing to keep the existing refresh alongside the websocket in all cases, at least until we move past supporting these servers.

src/actions/websocket.js
lieut-data: See the linked PR above, but basically the `SERVER_VERSION_CHANGED` implicitly includes changes to the config and license given the need to support older servers. Either way, we will receive discrete websocket events for the license changing even if the config does not change.

src/actions/websocket.js
hmhealey: Doing it that way also prevents us from needing to care if the action is a thunk action (that returns a function) or a regular redux action (that returns a JS object).

PR #421, opened 2018-03-08
Add typing selector based on channelId and postId

jespino:
#### Summary
Adding typing selector using channelId and postId as options parameter (this
will help to remove the user_typing_store from the frontend).

### DIFF
--- a/src/selectors/entities/typing.js
+++ b/src/selectors/entities/typing.js
@@ -27,14 +27,16 @@ const getUsersTypingImpl = (profiles, teammateNameDisplay, channelId, parentPost
     return [];
 };
 
-export const getUsersTypingByChannelAndPost = createSelector(
-    getUsers,
-    getTeammateNameDisplaySetting,
-    (state, options) => options.channelId,
-    (state, options) => options.postId,
-    (state) => state.entities.typing,
-    getUsersTypingImpl,
-);
+export const makeGetUsersTypingByChannelAndPost = () => {
+    return createSelector(
+        getUsers,
+        getTeammateNameDisplaySetting,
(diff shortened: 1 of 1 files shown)


src/selectors/entities/typing.js
```diff
@@ -11,25 +11,36 @@ import {getTeammateNameDisplaySetting} from 'selectors/entities/preferences';
...
+            return users.map((userId) => {
+                return displayUsername(profiles[userId], teammateNameDisplay);
+            });
+        }
+    }
+
+    return [];
+};
+
+export const getUsersTypingByChannelAndPost = createSelector(
```
hmhealey: This should be done as `makeGetUsersTyping` or something along those lines where it returns a selector. Otherwise, you'll be calling `getUsersTypingByChannelAndPost` for both the center channel and RHS with different arguments which will prevent it from being memoized

jespino: Very good catch, thanks! :)

src/selectors/entities/typing.js
```diff
@@ -11,25 +11,38 @@ import {getTeammateNameDisplaySetting} from 'selectors/entities/preferences';
 
 import {displayUsername} from 'utils/user_utils';
 
+const getUsersTypingImpl = (profiles, teammateNameDisplay, channelId, parentPostId, typing) => {
```
saturninoabril: `getUsersTypingImpl` does not sound obvious to me what it means. 0/5 on what name it should be.

jespino: Yes, I haven't a better proposal, and I understood that was clear enough from the context.

PR #423, opened 2018-03-09
Timezone feature

stanchan:
#### Summary
Adds timezone support for both webapp and mobile
* action for getting timezone from server
* selector for getting timezone array in `state.entities.general`
* mock http test for timezone endpoint

#### Checklist


miguelespinoza:
Reference to server PR: https://github.com/mattermost/mattermost-server/pull/8185

src/client/client4.js
```diff
@@ -225,6 +225,10 @@ export default class Client4 {
         return `${this.getPluginsRoute()}/${pluginId}`;
     }
 
+    getTimezonesRoute() {
+        return `${this.getBaseRoute()}/timezone`;
```
jwilander: Slight change, could we make the route `/api/v4/system/timezone` instead of `/api/v4/timezone`? Just since it's part of the system configuration and it would match better with our other endpoints

src/utils/timezone_utils.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return manualTimezone;
+}
+
+export function getTimezoneRegion(timezone) {
+    if (timezone) {
+        const split = timezone.split('/');
+        if (split.length > 1) {
+            const region = split.pop();
+            if (region.indexOf('_') >= 0) {
```
lieut-data: nit: is the check here worth it? Seems like an unconditional `replace` might just be clearer:
```
return split.pop().replace('_', ' ');
```

test/actions/general.test.js
```diff
@@ -117,4 +117,17 @@ describe('Actions.General', () => {
...
+        nock(Client4.getBaseRoute()).
+        get('/system/timezones').
+        query(true).
+        reply(200, ['America/New_York', 'America/Los_Angeles']);
+
+        await Actions.getSupportedTimezones()(store.dispatch, store.getState);
+
+        await TestHelper.wait(100);
+        const {timezones} = store.getState().entities.general;
+        assert.equal(timezones.length > 0, true);
```
lieut-data: Worth doing an equality comparison for that extra bit of sanity when something breaks down the road?

miguelespinoza:
Thanks @jwilander and @lieut-data for the review, this is looking good! 😄 

If we can get this merged it'll help with the webapp PR, the build fails because it lacks these timezone methods

jwilander:
Done, nice work @miguelespinoza !

jwilander:
FYI still need the server PR to merge before the webapp PR becomes testable

PR #424, opened 2018-03-13
Remove temporary post after createPostImmediately fail

jespino:


src/reducers/entities/posts.js
```diff
@@ -31,6 +31,35 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, action) {
...
+    const nextPosts = {
+        ...posts,
+    };
+
+    Reflect.deleteProperty(nextPosts, postId);
+
+    let nextPostsForChannel = postsInChannel;
+
+    // Only change postsInChannel if the order of the posts needs to change
+    if (!postsInChannel[channelId] || postsInChannel[channelId].indexOf(postId) !== -1) {
```
enahum: Can you change indexOf for includes?

src/reducers/entities/posts.js
```diff
@@ -31,6 +31,35 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, action) {
...
+        // If we don't already have the post, assume it's the most recent one
+        const postsForChannel = postsInChannel[channelId] || [];
+
+        nextPostsForChannel = {...postsInChannel};
+        nextPostsForChannel[channelId] = [];
+        for (const post of postsForChannel) {
```
enahum: Is this really necessary? Why not just get the postsInCannel and just remove the post Id if it exists

src/reducers/entities/posts.js
```diff
@@ -103,6 +132,16 @@ function handlePendingPosts(pendingPostIds = [], action) {
...
     }
+    case PostTypes.REMOVE_PENDING_POST: {
+        const postId = action.data.id;
+        const nextPendingPostIds = [];
+        for (const post of pendingPostIds) {
+            if (post !== postId) {
+                nextPendingPostIds.push(post);
+            }
+        }
+        return nextPendingPostIds;
```
enahum: Same here, you don’t need to loop, just remove the post id from the array

src/reducers/entities/posts.js
jespino: I have changed it here, and allover the file.

src/reducers/entities/posts.js
jespino: To remove an element from an array in javascript you have to search and remove it anyway. I changed it with a more compact and concrete way to do it, but it is still a loop.

src/reducers/entities/posts.js
jespino: Same here.

src/reducers/entities/posts.js
```diff
@@ -103,6 +127,11 @@ function handlePendingPosts(pendingPostIds = [], action) {
         }
         return nextPendingPostIds;
     }
+    case PostTypes.REMOVE_PENDING_POST: {
+        const postId = action.data.id;
+        const nextPendingPostIds = pendingPostIds.filter((post) => post !== postId);
```
saturninoabril: Can we change `post` to `pendingPostId` just not to confuse post ID with typical post object?

src/reducers/entities/posts.js
```diff
@@ -31,6 +31,30 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, action) {
     return {posts: nextPosts, postsInChannel: nextPostsForChannel};
 }
 
+function handleRemovePendingPost(posts = {}, postsInChannel = {}, action) {
```
saturninoabril: Might be good to have test for `handleRemovePendingPost`.

src/reducers/entities/posts.js
jespino: You are right :)

src/reducers/entities/posts.js
jespino: You are more than right again :)

jespino: Done! :)

src/reducers/entities/posts.js
jespino: Done :)

saturninoabril: I mean, can we have it something like:
```
const nextPendingPostIds = pendingPostIds.filter((pendingPostId) => pendingPostId !== postId);
```

jespino: Ohh, sorry my misunderstanding. I'll do it too :)

saturninoabril: No worries, I'm also not clear about my request :)

saturninoabril: I'm good with any of your preferred solution.

jespino: Done :)

PR #425, opened 2018-03-13
Add Out Of Office support

csduarte:
#### Summary
Add constants for mobile and webapp support
* OUT_OF_OFFICE = ooo
* SYSTEM_AUTO_RESPONDER = system_auto_response

via @miguelespinoza

### DIFF
--- a/src/constants/general.js
+++ b/src/constants/general.js
@@ -18,6 +18,7 @@ export default {
 
     MENTION: 'mention',
 
+    OUT_OF_OFFICE: 'ooo',
     OFFLINE: 'offline',
     AWAY: 'away',
     ONLINE: 'online',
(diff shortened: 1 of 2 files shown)


src/constants/general.js
```diff
@@ -18,6 +18,7 @@ export default {
 
     MENTION: 'mention',
 
+    OUT_OF_OFFICE: 'ooo',
```
jespino: I don't love the `ooo` value, why not `out_of_office`?

cpanato: +1 for ooo 😄

miguelespinoza: @jespino I followed the convention with DND.  `dnd`.  From my assumption, it seemed like the reasoning for `do_not_disturb` felt inadequate?

Therefore I went with `ooo`.

Also, from a feature standpoint, slash commands follow the same format /dnd, therefore once slash command for Out Of Office is implemented we would use `/ooo`

jespino: ok, makes sense :)

miguelespinoza:
@jespino if this looks good can we merge so clients can make use of the changes.

Thanks! 😄

jespino:
I'm ok with that, but I'm waiting for @hmhealey approval :)

PR #426, opened 2018-03-16
[MM-9516] Add new ephemeral post as EPHEMERAL_ADD_TO_CHANNEL

saturninoabril:
#### Summary
Add new ephemeral post as EPHEMERAL_ADD_TO_CHANNEL
- This is an ephemeral post at Thread of mobile app (or RHS on webapp) after a user has been added thru at-mention.

#### Ticket Link
Jira ticket: [MM-9516](https://mattermost.atlassian.net/browse/MM-9516)

#### Checklist

### DIFF
--- a/src/constants/posts.js
+++ b/src/constants/posts.js
@@ -5,6 +5,7 @@ export const PostTypes = {
     CHANNEL_DELETED: 'system_channel_deleted',
     DISPLAYNAME_CHANGE: 'system_displayname_change',
     EPHEMERAL: 'system_ephemeral',
+    EPHEMERAL_ADD_TO_CHANNEL: 'system_ephemeral_add_to_channel',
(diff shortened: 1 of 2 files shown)


jwilander:
Little confused as to why this is needed, could you provide a bit more clarification? I just ask because we have this feature on the web app and I don't see us using a special type there

saturninoabril:
Primary reason was that mobile app is using most of the constant from `mattermost-redux` such as those in `Posts.POST_TYPES`. I just thought that it may be better to add `EPHEMERAL_ADD_TO_CHANNEL` here. Then since it's here now, I added that to `isPostEphemeral` to centralize the use of that function. We can use it for webapp as well in the near future.

jwilander:
Ok sounds good

saturninoabril:
Cool, thanks!

PR #427, opened 2018-03-19
MM-9801 Performance improvements for large numbers of posts

jwilander:
#### Summary
After some profiling with 5000 posts loaded locally, the major bottleneck was in the `formatPostInChannel` function.

screen shot 2018-03-16 at 14 04 30

We were spending a ton of time in there looping through every single post we had locally to do two things:
* Check if the current user started or replied to the thread of the post in question (needed for notification purposes)
* Count the number of replies in the thread

While debugging, removing this one loop took switching to a channel with 5000 posts from taking ~60 seconds to just ~200ms.

To remove the loop without functionality loss, I added a new reducer `postsInThread`. It is very similar to the existing `postsInChannel` reducer, except instead of tracking posts in channels it was for tracking posts in threads.

Here's some profiling numbers showing the improvements on `formatPostInChannel` from doing that:

| Commit | Self Time | Total Time |
| --- | --- | --- |
| pre-changes | 30,735.6 ms | 94,823.6 ms |
| bc6c865 | 184.4 ms | 206.9 ms |

That reduced total time spent in that function by a factor of more than 450, which is quite the improvement. Even still, `formatPostInChannel` was still the performance bottleneck because it was being run on every single post in the channel, even if those posts weren't going to be used. I fixed this by adding another parameter to the `getPostsInChannel` selector called `numPosts` for specifying how many posts the selector should return. The results of that for time spent in `formatPostInChannel`:

 | Commit | Self Time | Total Time |
| --- | --- | --- |
| bc6c865 | 184.4 ms | 206.9 ms |
| f8ac2fe | 0.9 ms | 13.4 ms |

In addition to improving `formatPostInChannel`, I also went through and fixed every instance where we were looping through all the posts by using the `postsInThread` reducer.

This should be backwards compatible with old usages of the selectors (except for getPostsInChannel).

#### More Potential Improvements
This helped a lot but there is still more room for improvement:
* If a thread has 1000s of posts, we will still loop through them all for each post
* There is still a ton of loop invariant code contained in `formatPostInChannel` that could be pulled out to improve performance
* The selector recalculates very frequently, finding a way to reduce that would help a bunch
* We spend a lot of time sorting, both in the selectors and reducers for posts

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-9801

#### Checklist

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -14,32 +14,32 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         [post.id]: post,
     };
 
-    let nextPostsForChannel = postsInChannel;
+    let nextPostsInChannel = postsInChannel;
 
     // Only change postsInChannel if the order of the posts needs to change
     if (!postsInChannel[channelId] || !postsInChannel[channelId].includes(post.id)) {
         // If we don't already have the post, assume it's the most recent one
         const postsForChannel = postsInChannel[channelId] || [];
 
-        nextPostsForChannel = {...postsInChannel};
-        nextPostsForChannel[channelId] = [
+        nextPostsInChannel = {...postsInChannel};
+        nextPostsInChannel[channelId] = [
             post.id,
             ...postsForChannel,
         ];
     }
 
-    let nextPostsForThread = postsInThread;
+    let nextPostsInThread = postsInThread;
     if (post.root_id && (!postsInThread[post.root_id] || !postsInThread[post.root_id].includes(post.id))) {
         const postsForThread = postsInThread[post.root_id] || [];
 
-        nextPostsForThread = {...postsInThread};
-        nextPostsForThread[post.root_id] = [
+        nextPostsInThread = {...postsInThread};
+        nextPostsInThread[post.root_id] = [
             post.id,
             ...postsForThread,
         ];
     }
 
-    return {posts: nextPosts, postsInChannel: nextPostsForChannel, postsInThread: nextPostsForThread};
+    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
 }
 
 function handleRemovePendingPost(posts = {}, postsInChannel = {}, postsInThread = {}, action) {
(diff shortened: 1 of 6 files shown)


jwilander:
@sudheerDev if you want to give it a review as well that'd be great

src/selectors/entities/posts.js
```diff
@@ -190,10 +207,12 @@ function formatPostInChannel(post, previousPost, index, allPosts, postIds, curre
 export function makeGetPostsInChannel() {
     return createSelector(
         getAllPosts,
+        getPostsInThread,
         (state, channelId) => state.entities.posts.postsInChannel[channelId],
         getCurrentUser,
         getMyPreferences,
-        (allPosts, postIds, currentUser, myPreferences) => {
+        (state, channelId, numPosts) => numPosts || Posts.POST_CHUNK_SIZE,
```
jwilander: This is essentially the `postVisibility` value from the web app. It will change when the post visibility changes (some clicks load more messages, or a post is added to the channel while a user is in that channel) but it won't change very often and will change at the same time as posts changing so the effect is negligible

src/reducers/entities/posts.js
```diff
@@ -28,12 +28,24 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, action) {
         ];
     }
 
-    return {posts: nextPosts, postsInChannel: nextPostsForChannel};
+    let nextPostsForThread = postsInThread;
```
hmhealey: Can you fix this naming? You're mixing up ForThread vs InThread in a few of these functions. It looks like you want postsInThread to be the field in the state and postsForThread to be `postsInThread[rootId]`

src/selectors/entities/posts.js
```diff
@@ -59,19 +63,21 @@ export const getPostsInCurrentChannel = createSelector(
 export function makeGetPostIdsForThread() {
     return createIdsSelector(
         getAllPosts,
-        (state, rootId) => rootId,
-        (posts, rootId) => {
+        (state, rootId) => state.entities.posts.postsInThread[rootId] || [],
+        (state, rootId) => state.entities.posts.posts[rootId],
+        (posts, postsInThread, rootPost) => {
```
hmhealey: Same with this file

src/reducers/entities/posts.js
jwilander: It is a bit mixed up but I directly copied the naming from the `postsInChannel`. I think it's better to match that don't you think?

jwilander: Actually I'll just fix them both

jwilander:
@hmhealey updates made

jwilander:
@enahum added it to initial state and I'll change the middleware when I submit the PR to update redux in the mobile repo

koxen:
is anyone else seeing problems with "load new messages" not working after loading second page?
is it just coincidence that it started happening when I applied this patch?

src/selectors/entities/posts.js
```diff
@@ -284,19 +292,21 @@ export function makeGetPostsAroundPost() {
...
-                if (posts.hasOwnProperty(id)) {
-                    const post = posts[id];
+            if (rootPost) {
+                thread.push(rootPost);
+            }
 
-                    if (id === rootId || post.root_id === rootId) {
-                        thread.push(post);
-                    }
+            postsInThread.forEach((id) => {
```
koxen: why not for (const id in postsInThread)?

src/selectors/entities/posts.js
```diff
@@ -308,20 +318,20 @@ export function makeGetPostsForThread() {
...
-            if (currentPost) {
-                for (const id in posts) {
-                    if (posts.hasOwnProperty(id)) {
-                        const post = posts[id];
-
-                        if (post.root_id === currentPost.id && post.state !== Posts.POST_DELETED && !isPostEphemeral(post)) {
-                            count += 1;
-                        }
-                    }
+            postsInThread.forEach((id) => {
```
koxen: same as above

src/selectors/entities/posts.js
sudheerDev: Any particular reason for favouring `for in` instead of `forEach`? From my personal experience number of things that might screw up `for in` outweigh any benefits, it provides .https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea?lq=1.

koxen: I'd be fine with (for let i = 0; i < arr.length; ...) too, but .forEach syntax is really unnecessary, and makes code harder to read

jwilander: We use forEach all over the place, and personally I prefer it here

jwilander:
@koxen there needs to be webapp changes to support loading more changes, that's why it won't load past the second page for you

That will come in the web app PR to update mattermost-redux

PR #429, opened 2018-03-22
Add selectors to check if current channel and a channel is ReadOnly.

csduarte:
#### Summary
Selectors to check if the current channel isReadOnly or if a channel isReadOnly, byId or by channel passed. In the current logic only TownSquare can be ReadOnly when `ExperimentalTownSquareIsReadOnly=true`.
NOTE: For the ticket to make TownSquare read only it was recommended to move the isReadOnly logic to redux:
https://github.com/mattermost/mattermost-webapp/pull/831

#### Checklist

#### Test Information
This PR was tested on: MacOS HighSierra 10.13.3

via @dmeza

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -20,6 +20,7 @@ import {
 } from 'selectors/entities/preferences';
 import {getLastPostPerChannel} from 'selectors/entities/posts';
 import {getCurrentTeamId, getCurrentTeamMembership} from 'selectors/entities/teams';
+import {isCurrentUserSystemAdmin} from 'selectors/entities/users';
 
 import {
     buildDisplayableChannelList,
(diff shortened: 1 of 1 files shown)


dmeza:
@hmhealey these are the changes requested for the webapp PR: https://github.com/mattermost/mattermost-webapp/pull/831

src/selectors/entities/channels.js
```diff
@@ -136,6 +137,23 @@ export const getCurrentChannelStats = createSelector(
     }
 );
 
+export const isCurrentChannelReadOnly = createSelector(
+    (state) => state,
```
hmhealey: This doesn't work when we pass in the whole state since that's always changing. You could either break this up into
```
export const isCurrentChannelReadOnly = createSelector(
    getCurrentUser,
    getConfig,
    ...
);
```
Or you could probably just get rid of the createSelector since this isn't computationally expensive

dmeza:
@hmhealey rebased to latest from master and made the change to make it a function.

PR #433, opened 2018-03-25
Migrating some files (constants/action_types) to flow and the actions/admin.js

jespino:
#### Summary
Tagged all the constants files and action_types files to flow. That was pretty easy.

Migrated the actions/admin.js to flow, that includes the flow type definitions
for jobs in the global state.

Probably is easier to review each commit independently.

### DIFF
--- a/src/actions/admin.js
+++ b/src/actions/admin.js
@@ -1,5 +1,6 @@
 // Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
 // See License.txt for license information.
+// @flow
 
 import {AdminTypes} from 'action_types';
 import {General} from 'constants';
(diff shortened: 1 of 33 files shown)


src/actions/admin.js
```diff
@@ -118,7 +122,7 @@ export function getComplianceReport(reportId) {
     );
 }
 
-export function getComplianceReports(page = 0, perPage = General.PER_PAGE_DEFAULT) {
+export function getComplianceReports(page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT): ActionFunc {
```
jespino: I think this is a small bug detected by flow (General.PER_PAGE_DEFAULT is undefined), the General.PAGE_SIZE_DEFAULT

PR #434, opened 2018-03-26
PLT-4340 Sort Muted channels

enahum:
#### Summary
This is the redux part to sort the muted channels needed with the `mute` feature

[Mute Channel WebApp PR here](https://github.com/mattermost/mattermost-webapp/pull/196)

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/7617

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -34,6 +34,7 @@ import {
     isDirectChannelVisible,
     isGroupChannelVisible,
     isGroupOrDirectChannelVisible,
+    sortMutedChannels,
 } from 'utils/channel_utils';
 import {createIdsSelector} from 'utils/helpers';
 
@@ -415,7 +416,7 @@ export const getSortedUnreadChannelIds = createIdsSelector(
     getUnreadChannelIds,
     getTeammateNameDisplaySetting,
     (state, lastUnreadChannel = null) => lastUnreadChannel,
-    (currentUser, profiles, channels, members, unreadIds, settings, lastUnreadChannel) => {
+    (currentUser, profiles, channels, myMembers, unreadIds, settings, lastUnreadChannel) => {
         // If we receive an unread for a channel and then a mention the channel
         // won't be sorted correctly until we receive a message in another channel
         if (!currentUser) {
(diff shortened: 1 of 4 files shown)


src/selectors/entities/channels.js
```diff
@@ -521,7 +522,9 @@ export const getSortedPublicChannelWithUnreadsIds = createIdsSelector(
             const channel = channels[id];
             return !favoriteIds.includes(id) &&
                 teamChannelIds.includes(id) && channel.type === General.OPEN_CHANNEL;
-        }).map((id) => channels[id]).sort(sortChannelsByDisplayName.bind(null, locale));
+        }).map((id) => channels[id]).
+        sort(sortChannelsByDisplayName.bind(null, locale)).
+        sort(sortMutedChannels.bind(null, myMembers));
```
jespino: I'm not sure here about the concatenation of the sorts. Accordingly to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) the sort implementation in Javascript is not necessarily ["stable"](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) . I tested it manually in the last version of Firefox, Chromium and Edge, and everything looks like correct (but I'm not sure about older versions). For me is enough, but up to you to combine them in an single sort function if you think is needed.

src/selectors/entities/channels.js
```diff
@@ -442,16 +443,16 @@ export const getSortedUnreadChannelIds = createIdsSelector(
                 bIsMention = true;
             }
 
-            if (aIsMention === bIsMention) {
+            if (aIsMention === bIsMention && isChannelMuted(bMember) === isChannelMuted(aMember)) {
                 return sortChannelsByDisplayName(locale, a, b);
-            } else if (aIsMention) {
+            } else if (aIsMention || (isChannelMuted(bMember) && !isChannelMuted(aMember))) {
```
jespino: I think this conditions are wrong. For example, if I have two channels, both with mentions, one muted and the other not, the first `a` channel will be always the first one, independently which one is muted or which name it has.

jespino: I'm not able to reproduce the problem, so I will approve this :)

PR #442, opened 2018-03-27
Changed min version detection

jespino:


grundleborg:
eslint doesn't like it :(

src/selectors/entities/general.js
```diff
@@ -26,8 +26,8 @@ export function getCurrentUrl(state: GlobalState): string {
 export function hasNewPermissions(state) {
     const version = state.entities.general.serverVersion;
 
-    // FIXME This must be changed to 4, 8, 0 before we merge this branch into master
-    return isMinimumServerVersion(version, 4, 6, 0);
+    // FIXME This must be changed to 4, 9, 0 before we generate the 4.9.0 release
+    return isMinimumServerVersion(version, 4, 9, 0) || (version.indexOf("dev") !== -1 && isMinimumServerVersion(version, 4, 8, 0))
```
enahum: Not a fan of the dev flag here

grundleborg: ugly, but necessary so that it works prior to branching for 4.9.0. It can go as soon as we branch for 4.9.0.

PR #443, opened 2018-03-28
Changed how loadRolesIfNeeded calls are dispatched

jespino:
#### Summary
Change proposed by elias, use directly dispatch(loadRolesIfNeeded(...)) instead of loadRolesIfNeeded(...)(dispatch, getState)


jespino:
@enahum no rush about it, I know you are out today.

PR #446, opened 2018-03-28
Two flow errors fixed

jespino:


src/selectors/entities/common.js
```diff
@@ -13,7 +14,7 @@ export function getCurrentChannelId(state: GlobalState): string {
     return state.entities.channels.currentChannelId;
 }
 
-export function getMyChannelMemberships(state: GlobalState): string {
+export function getMyChannelMemberships(state: GlobalState): {[string]: ChannelMembership} {
```
saturninoabril: Not sure if I understand this correctly, but why not `{string: ChannelMembership}`?

enahum: It saying that it is an object with string values as properties and channelMembership as values

src/selectors/entities/common.js
```diff
@@ -4,6 +4,7 @@
 
 import type {GlobalState} from '../../types/store';
 import type {UserProfile} from '../../types/users';
+import type {ChannelMembership} from '../../types/channels';
```
enahum: Please change all these to be fullPath instead

enahum: I know is not the scope of the PR but why not do it, right?

src/selectors/entities/common.js
saturninoabril: Thanks @enahum! Good to know, I initially thought that `[string]` is an "array of string values" but when I checked again array of strings is `Array<string>`.

jespino: Yes, the problem is if you use {string: ChannelMembership} you are expressing that the objects have the key "string", and the value of type ChannelMemberships, in the other way you define "keys of type string" and values of type ChannelMemberships.

src/selectors/entities/common.js
jespino: Yes, the reason behind that is when you use it in other repositories, like in mattermost-webapp you find correctly it, if we use absolute path imports, it wont be found.

enahum: Why?

enahum: https://github.com/facebook/flow/issues/4186

jespino: Yes, is a common problem. While I'm developing I can use the global reference, but the types sharing with other library is basically "copy" the file to export the types, so the absolute path in the other project doesn't work. I going to try to do something like "resolve locally mattermost-redux/whatever to src/whatever", allowing all imports use mattermost-redux/types/store. That would work for you?

jespino: I uploaded a commit with the change, I have to test it in webapp flow changes, to verify that everything is working.

src/selectors/entities/common.js
```diff
@@ -2,8 +2,9 @@
 // See License.txt for license information.
 // @flow
 
-import type {GlobalState} from '../../types/store';
-import type {UserProfile} from '../../types/users';
+import type {GlobalState} from 'mattermost-redux/types/store';
```
enahum: So you are saying that `src/types/store` won’t work here?

This is sending mix signals to other developers, that is why I want us to be consistent

jespino: Yeah, the problem is we have the compiled version, which simply compile to mattermost-redux/whatever, and the flow-types, which can be follow the same pattern from the outside, but can't follow the same pattern in the inside (with the current approach). We can rethink it, flow integration is a Work in Progress, so we have to polish it.

jespino:
Already fixed in other PR

PR #454, opened 2018-04-02
Fix postsInThread being reset to null when getPostsSince(...) returns empty list

koxen:
#### Summary
Without this fix comment countners get reset on mobile with new redux
\+ the mobile app needs update to middleware & initial_state, but this is not in scope of this PR

#### Test Information
galaxy s7, android 7


koxen:
@jwilander

PR #456, opened 2018-04-11
[MM-3996] Improved (add/join/leave/remove from channel/team) system messages 

saturninoabril:
#### Summary
- This PR combine all system messages of type "add/join/leave/remove from channel/team".
- These system messages will be marked as deleted at redux store with `"delete_at: 1, "state": "DELETE"`.
- Those consecutive system messages will combine into a single post with following new/modified fields: 
```
"message": "",
"type": "system_combined_user_activity",
"user_activity_posts": [post1, post2],
"system_post_ids": ["post_id_1", "post_id_2"],
```
- `system_post_ids` is used for deleting system post
- `user_activity_posts` is used for rendering on client side

Note: This will have breaking change on mobile RN since system messages will be deleted and combine post don'e have meaningful `message` value (it's empty). I'll submit separate PR for mobile RN.

#### Ticket Link
Jira ticket: [MM-3996](https://mattermost.atlassian.net/browse/MM-3996)

#### Checklist

#### Test Information
This PR was tested on: [Chrome/FF, Mac OS] - to follow on mobile

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -248,10 +248,11 @@ export function resetCreatePostRequest() {
 
 export function deletePost(post) {
     return async (dispatch, getState) => {
+        const state = getState();
         const delPost = {...post};
         if (delPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY) {
             delPost.system_post_ids.forEach((systemPostId) => {
-                const systemPost = Selectors.getPost(getState(), systemPostId);
+                const systemPost = Selectors.getPost(state, systemPostId);
                 if (systemPost) {
                     dispatch(deletePost(systemPost));
                 }
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -247,23 +247,36 @@ export function resetCreatePostRequest() {
 }
 
 export function deletePost(post) {
-    return async (dispatch) => {
+    return async (dispatch, getState) => {
         const delPost = {...post};
+        if (delPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY) {
+            delPost.system_post_ids.forEach((systemPostId) => {
+                const systemPost = Selectors.getPost(getState(), systemPostId);
+                if (systemPost) {
+                    dispatch(deletePost(systemPost));
+                }
+            });
 
-        dispatch({
-            type: PostTypes.POST_DELETED,
-            data: delPost,
-            meta: {
-                offline: {
-                    effect: () => Client4.deletePost(post.id),
-                    commit: {type: PostTypes.POST_DELETED},
-                    rollback: {
-                        type: PostTypes.RECEIVED_POST,
-                        data: delPost,
+            dispatch({
+                type: PostTypes.POST_DELETED,
+                data: delPost,
+            });
+        } else {
+            dispatch({
+                type: PostTypes.POST_DELETED,
+                data: delPost,
+                meta: {
+                    offline: {
+                        effect: () => Client4.deletePost(post.id),
+                        commit: {type: PostTypes.POST_DELETED},
+                        rollback: {
+                            type: PostTypes.RECEIVED_POST,
+                            data: delPost,
+                        },
                     },
                 },
-            },
-        });
+            });
+        }
 
         return {data: true};
     };
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -3,7 +3,7 @@
 
 import {PostTypes, SearchTypes, UserTypes} from 'action_types';
 import {Posts} from 'constants';
-import {comparePosts} from 'utils/post_utils';
+import {comparePosts, combineSystemPost} from 'utils/post_utils';
 
 function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {}, action) {
     const post = action.data;
@@ -39,7 +39,10 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         ];
     }
 
-    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    const combineSystemPosts = combineSystemPost(nextPostsInChannel[channelId], nextPosts);
+    nextPostsInChannel[channelId] = combineSystemPosts.postsForChannel;
+
+    return {posts: combineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
 }
 
 function handleRemovePendingPost(posts = {}, postsInChannel = {}, postsInThread = {}, action) {
@@ -145,13 +148,14 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, postsInThread = {}
 
     // Sort to ensure that the most recent posts are first, with pending
     // and failed posts first
-    postsForChannel.sort((a, b) => {
+    const filteredPostsForChannel = postsForChannel.filter((id) => nextPosts[id]).sort((a, b) => {
         return comparePosts(nextPosts[a], nextPosts[b]);
     });
 
-    nextPostsInChannel[channelId] = postsForChannel;
+    const combineSystemPosts = combineSystemPost(filteredPostsForChannel, nextPosts);
+    nextPostsInChannel[channelId] = combineSystemPosts.postsForChannel;
 
-    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    return {posts: combineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
 }
 
 function handlePendingPosts(pendingPostIds = [], action) {
@@ -275,7 +279,7 @@ function handlePostDeleted(posts = {}, postsInChannel = {}, postsInThread = {}, 
         const channelPosts = postsInChannel[channelId] ? [...postsInChannel[channelId]] : [];
         const postsForChannel = [...channelPosts]; // make sure we don't modify the array we loop over
         for (const id of channelPosts) {
-            if (nextPosts[id].root_id === post.id) {
+            if (nextPosts[id] && nextPosts[id].root_id === post.id) {
                 Reflect.deleteProperty(nextPosts, id);
 
                 const commentIndex = postsForChannel.indexOf(id);
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -3,7 +3,7 @@
 
 import {PostTypes, SearchTypes, UserTypes} from 'action_types';
 import {Posts} from 'constants';
-import {comparePosts, combineSystemPost} from 'utils/post_utils';
+import {comparePosts, combineSystemPosts} from 'utils/post_utils';
 
 function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {}, action) {
     const post = action.data;
@@ -39,10 +39,10 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         ];
     }
 
-    const combineSystemPosts = combineSystemPost(nextPostsInChannel[channelId], nextPosts);
-    nextPostsInChannel[channelId] = combineSystemPosts.postsForChannel;
+    const withCombineSystemPosts = combineSystemPosts(nextPostsInChannel[channelId], nextPosts);
+    nextPostsInChannel[channelId] = withCombineSystemPosts.postsForChannel;
 
-    return {posts: combineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    return {posts: withCombineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
 }
 
 function handleRemovePendingPost(posts = {}, postsInChannel = {}, postsInThread = {}, action) {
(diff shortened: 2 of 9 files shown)


src/reducers/entities/posts.js
```diff
@@ -39,7 +39,10 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         ];
     }
 
-    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    const combineSystemPosts = combineSystemPost(nextPostsInChannel[channelId], nextPosts);
```
hmhealey: When I looked at this yesterday, I wasn't really sure about combining system posts in the store. I was thinking it would be better to have the redux store match what is on the server, but then we'd have to remember to combine system posts and such all over the place before showing them in the UI, so I think this actually ends up being a really elegant solution to this

src/reducers/entities/posts.js
```diff
@@ -145,13 +148,14 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, postsInThread = {}
 
     // Sort to ensure that the most recent posts are first, with pending
     // and failed posts first
-    postsForChannel.sort((a, b) => {
+    const filteredPostsForChannel = postsForChannel.filter((id) => nextPosts[id]).sort((a, b) => {
```
hmhealey: Why would we need to filter out post ids for posts that don't exist? Is there something that normally causes that to happen, or are you just being extra safe here?

src/reducers/entities/posts.js
```diff
@@ -3,7 +3,7 @@
 
 import {PostTypes, SearchTypes, UserTypes} from 'action_types';
 import {Posts} from 'constants';
-import {comparePosts} from 'utils/post_utils';
+import {comparePosts, combineSystemPost} from 'utils/post_utils';
```
hmhealey: This method should be `combineSystemPosts` so that it's clearer that it does all posts at once

src/utils/helpers.js
```diff
@@ -79,3 +79,24 @@ export const sendMessageToRemoveBackwardsCompatibility = (currentVersion, minVer
         console.warn('You can now remove backwards-compatibility code from the project, as referenced in the stacktrace below.'); //eslint-disable-line no-console
     }
 };
+
+// Generates a RFC-4122 version 4 compliant globally unique identifier.
+export function generateId() {
```
hmhealey: Can you make the `TestHelper.generateId` method in `test/test_helper.js` call this so that we're not duplicating the logic?

src/utils/post_utils.js
```diff
@@ -175,3 +183,98 @@ export function comparePosts(a, b) {
...
+            }
+        }
+
+        if (channelPost.type === '') {
+            if (userActivitySystemPosts.length > 0) {
+                const combinedPost = {
+                    id: combinedPostId || generateId(),
+                    root_id: '',
+                    type: Posts.POST_TYPES.COMBINED_USER_ACTIVITY,
+                    message: '',
```
hmhealey: You mentioned that having no message will make this not work with the mobile app temporarily, but what if you make the combined post's message just the messages of each system posts joined together? That would help solve that case and also make it easier if anyone else is building a client with this

src/utils/post_utils.js
```diff
@@ -175,3 +183,98 @@ export function comparePosts(a, b) {
...
+            postsForChannel.push(channelPost.id);
+        } else if (
+                userActivitySystemPosts.length === MAX_COMBINED_SYSTEM_POSTS ||
+                (userActivitySystemPosts.length > 0 && i === postsIds.length - 1)
+            ) {
+            const combinedPost = {
+                id: combinedPostId || generateId(),
+                root_id: '',
+                type: Posts.POST_TYPES.COMBINED_USER_ACTIVITY,
+                message: '',
```
hmhealey: Same here

src/utils/post_utils.js
```diff
@@ -175,3 +183,98 @@ export function comparePosts(a, b) {
...
+            } else if (channelPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY) {
+                userActivitySystemPosts = [...userActivitySystemPosts, ...channelPost.user_activity_posts];
+                systemPostIds = [...systemPostIds, ...channelPost.system_post_ids];
+
+                combinedPostId = channelPost.id;
+            }
+        }
+
+        if (channelPost.type === '') {
+            if (userActivitySystemPosts.length > 0) {
```
hmhealey: It would be nice if we could combine the two branches of this if statement since their logic is almost identical. Would this work if it was something like
```javascript
if (
    (channelPost.type === '' && userActivitySystemPosts.length > 0) ||
    userActivitySystemPosts.length === MAX_COMBINED_SYSTEM_POSTS ||
    (userActivitySystemPosts.length > 0 && i === postsIds.length - 1)
) {
    ...
}
```

src/reducers/entities/posts.js
saturninoabril: I'm seeing `undefined` after deleting combined post; somehow the ID of combined post remained in the postsForChannel.  Not sure what I'm missing there but that's something I need to investigate further.

src/utils/post_utils.js
saturninoabril: I think one of my worry is the localization but I'll see what I can do here.

hmhealey: I think it's fine if it's a simpler message like "Harrison joined the channel. Harrison left the channel." here, but it'll make sure that we have something to fall back to

src/reducers/entities/posts.js
saturninoabril: Fixed this by adding `channel_id` to combined post.

src/utils/helpers.js
saturninoabril: Done

src/utils/post_utils.js
saturninoabril: hmmm, forgot about this... will update after PM review.

src/utils/post_utils.js
saturninoabril: Done

src/reducers/entities/posts.js
```diff
@@ -39,7 +39,10 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         ];
     }
 
-    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    const withCombineSystemPosts = combineSystemPosts(nextPostsInChannel[channelId], nextPosts);
+    nextPostsInChannel[channelId] = withCombineSystemPosts.postsForChannel;
+
+    return {posts: withCombineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
```
enahum: `nextPostsInChannel` and `nextPostsInThread` here should have the ids for those posts that were removed from the combine stuff as you will end up having something like this for example: for 3 system messages in a channel you'll have 1 post in the posts entity and 3 posts in the postsInChannel entity.

sounds about right? probably the cause that you needed to filter later on

src/reducers/entities/posts.js
```diff
@@ -149,9 +152,10 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, postsInThread = {}
         return comparePosts(nextPosts[a], nextPosts[b]);
     });
 
-    nextPostsInChannel[channelId] = postsForChannel;
+    const withCombineSystemPosts = combineSystemPosts(postsForChannel, nextPosts, channelId);
+    nextPostsInChannel[channelId] = withCombineSystemPosts.postsForChannel;
 
-    return {posts: nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
+    return {posts: withCombineSystemPosts.nextPosts, postsInChannel: nextPostsInChannel, postsInThread: nextPostsInThread};
```
enahum: same as before

src/selectors/entities/posts.js
```diff
@@ -230,7 +230,9 @@ export function makeGetPostsInChannel() {
                 }
 
                 const previousPost = allPosts[postIds[i + 1]] || {create_at: 0};
-                posts.push(formatPostInChannel(post, previousPost, i, allPosts, postsInThread, postIds, currentUser));
+                if (post) {
```
enahum: I don't mind being defensive here at all, but out of curiosity why did you need it?

src/utils/helpers.js
enahum: are you going to be removing this helper?

src/utils/post_utils.js
```diff
@@ -3,10 +3,14 @@
...
 import {General, Posts, Preferences, Permissions} from 'constants';
 
-import {getPreferenceKey} from './preference_utils';
 import {hasNewPermissions} from 'selectors/entities/general';
 import {haveIChannelPermission} from 'selectors/entities/roles';
 
+import {generateId} from './helpers';
+import {getPreferenceKey} from './preference_utils';
+
+const MAX_COMBINED_SYSTEM_POSTS = 100;
```
enahum: curiosity: why 100? who came up with this number and why?

src/utils/post_utils.js
```diff
@@ -175,3 +183,82 @@ export function comparePosts(a, b) {
...
+        if (
+            channelPost.delete_at === 0 &&
+            (isUserActivityPost(channelPost.type) || channelPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY)
+        ) {
+            if (!createAt || createAt > channelPost.create_at) {
+                createAt = channelPost.create_at;
+            }
+
+            if (isUserActivityPost(channelPost.type)) {
+                userActivitySystemPosts = [...userActivitySystemPosts, channelPost];
```
enahum: why are you creating a new array every time instead of just adding the post to the existing array?

src/utils/post_utils.js
```diff
@@ -175,3 +183,82 @@ export function comparePosts(a, b) {
...
+                systemPostIds.push(channelPost.id);
+
+                if (nextPosts[channelPost.id]) {
+                    nextPosts[channelPost.id] = {...channelPost, state: Posts.POST_DELETED, delete_at: 1};
+                }
+            } else if (channelPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY) {
+                userActivitySystemPosts = [...userActivitySystemPosts, ...channelPost.user_activity_posts];
+                systemPostIds = [...systemPostIds, ...channelPost.system_post_ids];
```
enahum: same here and the line above

src/utils/post_utils.js
```diff
@@ -175,3 +183,82 @@ export function comparePosts(a, b) {
...
+            createAt = null;
+            combinedPostId = null;
+
+            if (channelPost.type === '') {
+                postsForChannel.push(channelPost.id);
+            }
+        } else if (channelPost.type === '') {
+            postsForChannel.push(channelPost.id);
+        }
+    });
```
enahum: after adding the new combined post to the `postsForChannel` you should remove those that were marked as `DELETED`? or do you see a better way to accomplish that?

src/reducers/entities/posts.js
saturninoabril: Got it, I'll update this.

src/selectors/entities/posts.js
saturninoabril: Ah, i missed to revert this. This change is not needed anymore. (The problem I'm seeing initially is already fixed. It was fixed by adding `channel_id` in the combinedPost)

src/utils/post_utils.js
saturninoabril: It's only out of convenience during my first iteration :) and just missed to clean up. Will update using push instead.

src/utils/post_utils.js
saturninoabril: I'm just marking those system posts as deleted so that for the next round of received post/s those are known as received already and no need to merge with existing combinedPost. I think it's straightforward rather than checking each time whether the combinedPost has already the system post or not.

src/utils/helpers.js
saturninoabril: Not on this PR, but after this has been merged. I'll clean up that code - will delete `TestHelper.generateId()` and will just directly call `UtilsHelper.generateId()` for our test files.

(I initially thought to consider that in this PR but there are too many `TestHelper.generateId()` everywhere and I find it unnecessary for this PR.)

src/reducers/entities/posts.js
saturninoabril: Can confirm that those system posts are included in `nextPostsInChannel` but not necessary in `nextPostsInThread`. Those system posts on `nextPostsInChannel` were marked as deleted.

saturninoabril:
@hmhealey @enahum I've updated the PR so that this will not break the current Mobile RN (like the screenshots below just in case this PR gets merged without change on it yet). I'm currently working on Mobile RN and will submit later.

screen shot 2018-05-14 at 9 45 54 pm
screen shot 2018-05-14 at 9 44 58 pm
screen shot 2018-05-14 at 9 45 12 pm

src/actions/posts.js
```diff
@@ -247,23 +247,36 @@ export function resetCreatePostRequest() {
 }
 
 export function deletePost(post) {
-    return async (dispatch) => {
+    return async (dispatch, getState) => {
         const delPost = {...post};
+        if (delPost.type === Posts.POST_TYPES.COMBINED_USER_ACTIVITY) {
+            delPost.system_post_ids.forEach((systemPostId) => {
+                const systemPost = Selectors.getPost(getState(), systemPostId);
```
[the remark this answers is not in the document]
saturninoabril: Thanks @enahum! You're right. I've made an update into it.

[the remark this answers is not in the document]
saturninoabril: Thanks @enahum! You're right. I've made an update into it.

PR #457, opened 2018-04-11
MM-10136: Not allowing to edit others posts in the interface

jespino:
#### Summary
Not allowing to edit others posts in the interface

#### Ticket Link
[MM-10136](https://mattermost.atlassian.net/browse/MM-10136)

#### Checklist

### DIFF
--- a/src/utils/post_utils.js
+++ b/src/utils/post_utils.js
@@ -72,7 +72,9 @@ export function canEditPost(state, config, license, teamId, channelId, userId, p
         if (hasNewPermissions(state)) {
             canEdit = canEdit && haveIChannelPermission(state, {team: teamId, channel: channelId, permission: Permissions.EDIT_POST});
             if (!isOwner) {
-                canEdit = canEdit && haveIChannelPermission(state, {team: teamId, channel: channelId, permission: Permissions.EDIT_OTHERS_POSTS});
+                // I only can edit my posts, at least until phase-2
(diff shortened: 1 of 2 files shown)


src/utils/post_utils.js
```diff
@@ -72,7 +72,9 @@ export function canEditPost(state, config, license, teamId, channelId, userId, p
         if (hasNewPermissions(state)) {
             canEdit = canEdit && haveIChannelPermission(state, {team: teamId, channel: channelId, permission: Permissions.EDIT_POST});
             if (!isOwner) {
-                canEdit = canEdit && haveIChannelPermission(state, {team: teamId, channel: channelId, permission: Permissions.EDIT_OTHERS_POSTS});
+                // I only can edit my posts, at least until phase-2
```
lieut-data: I could sworn I had posted this comment already, but it vanished...

Is the only thing stopping this from working a client-side check? Do we need server changes to complement this change?

jespino: Same here, Server change not needed because it is intentionally disabled only in the interface.

koxen:
are you aware it needs a fix on server side too ?

lieut-data:
@koxen, @jespino and I discussed this above (inline). It turns out that the ability for an admin to edit posts was always in place, but the UI was incorrectly exposed (as it's not likely to be a common use-case). Going forward, my understanding is that advanced permissions is precisely about constraining such abilities as individual use cases demand.

PR #461, opened 2018-04-17
MM-8786: Adds Schemes.

mkraft:
#### Summary
Adds Schemes support.

#### Ticket Link
[MM-8786](https://mattermost.atlassian.net/browse/MM-8786)

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/schemes.js
+++ b/src/actions/schemes.js
@@ -1,0 +1,85 @@
+// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+// @flow
+
+import {Client4} from 'client';
+import {SchemeTypes} from 'action_types';
+import {General} from 'constants';
+import {bindClientFunc} from './helpers';
+import type {Scheme, SchemeScope, SchemePatch} from '../types/schemes';
+
+export function getScheme(schemeId: string) {
+    return bindClientFunc(
+        Client4.getScheme,
+        SchemeTypes.GET_SCHEME_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEME, SchemeTypes.GET_SCHEME_SUCCESS],
+        SchemeTypes.GET_SCHEME_FAILURE,
+        schemeId
+    );
+}
+
+export function getSchemes(scope: SchemeScope, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
+    return bindClientFunc(
+        Client4.getSchemes,
+        SchemeTypes.GET_SCHEMES_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEMES, SchemeTypes.GET_SCHEMES_SUCCESS],
+        SchemeTypes.GET_SCHEMES_FAILURE,
+        scope,
+        page,
+        perPage
+    );
+}
+
+export function createScheme(scheme: Scheme) {
+    return bindClientFunc(
+        Client4.createScheme,
+        SchemeTypes.CREATE_SCHEME_REQUEST,
+        [SchemeTypes.CREATED_SCHEME, SchemeTypes.CREATE_SCHEME_SUCCESS],
+        SchemeTypes.CREATE_SCHEME_FAILURE,
+        scheme
+    );
+}
+
+export function deleteScheme(schemeId: string) {
+    return bindClientFunc(
+        Client4.deleteScheme,
+        SchemeTypes.DELETE_SCHEME_REQUEST,
+        [SchemeTypes.DELETED_SCHEME, SchemeTypes.DELETE_SCHEME_SUCCESS],
+        SchemeTypes.DELETE_SCHEME_FAILURE,
+        schemeId
+    );
+}
+
+export function patchScheme(scheme: SchemePatch) {
+    return bindClientFunc(
+        Client4.patchScheme,
+        SchemeTypes.PATCH_SCHEME_REQUEST,
+        [SchemeTypes.PATCHED_SCHEME, SchemeTypes.PATCH_SCHEME_SUCCESS],
+        SchemeTypes.PATCH_SCHEME_FAILURE,
+        scheme
+    );
+}
+
+export function getSchemeTeams(schemeId: string, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
+    return bindClientFunc(
+        Client4.getSchemeTeams,
+        SchemeTypes.GET_SCHEME_TEAMS_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEME_TEAMS, SchemeTypes.GET_SCHEME_TEAMS_SUCCESS],
+        SchemeTypes.GET_SCHEME_TEAMS_FAILURE,
+        schemeId,
+        page,
+        perPage
+    );
+}
+
+export function getSchemeChannels(schemeId: string, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
+    return bindClientFunc(
+        Client4.getSchemeChannels,
+        SchemeTypes.GET_SCHEME_CHANNELS_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEME_CHANNELS, SchemeTypes.GET_SCHEME_CHANNELS_SUCCESS],
+        SchemeTypes.GET_SCHEME_CHANNELS_FAILURE,
+        schemeId,
+        page,
+        perPage
+    );
+}
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -233,6 +233,10 @@ export default class Client4 {
         return `${this.getBaseRoute()}/system/timezones`;
     }
 
+    getSchemesRoute() {
+        return `${this.getBaseRoute()}/schemes`;
+    }
+
     getOptions(options) {
         const newOptions = Object.assign({}, options);
 
@@ -284,19 +288,19 @@ export default class Client4 {
         );
     }
 
-    patchMe = async (user) => {
+    patchMe = async (userPatch) => {
         return this.doFetch(
             `${this.getUserRoute('me')}/patch`,
-            {method: 'put', body: JSON.stringify(user)}
+            {method: 'put', body: JSON.stringify(userPatch)}
         );
     }
 
-    patchUser = async (user) => {
+    patchUser = async (userPatch) => {
         this.trackEvent('api', 'api_users_patch');
 
         return this.doFetch(
-            `${this.getUserRoute(user.id)}/patch`,
-            {method: 'put', body: JSON.stringify(user)}
+            `${this.getUserRoute(userPatch.id)}/patch`,
+            {method: 'put', body: JSON.stringify(userPatch)}
         );
     }
 
@@ -1104,12 +1108,12 @@ export default class Client4 {
         );
     };
 
-    patchChannel = async (channelId, patch) => {
+    patchChannel = async (channelId, channelPatch) => {
         this.trackEvent('api', 'api_channels_patch', {channel_id: channelId});
 
         return this.doFetch(
             `${this.getChannelRoute(channelId)}/patch`,
-            {method: 'put', body: JSON.stringify(patch)}
+            {method: 'put', body: JSON.stringify(channelPatch)}
         );
     };
 
@@ -1252,6 +1256,7 @@ export default class Client4 {
     };
 
     // Post Routes
+
     createPost = async (post) => {
         this.trackEvent('api', 'api_posts_create', {channel_id: post.channel_id});
 
@@ -1281,12 +1286,12 @@ export default class Client4 {
         );
     };
 
-    patchPost = async (post) => {
-        this.trackEvent('api', 'api_posts_patch', {channel_id: post.channel_id});
+    patchPost = async (postPatch) => {
+        this.trackEvent('api', 'api_posts_patch', {channel_id: postPatch.channel_id});
 
         return this.doFetch(
-            `${this.getPostRoute(post.id)}/patch`,
-            {method: 'put', body: JSON.stringify(post)}
+            `${this.getPostRoute(postPatch.id)}/patch`,
+            {method: 'put', body: JSON.stringify(postPatch)}
         );
     };
 
@@ -1864,6 +1869,7 @@ export default class Client4 {
     };
 
     // Timezone Routes
+
     getTimezones = async () => {
         return this.doFetch(
             `${this.getTimezonesRoute()}`,
@@ -2171,6 +2177,7 @@ export default class Client4 {
     };
 
     // Role Routes
+
     getRole = async (roleId) => {
         return this.doFetch(
             `${this.getRolesRoute()}/${roleId}`,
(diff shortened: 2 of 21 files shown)


src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,60 @@
...
+export function getScheme(state: GlobalState, id: string): Scheme {
+    const schemes = getSchemes(state);
+    return schemes[id];
+}
+
+export function getSchemeChannels(state: GlobalState, schemeId: string): Array<Channel> {
+    const {scope} = getScheme(state, schemeId);
+
+    if (scope === ScopeTypes.TEAM) {
+        throw new Error(`Not implemented: scheme '${schemeId}' is team-scope but 'getSchemeChannels' only accepts channel-scoped schemes.`);
```
mkraft: Does anyone have any opinions about throwing an error here? I'm 0/5 about whether this rather returns an empty array in this case.

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    return schemeChannels;
+}
+
+export function getSchemeTeams(state: GlobalState, schemeId: string): Array<Team> {
+    const {scope} = getScheme(state, schemeId);
+
+    if (scope === ScopeTypes.CHANNEL) {
+        throw new Error(`Error: scheme '${schemeId}' is channel-scoped but 'getSchemeChannels' only accepts team-scoped schemes.`);
```
mkraft: Same goes for here re: error throw.

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,58 @@
...
+    const schemeChannels: Array<Channel> = [];
+
+    Object.entries(getAllChannels(state)).forEach((item: [string, Channel]) => {
+        const [, channel: Channel] = item;
+        if (channel.scheme_id === schemeId) {
+            schemeChannels.push(channel);
+        }
+    });
+
+    return schemeChannels;
```
mkraft: Also any thoughts about returning an array vs object here?

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,58 @@
...
+    const schemeTeams: Array<Team> = [];
+
+    Object.entries(getTeams(state)).forEach((item: [string, Team]) => {
+        const [, team: Team] = item;
+        if (team.scheme_id === schemeId) {
+            schemeTeams.push(team);
+        }
+    });
+
+    return schemeTeams;
```
mkraft: Same goes for here re: array vs object.

src/selectors/entities/schemes.js
jwilander: Yeah I think it's totally valid to just return an empty array instead of throwing the error and I'd suggest doing that instead 👍

src/selectors/entities/schemes.js
jwilander: 0/5, I usually lean towards an array if it's likely I'm to need to order the results and an object otherwise but I have no strong opinion

src/selectors/entities/schemes.js
mkraft: @jwilander Maybe I'll return an empty array and print a JS warning so that a developer using it realizes that it's not a valid param to pass a `schemeId` associated to the wrong scope.

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,62 @@
...
+export function getSchemes(state: GlobalState): { [string]: Scheme } {
+    return state.entities.schemes.schemes;
+}
+
+export function getScheme(state: GlobalState, id: string): Scheme {
+    const schemes = getSchemes(state);
+    return schemes[id];
+}
+
+export function getSchemeChannels(state: GlobalState, schemeId: string): Array<Channel> {
```
jwilander: Is this going to get called often or used in a `mapStatesToProps` function? If so, I'd suggest using `createSelector` [similar to this](https://github.com/mattermost/mattermost-redux/blob/master/src/selectors/entities/channels.js#L75) so that it can be memoized and not recalculated every time

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,62 @@
...
+        const [, channel: Channel] = item;
+        if (channel.scheme_id === schemeId) {
+            schemeChannels.push(channel);
+        }
+    });
+
+    return schemeChannels;
+}
+
+export function getSchemeTeams(state: GlobalState, schemeId: string): Array<Team> {
```
jwilander: Same comment as above

mkraft:
@jwilander Good idea. Switched those two selectors to the cached versions.

src/actions/schemes.js
```diff
@@ -0,0 +1,83 @@
...
+    return bindClientFunc(
+        Client4.getScheme,
+        SchemeTypes.GET_SCHEME_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEME, SchemeTypes.GET_SCHEME_SUCCESS],
+        SchemeTypes.GET_SCHEME_FAILURE,
+        schemeId
+    );
+}
+
+export function getSchemes(page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
```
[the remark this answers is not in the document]
mkraft: Added.

src/client/client4.js
```diff
@@ -1090,12 +1094,12 @@ export default class Client4 {
         );
     };
 
-    patchChannel = async (channelId, patch) => {
+    patchChannel = async (channelId, channel) => {
```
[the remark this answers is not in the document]
mkraft: For consistency. But I have actually renamed them all since.

src/client/client4.js
```diff
@@ -2178,6 +2185,63 @@ export default class Client4 {
         );
     };
 
+    // Scheme Routes
+
+    getSchemes = async (scope = '', page = 0, perPage = PER_PAGE_DEFAULT) => {
+        return this.doFetch(
+            `${this.getSchemesRoute()}${buildQueryString({page, per_page: perPage})}`,
```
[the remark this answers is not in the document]
mkraft: Added.

src/client/client4.js
```diff
@@ -2178,6 +2185,63 @@ export default class Client4 {
...
+    deleteScheme = async (schemeId) => {
+        this.trackEvent('api', 'api_schemes_delete');
+
+        return this.doFetch(
+            `${this.getSchemesRoute()}/${schemeId}`,
+            {method: 'delete'}
+        );
+    };
+
+    patchScheme = async (schemeId, scheme) => {
```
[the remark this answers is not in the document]
mkraft: Agreed. I was doing this for consistency with all of the other `patch*` methods in `Client4` but I have changed them all.

src/types/schemes.js
```diff
@@ -0,0 +1,21 @@
...
+
+export type Scheme = {
+    id: string,
+    name: string,
+    description: string,
+    create_at: number,
+    update_at: number,
+    delete_at: number,
+    scope: SchemeScope,
+    defaultteamadminrole: string,
```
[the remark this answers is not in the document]
mkraft: It seems to be the convention.

src/actions/schemes.js
```diff
@@ -0,0 +1,84 @@
...
+        Client4.getSchemeTeams,
+        SchemeTypes.GET_SCHEME_TEAMS_REQUEST,
+        [SchemeTypes.RECEIVED_SCHEME_TEAMS, SchemeTypes.GET_SCHEME_TEAMS_SUCCESS],
+        SchemeTypes.GET_SCHEME_TEAMS_FAILURE,
+        schemeId,
+        page,
+        perPage
+    );
+}
+export function getSchemeChannels(schemeId: string, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
```
jespino: This actions haven't a reducer associated with the result, so I can request to the server, but I not handle the result at all.

src/actions/schemes.js
```diff
@@ -0,0 +1,84 @@
...
+    return bindClientFunc(
+        Client4.patchScheme,
+        SchemeTypes.PATCH_SCHEME_REQUEST,
+        [SchemeTypes.PATCHED_SCHEME, SchemeTypes.PATCH_SCHEME_SUCCESS],
+        SchemeTypes.PATCH_SCHEME_FAILURE,
+        scheme
+    );
+}
+
+export function getSchemeTeams(schemeId: string, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
```
jespino: This actions haven't a reducer associated with the result, so I can request to the server, but I not handle the result at all.

src/selectors/entities/schemes.js
```diff
@@ -0,0 +1,75 @@
...
+                    schemeChannels.push(channel);
+                }
+            });
+
+            return schemeChannels;
+        }
+    );
+}
+
+export function makeGetSchemeTeams() {
```
jespino: I would expect to user here the getSchemeTeams action results. We can implement it this way, but then, the action is not needed. Actually, I think that is better to use the getSchemeTeams because is possible that we have a lot of teams that aren't loaded in our current context.

jespino: The same comment can be applied to the makeGetSchemeChannels.

src/store/initial_state.js
```diff
@@ -501,6 +501,36 @@ const state: GlobalState = {
                 error: null,
             },
         },
+        schemes: {
```
jespino: I miss the initial state for entities.schemes

src/types/schemes.js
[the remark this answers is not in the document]
jespino: It is what comes from the server, without any conversion of casing.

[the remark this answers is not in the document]
grundleborg: OK, in that case it should have underscores between the words.

src/actions/schemes.js
mkraft: Added to channels reducer.

mkraft:
One thing I don't love is the need to make call `this.props.actions.getSchemes('team')` before calling `this.props.actions.getSchemeTeams('qctk9aoy47dpf8pxe33ardkoty')` or else you're likely to get state data, but on the other hand it gives the developer the power to choose whether the call is expensive or not. Since I'm new to the patterns in the repo let me know if there's a better way.

mkraft:
If we knew the number of total pages of results we could do something like this:
```js
export function getSchemesAndSchemeTeams(schemeId: string, page: number = 0, perPage: number = General.PAGE_SIZE_DEFAULT) {
    return (dispatch: ReduxDispatch) => Promise.all([
        dispatch(getSchemes('team', 0, General.PAGE_SIZE_DEFAULT)),
        dispatch(getSchemes('team', 1, General.PAGE_SIZE_DEFAULT)),
        dispatch(getSchemeTeams(schemeId, page, perPage)),
    ]);
}
```

mkraft:
```diff
index f6f3c4e8..f977cdbf 100644
--- a/components/channel_header/channel_header.jsx
+++ b/components/channel_header/channel_header.jsx
@@ -63,6 +63,8 @@ export default class ChannelHeader extends React.Component {
             openModal: PropTypes.func.isRequired,
             getCustomEmojisInText: PropTypes.func.isRequired,
             updateChannelNotifyProps: PropTypes.func.isRequired,
+            getSchemeAndSchemeTeams: PropTypes.func,
+            getSchemeAndSchemeChannels: PropTypes.func,
         }).isRequired,
         channel: PropTypes.object.isRequired,
         channelMember: PropTypes.object.isRequired,
@@ -78,6 +80,8 @@ export default class ChannelHeader extends React.Component {
             Object.values(RHSStates)
         ),
         enableWebrtc: PropTypes.bool.isRequired,
+        schemeTeams: PropTypes.array,
+        schemeChannels: PropTypes.array,
     };
 
     static defaultProps = {
@@ -101,6 +105,8 @@ export default class ChannelHeader extends React.Component {
     }
 
     componentDidMount() {
+        this.props.actions.getSchemeAndSchemeTeams('qctk9aoy47dpf8pxe33ardkoty');
+        this.props.actions.getSchemeAndSchemeChannels('wa7i5u37gpg5xdmdgf1th4531w');
         this.props.actions.getCustomEmojisInText(this.props.channel.header);
         WebrtcStore.addChangedListener(this.onWebrtcChange);
         WebrtcStore.addBusyListener(this.onBusy);
@@ -296,6 +302,8 @@ export default class ChannelHeader extends React.Component {
     };
 
     render() {
+        console.log('schemeTeams: ', this.props.schemeTeams);
+        console.log('schemeChannels: ', this.props.schemeChannels);
         if (Utils.isEmptyObject(this.props.channel) ||
                 Utils.isEmptyObject(this.props.channelMember) ||
                 Utils.isEmptyObject(this.props.currentUser)) {
diff --git a/components/channel_header/index.js b/components/channel_header/index.js
index 1bcb3ca0..de079b2b 100644
--- a/components/channel_header/index.js
+++ b/components/channel_header/index.js
@@ -4,6 +4,8 @@
 import {connect} from 'react-redux';
 import {bindActionCreators} from 'redux';
 import {favoriteChannel, leaveChannel, unfavoriteChannel, updateChannelNotifyProps} from 'mattermost-redux/actions/channels';
+import {getSchemeAndSchemeTeams, getSchemeAndSchemeChannels} from 'mattermost-redux/actions/schemes';
+import {makeGetSchemeTeams, makeGetSchemeChannels} from 'mattermost-redux/selectors/entities/schemes';
 import {getCustomEmojisInText} from 'mattermost-redux/actions/emojis';
 import {General, Preferences} from 'mattermost-redux/constants';
 import {getChannel, getMyChannelMember, isCurrentChannelReadOnly} from 'mattermost-redux/selectors/entities/channels';
@@ -31,7 +33,8 @@ function mapStateToProps(state, ownProps) {
     const channel = getChannel(state, ownProps.channelId) || {};
     const prefs = state.entities.preferences.myPreferences;
     const user = getCurrentUser(state);

+    const gst = makeGetSchemeTeams();
+    const gsc = makeGetSchemeChannels();
     let dmUser;
     let dmUserStatus;
     if (channel && channel.type === General.DM_CHANNEL) {
@@ -57,6 +60,8 @@ function mapStateToProps(state, ownProps) {
         isLicensed: license.IsLicensed === 'true',
         enableWebrtc: config.EnableWebrtc === 'true',
         isReadOnly: isCurrentChannelReadOnly(state),
+        schemeTeams: gst(state, {schemeId: 'qctk9aoy47dpf8pxe33ardkoty'}),
+        schemeChannels: gsc(state, {schemeId: 'wa7i5u37gpg5xdmdgf1th4531w'}),
     };
 }
 
@@ -74,6 +79,8 @@ function mapDispatchToProps(dispatch) {
             openModal,
             getCustomEmojisInText,
             updateChannelNotifyProps,
+            getSchemeAndSchemeTeams,
+            getSchemeAndSchemeChannels,
         }, dispatch),
     };
 }
```

src/reducers/entities/channels.js
```diff
@@ -102,6 +102,14 @@ function channels(state = {}, action) {
         };
     }
 
+    case SchemeTypes.RECEIVED_SCHEME_CHANNELS: {
+        const nextState = {...state};
+        for (const channel of action.data) {
+            nextState[channel.id] = channel;
+        }
+        return nextState;
+    }
```
jwilander: Since this is the same as `RECEIVED_CHANNELS` you could do this so you don't have to duplicate the code:

```javascript
        case SchemeTypes.RECEIVED_SCHEME_CHANNELS:
        case ChannelTypes.RECEIVED_CHANNELS: {
        const nextState = {...state};
        for (const channel of action.data) {
            nextState[channel.id] = channel;
        }
        return nextState;
    }
```

src/reducers/entities/teams.js
```diff
@@ -42,6 +42,15 @@ function teams(state = {}, action) {
...
         return state;
     }
+
+    case SchemeTypes.RECEIVED_SCHEME_TEAMS: {
+        const nextState = {...state};
+        for (const team of action.data) {
+            nextState[team.id] = team;
+        }
+        return nextState;
+    }
```
jwilander: Same as for channels

src/reducers/entities/teams.js
```diff
@@ -22,6 +22,7 @@ function teams(state = {}, action) {
     case TeamTypes.RECEIVED_TEAMS_LIST:
         return Object.assign({}, state, teamListToMap(action.data));
     case TeamTypes.RECEIVED_TEAMS:
+    case SchemeTypes.RECEIVED_SCHEME_TEAMS:
         return Object.assign({}, state, action.data);
```
jwilander: Isn't `action.data` an array for `RECEIVED_SCHEME_TEAMS`? Probably needs to be moved up with `RECEIVED_TEAMS_LIST`

mkraft: blargh

PR #463, opened 2018-04-18
Replacing hash usage with token for invitations

jespino:
#### Summary
Replacing hash usage with token for invitations. It depends on the PR mattermost/mattermost-server#8604

#### Ticket Link
[MM-9779](https://mattermost.atlassian.net/browse/MM-9779)

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -58,13 +58,13 @@ export function generateMfaSecret(userId) {
     );
 }
 
-export function createUser(user, data, token, inviteId) {
+export function createUser(user, token, inviteId) {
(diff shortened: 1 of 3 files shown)


saturninoabril:
I think this should be updated as well - https://github.com/mattermost/mattermost-redux/blob/master/src/client/client.js#L216

jespino:
Yes, you are right. Actually I'm enteirly removing the "data" param for the server, because is not needed.

enahum:
@jespino do we need to remove it from where @saturninoabril is saying? cause I think that is for the APIv3 stuff

PR #466, opened 2018-04-20
MM-8528: manage sendingPostIds and avoid double-sending

lieut-data:
#### Summary
A long time ago, the webapp managed its own deduplication of retried posts itself by relying on the resolved/rejected promise returned by the client. That was later abstracted away under a dispatch, resulting in the webapp being unable to distinguish success/failure, and effectively allowing the user to mash the "Retry" button and duplicate posts.

This moves that logic into Redux by maintaining a `sendingPostIds` array (analogous to `pendingPostIds`, and ignoring any attempts to `createPost` if it's already sending. Note that this doesn't impact the `createPostImmediately` path -- you can't even retry on iOS, and I'm hesitant to touch that code at present.

I've added unit tests for the selectors and reducers... but cannot for the life of me add a unit test for the action itself without breaking all the other unit tests when I attempt to `nock` calls. I see there's a corresponding test that's commented out probably for similar difficulties. Once we switch to `jest`, I anticipate an easier pattern might be to mock `Client4` instead.

This PR will be accompanied by a webapp PR to remove the old deduplication code and update to this version of Redux.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-8528

#### Checklist

#### Test Information
This PR was tested on: Chrome

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -191,6 +191,42 @@ function handlePendingPosts(pendingPostIds = [], action) {
     }
 }
 
+function handleSendingPosts(sendingPostIds = [], action) {
+    switch (action.type) {
+    case PostTypes.RECEIVED_NEW_POST: {
+        const sendingPostId = action.data.id;
+        if (sendingPostIds.includes(sendingPostId)) {
+            return sendingPostIds;
+        }
+
+        return [
+            ...sendingPostIds,
+            sendingPostId,
+        ];
+    }
+    case PostTypes.RECEIVED_POST: {
+        const sendingPostId = action.data.id;
+        if (!sendingPostIds.includes(sendingPostId)) {
+            return sendingPostIds;
+        }
+
+        return sendingPostIds.filter((postId) => postId !== sendingPostId);
+    }
+    case PostTypes.RECEIVED_POSTS: {
+        const postIds = Object.values(action.data.posts).map((post) => post.pending_post_id);
+
+        const nextSendingPostIds = sendingPostIds.filter((sendingPostId) => !postIds.includes(sendingPostId));
+        if (nextSendingPostIds.length === sendingPostIds.length) {
+            return sendingPostIds;
+        }
+
+        return nextSendingPostIds;
+    }
+    default:
+        return sendingPostIds;
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/reducers/entities/posts.js
```diff
@@ -191,6 +191,42 @@ function handlePendingPosts(pendingPostIds = [], action) {
...
+    }
+    case PostTypes.RECEIVED_POSTS: {
+        const postIds = Object.values(action.data.posts).map((post) => post.pending_post_id);
+
+        const nextSendingPostIds = sendingPostIds.filter((sendingPostId) => !postIds.includes(sendingPostId));
+        if (nextSendingPostIds.length === sendingPostIds.length) {
+            return sendingPostIds;
+        }
+
+        return nextSendingPostIds;
```
saturninoabril: Maybe directly return:
```
return sendingPostIds.filter((sendingPostId) => !postIds.includes(sendingPostId));
```

test/reducers/entities/posts.test.js
```diff
@@ -37,10 +37,130 @@ describe('Reducers.posts', () => {
...
+                            id: 'different_actual_post_id',
+                            pending_post_id: 'different_post_id',
+                        },
+                    },
+                },
+            };
+            const expectedState = state;
+
+            const actualState = postsReducer({sendingPostIds: state}, action);
+            assert.strictEqual(actualState.sendingPostIds, expectedState);
```
saturninoabril: Will change to `assert.deepEqual` if suggested changes above are applied.

src/reducers/entities/posts.js
```diff
@@ -191,6 +191,42 @@ function handlePendingPosts(pendingPostIds = [], action) {
...
+        }
+
+        return [
+            ...sendingPostIds,
+            sendingPostId,
+        ];
+    }
+    case PostTypes.RECEIVED_POST: {
+        const sendingPostId = action.data.id;
+        if (!sendingPostIds.includes(sendingPostId)) {
```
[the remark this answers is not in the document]
enahum: I would leave it as there might be various posts being sent, (different channels, comments, etc..)

src/reducers/entities/posts.js
enahum: actually here, I'm with @lieut-data so it does not return a new array every time

PR #469, opened 2018-04-24
[MM-9708] Add load more at bottom of permalink view

saturninoabril:
#### Summary
This is to support loading of more messages below on permalink view.
- remove slicing of all posts since posts here are loaded on client demand (getPosts...) and is being emptied on channel switch or removal of focused post id.

(Note: `makeGetPostsAroundPost` is not being used by mobile RN so this is not expected to have breaking change with it.)

On webapp, I'll ask PM for a smoke test before releasing this for dev review.

#### Ticket Link
Jira ticket: [MM-9708](https://mattermost.atlassian.net/browse/MM-9708)

#### Test Information

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -251,29 +251,19 @@ export function makeGetPostsAroundPost() {
                 return null;
             }
 
-            const focusedPostIndex = postIds.indexOf(focusedPostId);
-            if (focusedPostIndex === -1) {
-                return null;
-            }
-
-            const desiredPostIndexBefore = focusedPostIndex - (Posts.POST_CHUNK_SIZE / 2);
-            const minPostIndex = desiredPostIndexBefore < 0 ? 0 : desiredPostIndexBefore;
-
-            const slicedPostIds = postIds.slice(minPostIndex);
-
             const posts = [];
             const joinLeavePref = myPreferences[getPreferenceKey(Preferences.CATEGORY_ADVANCED_SETTINGS, 'join_leave')];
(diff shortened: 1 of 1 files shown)


src/selectors/entities/posts.js
```diff
@@ -251,29 +251,19 @@ export function makeGetPostsAroundPost() {
...
-
-            const slicedPostIds = postIds.slice(minPostIndex);
-
             const posts = [];
             const joinLeavePref = myPreferences[getPreferenceKey(Preferences.CATEGORY_ADVANCED_SETTINGS, 'join_leave')];
             const showJoinLeave = joinLeavePref ? joinLeavePref.value !== 'false' : true;
 
-            for (let i = 0; i < slicedPostIds.length; i++) {
-                const post = allPosts[slicedPostIds[i]];
+            for (let i = 0; i < postIds.length; i++) {
```
jwilander: I'm a little worried about the performance of this for a channel that's accumulated a large number of posts. What about doing something similar to https://github.com/saturninoabril/mattermost-redux/blob/2f4df765d377f1052d7f2dc90f45c7aa7412fc87/src/selectors/entities/posts.js#L225 where we pass in the number of posts to display. In this case I think you'd need the number of posts before and after to display

That way we're not formatting 5000 posts when the user can only see 100 or so

enahum: actually @jwilander brings a very good point, and I totally missed it

enahum:
@saturninoabril is this still a valid PR?

saturninoabril:
I'm closing this since this will be covered by the on going Scrolling Overhaul epic.

PR #475, opened 2018-04-28
Add a default string for username if user is missing

sudheerDev:
#### Summary
Adding a default string `Someone` if user is missing from store when popualating username.

#### Ticket Link
[MM-10331](https://mattermost.atlassian.net/browse/MM-10331)

#### Checklist

#### Test Information


src/utils/user_utils.js
```diff
@@ -21,7 +21,7 @@ export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: Preferences.DISPLAY_PREFER_NICKNAME | Preferences.DISPLAY_PREFER_FULL_NAME
 ): string {
-    let name = '';
+    let name = General.UNKNOWN_USERNAME;
```
sudheerDev: Instead of adding default username here i could do it for typing component alone by changning below selector if you guys prefer i am 1/5 on this. https://github.com/mattermost/mattermost-redux/blob/b1e1ec91c176d895044b3923b7ce4d6de0dc03af/src/selectors/entities/typing.js#L22

There are more than few places this util function is used so, i am not sure which is more favourable a empty srting or default string.

enahum: I think this cover all cases no just for the use typing

src/constants/general.js
```diff
@@ -76,4 +76,5 @@ export default {
 
     MAX_USERS_IN_GM: 8,
     MIN_USERS_IN_GM: 3,
+    UNKNOWN_USERNAME: 'Someone',
```
lieut-data: Do we need to localize `Someone`? (I note `is typing` is a localized string.)

src/utils/user_utils.js
lieut-data: This seems like a straightforward solution! @sudheerDev, can you add test coverage for this path?

jasonblais:
@sudheerDev Big thanks for working on this issue.

We're approaching the v4.10 release, and hence would like to fix this issue by end of this week. Let me know if you have time to work on it - if not, no worries as one of our team members can help take a look.

sudheerDev:
@jasonblais I understand. I have some time today. I will try and fix my commit. If I don't make any progress with it. I will let you know so someone can grab it.

src/utils/user_utils.js
```diff
@@ -21,16 +22,14 @@ export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: Preferences.DISPLAY_PREFER_NICKNAME | Preferences.DISPLAY_PREFER_FULL_NAME
 ): string {
-    let name = '';
+    let name = localizeMessage('channel_loader.someone', 'Someone');
```
sudheerDev: In order to test this, we need to mock this import. Should we add a library for this?

test/utils/i18n_utils.test.js
```diff
@@ -0,0 +1,21 @@
...
+
+import {setLocalizeFunction, localizeMessage} from 'utils/i18n_utils';
+
+describe('i18n utils', () => {
+    it('should return default message', () => {
+        assert.equal(localizeMessage('someting.string', 'defaultString'), 'defaultString');
+    });
+
+    it('should return previously set Localized function return value', () => {
+        function mockFunc() {
```
sudheerDev: I couldn't think of way to mock this without a spy lib such as Sinon. So added this as a hack instead

test/utils/user_utils.test.js
```diff
@@ -0,0 +1,35 @@
...
+        assert.equal(displayUsername(userObj, Preferences.DISPLAY_PREFER_NICKNAME), 'nick');
+    });
+
+    it('should return fullname', () => {
+        assert.equal(displayUsername(userObj, Preferences.DISPLAY_PREFER_FULL_NAME), 'test user');
+    });
+
+    it('should return default username string', () => {
+        let noUserObj;
+        setLocalizeFunction(null);
```
sudheerDev: If I add a spy lib for the above test case then this can be removed

sudheerDev:
Feel free to review/add changes to the commit.

src/utils/user_utils.js
```diff
@@ -21,16 +22,14 @@ export function displayUsername(
...
 
     if (user) {
         if (teammateNameDisplay === Preferences.DISPLAY_PREFER_NICKNAME) {
             name = user.nickname || getFullName(user);
         } else if (teammateNameDisplay === Preferences.DISPLAY_PREFER_FULL_NAME) {
             name = getFullName(user);
-        }
```
lieut-data: Do we still want the `name.trim().length` check, to keep "Someone" in the event the name really does only contain whitespace? Not sure if that's really possible or not, so it might be fine as-is.

test/utils/user_utils.test.js
lieut-data: Instead of making this test undo the changes made by the test case above, should we have an `afterEach` in the above test case that sets localize function back to `null`? (Isolating the impact of that test case's mocking?)

lieut-data:
Nice work, @sudheerDev! Thanks for pushing the envelope on this one :)

enahum:
@sudheerDev eslint issues

```
> mattermost-redux@1.2.0 check /home/ubuntu/jenkins/workspace/mmrdx/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux
> eslint --ignore-path .gitignore --ignore-pattern node_modules --quiet .


/home/ubuntu/jenkins/workspace/mmrdx/mattermost-redux-pr/src/github.com/mattermost/mattermost-redux/test/utils/user_utils.test.js
  7:9  error  'setLocalizeFunction' is defined but never used  no-unused-vars

```

src/utils/user_utils.js
sudheerDev: Trim was used for detecting empty default string if  `teammateNameDisplay` does not satisfy both the preference settings. This was added in the very first commit of the lib so no history to traceback here. Before removing trim and moving it to else my UI broke to 
screen shot 2018-05-02 at 11 21 51 pm as the default string was important for returning displayNames in most scenarios.

We could trim but it is odd doing it only in one condtion and not everywhere as this could happen for firstname and lastname as well.

sudheerDev:
@enahum Done

PR #476, opened 2018-05-02
MM-10344 & MM-10345: Adds actions to update team and channel scheme.

mkraft:
#### Summary
* Adds action to update team scheme.
* Adds action to update channel scheme.

#### Ticket Link
[MM-10344](https://mattermost.atlassian.net/browse/MM-10344) & [MM-10345](https://mattermost.atlassian.net/browse/MM-10345)

#### Checklist

#### Test Information
Chrome desktop, macOS.


src/actions/channels.js
```diff
@@ -1283,6 +1283,32 @@ export function unfavoriteChannel(channelId) {
     };
 }
 
+export function updateChannelScheme(channelId, schemeId) {
+    return async (dispatch, getState) => {
+        dispatch({type: ChannelTypes.UPDATE_CHANNEL_SCHEME_REQUEST}, getState);
```
jespino: It's there any reason to not use the `bindClientFunc`?

src/client/client4.js
```diff
@@ -1126,6 +1145,17 @@ export default class Client4 {
...
     };
 
+    updateChannelScheme = async (channelId, schemeId) => {
+        const patch = {scheme_id: schemeId};
+
+        this.trackEvent('api', 'api_channels_update_scheme', {channel_id: channelId, ...patch});
+
+        return this.doFetch(
+            `${this.getChannelSchemeRoute(channelId)}`,
+                    {method: 'put', body: JSON.stringify(patch)}
```
jespino: Looks like it has some indentation problem

src/actions/channels.js
mkraft: Yes, the reason is that `bindClientFunc` forwards the json response from the server to the reducer, so in a case such as this with a response body of `{"status": "OK"}` one wouldn't have any useful parameters to use in the reducer. One must manually create the parameters for the reducer as done on line 1304.

jespino: Yes, you are right.

mkraft: Come to think of it one *could* still use `bindClientFunc` and wrap the `clientFunc` parameter in a custom function that ultimately returns the data in the format one wants. That is another option... I'm going to try it and see if it saves keystrokes.

src/actions/channels.js
```diff
@@ -1283,6 +1283,18 @@ export function unfavoriteChannel(channelId) {
...
+        async () => {
+            await Client4.updateChannelScheme(channelId, schemeId);
+            return {channelId, schemeId};
+        },
+        ChannelTypes.UPDATE_CHANNEL_SCHEME_REQUEST,
+        [ChannelTypes.UPDATE_CHANNEL_SCHEME_SUCCESS, ChannelTypes.UPDATED_CHANNEL_SCHEME],
+        ChannelTypes.UPDATE_CHANNEL_SCHEME_FAILURE,
+    );
+}
+
```
mkraft: @jespino Your question about `bindClientFunc` made me think of a cleaner way to write this. However, `bindClientFunc` strips out your data if you have the word `SUCCESS` in your action! So that was a little odd to debug.

src/client/client4.js
```diff
@@ -822,6 +830,17 @@ export default class Client4 {
         );
     };
 
+    updateTeamScheme = async (teamId, schemeId) => {
+        const patch = {scheme_id: schemeId};
+
+        this.trackEvent('api', 'api_teams_update_scheme', {team_id: teamId, ...patch});
```
grundleborg: I think the diagnostics should be explicit in what it tracks, otherwise this could potentially in future end up accidentally leaking data that is inappropriate to track.

mkraft: @grundleborg Can you be more specific?

grundleborg: Sorry, didn't explain that at all clearly.

The `...patch` in a call to `trackEvent()` worries me slightly, because if an extra field were to be added to that `patch` object that it isn't appropriate to send back in the telemetry data, it would get sent unless the person adding it remembered to update these `trackEvent` calls. I know it's unlikely, but where telemetry is concerned, I'd prefer to be explicit about which fields are getting sent.

mkraft: So your concern is the use of the spread operator? You know it's defined explicitly in the line above right? It's not some magic parameter.

grundleborg: Oh yeah, I missed that (assumed it was a parameter being passed into the function). In which case, no objection here.

PR #480, opened 2018-05-03
MM-10454: Removing Set usage in store for roles

jespino:
#### Summary
Sets serialization is not supported by the standard `JSON.stringify`, so, to be sure, I removed any "Set" used for roles in the store.

I'm not 100% sure that is fixing the problem.

#### Ticket Link
[MM-10454](https://mattermost.atlassian.net/browse/MM-10454)

### DIFF
--- a/src/actions/roles.js
+++ b/src/actions/roles.js
@@ -65,12 +65,12 @@ export function loadRolesIfNeeded(roles) {
             pendingRoles.add(role);
         }
         if (!state.entities.general.serverVersion) {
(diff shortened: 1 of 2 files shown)


jasonblais:
@jespino Do we need this for v4.10? I'm not sure what the impact of this error is

jespino:
I think so, but I have to reproduce the bug that it solves, and verify that is solved.

jespino:
@enahum thanks, I will take a look to the cleanup middleware.

PR #482, opened 2018-05-04
MM-10464: Handle all cases display username

jespino:
#### Summary
In some cases the display name wasn't showed because it generates an empty
string, now if that happens it use the username as final fallback.

#### Ticket Link
[MM-10464](https://mattermost.atlassian.net/browse/MM-10464)

#### Checklist

#### Test Information
This PR was tested on: [Firefox, Arch Linux]

### DIFF
--- a/src/utils/user_utils.js
+++ b/src/utils/user_utils.js
@@ -32,6 +32,10 @@ export function displayUsername(
         } else {
             name = user.username;
         }
+
+        if (!name.trim().length) {
+            name = user.username;
+        }
     }
 
     return name;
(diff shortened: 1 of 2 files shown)


sudheerDev:
Aah man, I should have gone with some defensive programming with my earlier change.

jespino:
No problem @sudheerDev the error was subtle.

jespino:
@cpanato I already have 2 aprovals, so I going to merge without your review, because we want to fix this in master/pre-release quickly.

PR #490, opened 2018-05-07
MM-8622: Improved plugin error reporting

lieut-data:
#### Summary
This represents the implementation effort of a design proposed in ~developer-toolkit on pre-release.mattermost.com.

Here's the main take aways from the redux PR. More details are given in the accompanying [server](https://github.com/mattermost/mattermost-server/pull/8737) and [webapp](https://github.com/mattermost/mattermost-webapp/pull/1180) PRs:
* New reducers to manage the plugin statuses
* Mirrored (but technically unused) websocket event handling for the new plugin statuses. (We really need to clean up the websocket handling to converge on Redux where appropriate.)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-8622

#### Checklist

#### Test Information
This PR was tested on: Chrome


src/reducers/entities/admin.js
```diff
@@ -1,9 +1,11 @@
 // Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
 // See License.txt for license information.
 
+/*eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }]*/
```
lieut-data: Not sure if we should promote this to be default, but it allow me to write:
```
const {[pluginId]: plugin, ...nextState} = state;
return nextState;
```

to remove the plugin. Maybe this is just being too "clever". Let me know what you think ;P

src/actions/admin.js
```diff
@@ -380,17 +380,43 @@ export function getPlugins(): ActionFunc {
     };
 }
 
+// EXPERIMENTAL - SUBJECT TO CHANGE
+export function getPluginStatuses(): ActionFunc {
```
jespino: Is there any reason to not use `bindClientFunc`?

src/reducers/entities/admin.js
jespino: I see it a bit confusing, you are making an assignation to discard data. I don't see it. And I don't see promoting the rule to default, I prefer false positives than false negatives.

The way that we are doing this now is something like
```
nextPosts = {...state};
Reflect.deleteProperty(nextPosts, pluginId);
```

It's way more explicit and enough compact.

jwilander: Yeah generally I agree with @jespino on this one, I'd rather be more explicit (only 1/5 though). I must say it is quite clever though :P

src/reducers/entities/admin.js
```diff
@@ -389,6 +391,111 @@ function plugins(state = {}, action) {
...
+
+    case AdminTypes.RECEIVED_PLUGIN_STATUSES: {
+        const nextState = {};
+
+        for (const plugin of (action.data || [])) {
+            const id = plugin.plugin_id;
+
+            // The plugin may be in different states across the cluster. Pick the highest one to
+            // surface an error.
+            const pluginState = Math.max((nextState[id] && nextState[id].state) || 0, plugin.state);
```
jwilander: I like the enumeration ordered by priority, very clean way to handle this 👍

src/actions/admin.js
lieut-data: Hah! I hadn't looked earlier in the file to see that pattern.

I can use it here, but without modifying the helper, I'll need to continue with this pattern so that I can pass the `pluginId` affected by the other actions.

src/reducers/entities/admin.js
lieut-data: Agreed! Excellent feedback: thank you :)

src/actions/admin.js
jespino: This https://github.com/mattermost/mattermost-redux/pull/476/files#diff-11fe928fa6662326b67b6a3e602b987dR1286 was the solution from @mkraft to that problem.

PR #492, opened 2018-05-08
Switches from SI to binary prefix.

mkraft:
#### Summary
Switches from the SI prefix MB to the binary prefix MiB in cases where the underlying value represents 1024^2 bytes.

#### Ticket Link
n/a

#### Checklist
n/a

#### Test Information
n/a

### DIFF
--- a/src/utils/file_utils.js
+++ b/src/utils/file_utils.js
@@ -8,10 +8,10 @@ import mimeDB from 'mime-db';
 export function getFormattedFileSize(file) {
     const bytes = file.size;
     const fileSizes = [
-        ['TB', 1024 * 1024 * 1024 * 1024],
-        ['GB', 1024 * 1024 * 1024],
-        ['MB', 1024 * 1024],
-        ['KB', 1024],
+        ['TiB', 1024 * 1024 * 1024 * 1024],
+        ['GiB', 1024 * 1024 * 1024],
+        ['MiB', 1024 * 1024],
+        ['KiB', 1024],
     ];
     const size = fileSizes.find((unitAndMinBytes) => {
         const minBytes = unitAndMinBytes[1];


enahum:
I understand the difference between the two but the result is normally presented to the end user who has no idea and it might be confusing for them, so my question here is why do we want to make this change?

mkraft:
@enahum Disambiguation.

enahum:
Can we run this by a PM?

mkraft:
Of course.

jasonblais:
What is the end user impact? I don't really understand it. Is it that file size is now shown as "50MiB" instead of "50MB"?

mkraft:
@jasonblais Yes `MB` or *mega*byte refers to 1,000,000 (mega) bytes whereas `MiB` *mebi*byte is the ISO standard for referring to 1,048,576 bytes.

enahum:
Exactly but my guess is that end users do not really know the difference between the two.

I’m 1/5 on this but is not a bug deal to change it, but perhaps we should change it everywhere in help text and stuff

mkraft:
@enahum I have PRs for [the server](https://github.com/mattermost/mattermost-server/pull/8690) and [the webapp](https://github.com/mattermost/mattermost-webapp/pull/1190). I can also make the changes to all the docs.

enahum:
Cool will leave PM to make the decision on this ;)

jasonblais:
I'm not sure, I haven't really seeing `MiB` used anywhere but also don't know enough to make a decision here.

I've deferred to @esethna since it's an end user related change

esethna:
Agree with JB and EN here. While it may technically be more accurate, I've never seen "MiB" facing the end user in other software and I don't think the difference between MiB and MB is not commonly known. I would prefer that we stick what's standard and well known.

PR #493, opened 2018-05-09
[MM-10430] Add selector to check if the channel is hidden

saturninoabril:
#### Summary
Add selector to check if the channel is hidden

(Didn't added a test as I'm not sure how to effectively test it. Will try to revisit next time)

Server PR - https://github.com/mattermost/mattermost-server/pull/8751
Webapp PR - https://github.com/mattermost/mattermost-webapp/pull/1194

#### Ticket Link
Jira ticket: [MM-10430](https://mattermost.atlassian.net/browse/MM-10430)


saturninoabril:
Update: While working on Mobile RN, I realized that there are some store related implementation in webapp that is needed by mobile RN as well. This update moves the "hiddenDefaultChannelId" here which was initially introduced at webapp.

@lieut-data Sorry but you might want to take a look again.

lieut-data:
@saturninoabril, will do! Sometimes I miss comments, though, so feel free to always re-add me as a reviewer to put it in my https://github.com/pulls/review-requested list :)

test/actions/channels.test.js
```diff
@@ -1833,4 +1833,21 @@ describe('Actions.Channels', () => {
...
+        let state = store.getState();
+        let hiddenChannel = state.entities.channels.hiddenDefaultChannelId;
+        assert.ok(hiddenChannel);
+        assert.ok(hiddenChannel[teamId] === channelId);
+
+        Actions.removeHiddenDefaultChannel(teamId, channelId)(store.dispatch, store.getState);
+        state = store.getState();
+        hiddenChannel = state.entities.channels.hiddenDefaultChannelId;
+        assert.ok(hiddenChannel);
+        assert.ok(!hiddenChannel[teamId]);
```
lieut-data: Worth adding test cases for when the `teamId` or `channelId` is `falsy` when passed to the action creators?

src/reducers/entities/channels.js
```diff
@@ -382,6 +382,27 @@ function stats(state = {}, action) {
     }
 }
 
+function hiddenDefaultChannelId(state = {}, action) {
```
lieut-data: Worth adding a test case for this reducer as well?

saturninoabril:
Thanks @hmhealey and @lieut-data!
Sorry for not giving update on this PR which I'm still in the process of updating.  Actually you (@hmhealey) and @enahum have the same concern about storing that ID. 
Will just ping both of you whenever this PR is ready for dev review :)

saturninoabril:
@lieut-data @hmhealey I've updated the PR and just retain the selector `shouldHideDefaultChannel`. I initially add other selectors before when I was trying to figure out how to make the default channel become available at "More Channels" modal but now that it's not needed per ticket, I've removed all of those.

PR #494, opened 2018-05-10
Fix blank usernames and commented on someone post in RN

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10558
https://mattermost.atlassian.net/browse/MM-10559

### DIFF
--- a/src/utils/user_utils.js
+++ b/src/utils/user_utils.js
@@ -22,9 +22,10 @@ export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: string
 ): string {
-    let name = localizeMessage('channel_loader.someone', 'Someone');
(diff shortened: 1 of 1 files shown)


src/utils/user_utils.js
```diff
@@ -22,9 +22,10 @@ export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: string
 ): string {
-    let name = localizeMessage('channel_loader.someone', 'Someone');
+    let name = '';
 
     if (user) {
+        name = localizeMessage('channel_loader.someone', 'Someone');
```
jwilander: How does this work? It looks like 'Someone' will always get overridden here by the else of the if statement below

PR #495, opened 2018-05-10
Prevent duplicate flagged posts

enahum:
#### Summary
Sometimes when flagging a post it can be added twice to the flagged results this PR prevents that from happening

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10333


lieut-data:
Is the basic gist here that the server might send the updated preferences delta twice?

enahum:
Actually is cause when you flag a post it dispatches a `RECEIVED_PREFERENCES` and then the websocket also does it, so it was being included twice

PR #497, opened 2018-05-13
MM-8788: Fixing small problems with scheme redux actions/reducers

jespino:
#### Summary
The delete action wasn't removing the scheme from the list of schemes on success

#### Ticket Link
[MM-8788](https://mattermost.atlassian.net/browse/MM-8788)

#### Test Information
This PR was tested on: [Arch Linux]


src/actions/schemes.js
```diff
@@ -40,17 +44,30 @@ export function createScheme(scheme: Scheme) {
...
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: SchemeTypes.DELETE_SCHEME_FAILURE, error},
+                logError(error)(dispatch),
+            ]), getState);
+            return {data: null, error};
+        }
+
+        dispatch({type: SchemeTypes.DELETED_SCHEME, data: {schemeId}}, getState);
```
jespino: Here is the real change. Here we send the schemeId in the DELETE_SCHEME event to properly remove the scheme from the store on success.

src/actions/schemes.js
```diff
@@ -40,17 +44,30 @@ export function createScheme(scheme: Scheme) {
     );
 }
 
-export function deleteScheme(schemeId: string) {
-    return bindClientFunc(
-        Client4.deleteScheme,
-        SchemeTypes.DELETE_SCHEME_REQUEST,
-        [SchemeTypes.DELETED_SCHEME, SchemeTypes.DELETE_SCHEME_SUCCESS],
-        SchemeTypes.DELETE_SCHEME_FAILURE,
-        schemeId
```
mkraft: `bindClientFunc` wasn't sending `schemeId`?

jespino: bindClientFunc returns the data returned from the server, and the server doesn't return anything on delete.

mkraft: @jespino Right. The server deletion was tested but not the deletion from redux. Good fix.

PR #501, opened 2018-05-16
Fix small bug in patchScheme action and adding rolesById in the store

jespino:
#### Summary
This changes are needed for properly manage custom team schemes in the admin panel.

#### Ticket Link
[MM-8789](https://mattermost.atlassian.net/browse/MM-8789)

#### Test Information
This PR was tested on: [Arch linux]


jespino:
Sure, the reason was basically that the roles where there by name, but the schemes are using ids to reference the roles, so, I added the rolesById. But I think you make a good point here, the data duplication, I think is not a big deal because the size of the roles list. But at the same time, is pretty easy to have an id:name binding, and return properly in the selector. So I going to make the change.

jespino:
Thanks @mkraft now the solution is way simpler :)

PR #502, opened 2018-05-16
[MM-10519] Updated handleChannelUpdatedEvent

saturninoabril:
#### Summary
Updated handleChannelUpdatedEvent to use `channel_id` instead of `channel` object from a web socket event

Server PR - https://github.com/mattermost/mattermost-server/pull/8798

Note: Once merged, mobile RN's should be updated to reflect the corresponding commit hash. 

#### Ticket Link
Jira ticket: [MM-10519](https://mattermost.atlassian.net/browse/MM-10519)

#### Checklist

#### Test Information
This PR was tested on: [MacOC|Chrome/FF, iOS 11.2 iPhone 8 simulator]


src/actions/websocket.js
```diff
@@ -552,27 +553,15 @@ function handleChannelDeletedEvent(msg, dispatch, getState) {
...
-            data: channel,
-        });
-
-        if (currentChannelId === channel.id) {
-            // Emit an event with the channel received as we need to handle
-            // the changes without listening to the store
-            EventEmitter.emit(WebsocketEvents.CHANNEL_UPDATED, channel);
-        }
+    const {currentChannelId} = getState().entities.channels;
+    if (currentChannelId === channelId) {
```
lieut-data: This seems racey: what happens if the `currentChannelId` changes before the `await` completes above?

src/actions/websocket.js
```diff
@@ -552,27 +553,15 @@ function handleChannelDeletedEvent(msg, dispatch, getState) {
...
+        // Emit an event with the channel received as we need to handle
+        // the changes without listening to the store
+        EventEmitter.emit(WebsocketEvents.CHANNEL_UPDATED, channel);
```
lieut-data: Is it no longer necessary to unconditionally emit a `RECEIVED_CHANNEL` event?

saturninoabril: It's being used by `mattermost-mobile` - https://github.com/mattermost/mattermost-mobile/blob/master/app/components/channel_drawer/channel_drawer.js#L82

saturninoabril: Did you mean dispatch `RECEIVED_CHANNEL`?  If that's the case, it's dispatched above with [getChannel](https://github.com/mattermost/mattermost-redux/blob/master/src/actions/channels.js#L414)

src/actions/websocket.js
saturninoabril: If the user switches to a different channel then there's no need to update the user with new channel info (in case of [mobile RN](https://github.com/mattermost/mattermost-mobile/blob/master/app/components/channel_drawer/channel_drawer.js#L82) - updating display name is not necessary).

jwilander:
See https://github.com/mattermost/mattermost-server/pull/8798/files#r190234927

jasonblais:
@jwilander Should be ready for re-review

PR #505, opened 2018-05-18
MM-10666: ignore unknown profiles in `sortAndInjectProfiles`

lieut-data:
#### Summary
The `profileSet` might reference a user id that is not yet known to the `profiles` -- in that case, filter out the profile and wait for the reducer to update with same.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10666

#### Checklist

#### Test Information
This PR was tested on: Chrome

### DIFF
--- a/src/selectors/entities/users.js
+++ b/src/selectors/entities/users.js
@@ -192,6 +192,8 @@ function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
         currentProfiles = Array.from(profileSet).map((p) => profiles[p]);
     }
 
+    currentProfiles = currentProfiles.filter((profile) => Boolean(profile));
+
     if (skipInactive) {
         currentProfiles = currentProfiles.filter((profile) => !(profile.delete_at && profile.delete_at !== 0));
     }


lieut-data:
Note that this doesn't actually fix the user not appearing in the list, despite the count increasing. The count is extracted separately from updated channel stats, but because the "team member" hasn't yet been seen, the addition to the channel gets filtered out by the connector. The second issue is to be tackled as part of https://mattermost.atlassian.net/browse/MM-10546.

src/selectors/entities/users.js
```diff
@@ -192,6 +192,8 @@ function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
         currentProfiles = Array.from(profileSet).map((p) => profiles[p]);
     }
 
+    currentProfiles = currentProfiles.filter((profile) => Boolean(profile));
+
```
mkraft: I'm not sure what the possible values of `profile` might be but be warned:

https://jsbin.com/kifinulaqa/edit?js,console

lieut-data: Yep, this is just a filter on `falsy` check, specifically `undefined` in this case. 👍

PR #512, opened 2018-05-26
Moved getFlaggedPosts action from webapp to redux

lisakycho:
#### Summary
Moved getFlaggedPosts action to mattermost-redux

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/8497

#### Checklist

#### Test Information
This PR was tested on: macOS High Sierra

### DIFF
--- a/src/action_types/posts.js
+++ b/src/action_types/posts.js
@@ -40,6 +40,10 @@ export default keyMirror({
     GET_POSTS_AFTER_SUCCESS: null,
     GET_POSTS_AFTER_FAILURE: null,
 
+    GET_POSTS_FLAGGED_REQUEST: null,
+    GET_POSTS_FLAGGED_SUCCESS: null,
+    GET_POSTS_FLAGGED_FAILURE: null,
+
     REACTION_REQUEST: null,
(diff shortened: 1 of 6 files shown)


enahum:
Hmmm this action already exists

https://github.com/mattermost/mattermost-redux/blob/master/src/actions/search.js#L88

lisakycho:
@enahum Weird, the webapp doesn't seem to be using the getFlaggedPost from /seach.js and was still getting called from webapp actions. 
@jasonblais should I just changed the webapp to call this action from /search.js then?

enahum:
The mobile app does, and I think the webapp should use that action instead

jwilander:
Yeah if there's already an existing action we should use it but also update it to also do what your new action here does

enahum:
@lisakycho just a follow up,  are you going to make the changes with @jwilander suggestion?

enahum:
is this still something you want to push through @lisakycho ?

lisakycho:
@enahum Sorry for the late reply! I won't be able to do this until at least after the summer!

PR #515, opened 2018-05-30
MM-9547 Added selector for CustomUrlSchemes setting

hmhealey:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-9547

#### Checklist


hmhealey:
I made a quick change to move the default url schemes into redux so that they can be shared more easily between the mobile and web apps

PR #522, opened 2018-06-06
Disable fetch calls when offline

csduarte:
#### Summary
Set's Client4 online status,  when `this.online == false` avoid making fetch calls.
Redux actions will reconcile the error.

This is to avoid the `OutOfMemoryError` caused on Android.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10773

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -34,6 +34,7 @@ export default class Client4 {
         this.userId = '';
         this.diagnosticId = '';
         this.includeCookies = true;
+        this.online = true;
 
         this.translations = {
             connectionError: 'There appears to be a problem with your internet connection.',
@@ -47,6 +48,10 @@ export default class Client4 {
 
     setUrl(url) {
         this.url = url;
+    }
+
+    setOnline(online) {
+        this.online = online;
     }
 
     setUserAgent(userAgent) {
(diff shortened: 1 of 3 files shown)


miguelespinoza:
@enahum and @hmhealey here's the fix for OOM error

src/client/client4.js
```diff
@@ -2372,6 +2377,13 @@ export default class Client4 {
     // Client Helpers
 
     doFetch = async (url, options) => {
+        if (!this.online) {
```
hmhealey: Shouldn't this be in `doFetchWithResponse`?

hmhealey: I'd just expect this check to cover all fetch calls made by Client4. Having it not cover a few of them seems weird to me

miguelespinoza: Sounds fair.  I'll test to make sure everything looks good moving to ` doFetchWithResponse`

miguelespinoza:
@enahum and @hmhealey done 😄

jwilander:
@enahum if this is for 5.0 we'll need to update the base branch

enahum:
@jwilander this is for 1.9, anyway I could update the base branch to `release-5.0` and do the same for the RN app.

thoughts??

jwilander:
Yeah let's do that 👍

PR #527, opened 2018-06-08
Safely trying to instantiate Set.

mkraft:
#### Summary
Fix for set throwing an error when being instantiated with an object.

#### Ticket Link
n/a

#### Checklist

#### Test Information
iOS Emulator

### DIFF
--- a/src/actions/roles.js
+++ b/src/actions/roles.js
@@ -60,7 +60,12 @@ export function setPendingRoles(roles) {
 export function loadRolesIfNeeded(roles) {
     return async (dispatch, getState) => {
         const state = getState();
-        const pendingRoles = new Set(state.entities.roles.pending);
+        let pendingRoles = new Set();
+        try {
+            pendingRoles = new Set(state.entities.roles.pending);
+        } catch (e) {
+            // eslint-disable-line
+        }
         for (const role of roles) {
             pendingRoles.add(role);
         }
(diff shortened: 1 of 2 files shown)


enahum:
don't know why the tests do not finish, going to close and re-open just in case

src/actions/roles.js
```diff
@@ -60,7 +60,12 @@ export function setPendingRoles(roles) {
 export function loadRolesIfNeeded(roles) {
     return async (dispatch, getState) => {
         const state = getState();
-        const pendingRoles = new Set(state.entities.roles.pending);
+        let pendingRoles = new Set();
+        try {
+            pendingRoles = new Set(state.entities.roles.pending);
+        } catch (e) {
+            // eslint-disable-line
```
hmhealey: Can we include a `console.warn` here? I don't think we should completely silence this error

enahum: I'm ok with that as long as someone actually investigates why sometimes the value of `...roles.pending` is an object

mkraft: We know it's happening a lot; we know we need to investigate. What's the value of adding a warning to everyone's console?

enahum: I'm 0/5 on the decision

hmhealey: I'd prefer to have a warning so that we actually go and fix the root problem here. Ideally, we'd fix whatever is causing `state.entities.roles.pending` to be an object right now, but I understand there's not time to do that for 1.9

hmhealey: Does it maybe need to be added to the set transformer in `app/store/index.js` like the transformers for the other Set parts of the reducer?

enahum: @hmhealey that is another PR that goes along with this one that I have ready to submit once this one gets merged

PR #535, opened 2018-06-14
MM-8701 Load client config after logging in

hmhealey:
This is for an upcoming security improvement to only provide a limited version of the client config to non-logged-in users, so we need to be sure to pull down the complete client config after logging in.

Any thoughts on whether we need the version check here or not? It's not like it hurts to pull the client config twice on an older server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-8701


src/actions/users.js
```diff
@@ -166,14 +167,16 @@ function completeLogin(data) {
...
+            dispatch(getMyPreferences()),
+            dispatch(getMyTeams()),
         ];
 
-        const state = getState();
-        const serverVersion = Client4.getServerVersion();
-        if (!isMinimumServerVersion(serverVersion, 4, 7) && getConfig(state).EnableCustomEmoji === 'true') {
+        if (isMinimumServerVersion(serverVersion, 5, 0)) { // should probably be 5.1 depending on when the version is updated on master
+            promises.push(dispatch(getClientConfig()));
+        } else if (!isMinimumServerVersion(serverVersion, 4, 7) && getConfig(getState()).EnableCustomEmoji === 'true') {
```
grundleborg: Why is this an else block here?

src/actions/users.js
```diff
@@ -166,14 +167,16 @@ function completeLogin(data) {
...
-            getMyTeams()(dispatch, getState),
+            dispatch(getMyPreferences()),
+            dispatch(getMyTeams()),
         ];
 
-        const state = getState();
-        const serverVersion = Client4.getServerVersion();
-        if (!isMinimumServerVersion(serverVersion, 4, 7) && getConfig(state).EnableCustomEmoji === 'true') {
+        if (isMinimumServerVersion(serverVersion, 5, 0)) { // should probably be 5.1 depending on when the version is updated on master
+            promises.push(dispatch(getClientConfig()));
```
grundleborg: I'm 0/5 on the fetching the config twice on older servers approach. I guess it'll only really affect RN so best to get the opinion of the experts on that I think.

src/actions/users.js
hmhealey: If the version is >= 5.0, then it will fail the second check anyway since that's looking for a version before 4.7. I could change it to make the code clearer though

src/actions/users.js
lieut-data: Maybe out of scope for this review, but I know we talked about avoiding version checks in favour of checking capability flags or some such. I'm not too concerned about double fetching, so if that got rid of the version check and complexity, I'd be in favour of that.

lieut-data:
Related to this: the server emits configuration change websocket events. Do we need to sanitize those as well?

src/actions/users.js
hmhealey: Okay, I'll go change that then to keep everything simple

hmhealey:
No, we don't need to sanitize the websocket events since you only get those when you're logged in. I didn't think to check that earlier so good catch.

hmhealey:
@grundleborg Changes have been made

PR #536, opened 2018-06-18
Fix make bundle command

jespino:
#### Summary
Remove the client.js (version 3) from the webpack.config.js.

#### Checklist

#### Test Information
This PR was tested on: [Arch Linux]

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ $ cd mattermost-redux
 $ make bundle
 ```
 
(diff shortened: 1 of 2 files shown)


sudheerDev:
Just one thing should this line in readme be updated?

`This will generate lib/mattermost.client.js, lib/mattermost.client4.js, and lib/mattermost.websocket.js which can be loaded in a browser. Also note that babel-polyfill is required``

jespino:
@sudheerDev very good point

jespino:
Fixed, and after the test passes I'll merge it :)

PR #537, opened 2018-06-18
fix jenkinspipeline

cpanato:
#### Summary
@jwilander @lieut-data 
sorry for the changes, I realized that previous one does not work with the directive `when`. so changing the pipeline to be declarative


cpanato:
the build will not pass, because it will use the existing jenkins file in master and not this one. This is to avoid external users to try steal information from jenkins

PR #544, opened 2018-06-20
Gfycat integration

kennethjeremyau:
#### Summary
Updated the emoji picker to also select and post gifs from Gfycat!

#### Checklist

#### Test Information
This PR was tested on: [Macbook Air (macOS High Sierra), AWS EC2 t2.mediium (Ubuntu 16.04) ]


src/utils/gfycatSdk.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import Gfycat from './gfycat-sdk.umd.bundle.min';
+
+export default new Gfycat({
```
lieut-data: Since Mattermost is not a single SAAS offering, I'm wondering if this needs to be customizable by the end user deploying gfycat.

src/actions/gifs.js
```diff
@@ -0,0 +1,604 @@
...
+        const url = `https://px.gfycat.com/px.gif?${queryString}`;
+        fetch(url);
+    };
+}
+
+export function splitBatchRequest({gfycats, params, dispatch}) {
+    dispatch(sendViewCountBatch({gfycats, params}));
+}
+
+export function sendEvent({event, params}) {
```
lieut-data: I understand this was discussed at the design level, but I'm concerned by the event tracking here.

1) Is this required to be allowed to use the Gfycat API? (If so, I think this lends credence to needing to put the onus on the enduser to acquire a Gfycat credentials as asked below.)
2) If required, but the end user is required to configure this, is it required to report metrics at the granularity designed here? I'm concerned that this would unnecessarily expose user count/engagement. Is it sufficient to ping the metrics endpoint instead (recognizing that IP address leakage still occurs)?

src/utils/gif_utils.js
```diff
@@ -0,0 +1,68 @@
...
+        return;
+    }
+    let expires = '';
+    if (exseconds) {
+        const d = new Date();
+        d.setTime(d.getTime() + (exseconds * 1000));
+        expires = `expires=${d.toUTCString()};`;
+    }
+
+    document.cookie = `${name}=${value};${expires}domain=.gfycat.com;path=/;`;
```
lieut-data: Independent of the security concerns above, I don't think domain=.gfycat.com even works, since this application isn't being served from .gfycat.com.

src/actions/gifs.js
```diff
@@ -0,0 +1,604 @@
...
+
+        const queryString = queryParamsString(extendedParams);
+
+        const url = `https://metrics.gfycat.com/pix.gif?${queryString}`;
+        fetch(url);
+    };
+}
+
+function queryParamsString(params) {
+    return params ? Object.keys(params).map((key) => {
```
hmhealey: Can this function be refactored like
```
if (!params) {
    return '';
}

return Object.keys(params)...;
```
to follow our conventions?

src/actions/gifs.js
```diff
@@ -0,0 +1,604 @@
...
+                dispatch(receiveSearchById(gfyId, response.gfyItem));
+                dispatch(cacheGifsRequest([response.gfyItem]));
+            }
+        ).catch((err) => dispatch(errorSearchById(err, gfyId)));
+    };
+}
+
+export function shouldSearchById(state, gfyId) {
+    const gif = state.cache.gifs[gfyId];
+    if (!gif) {
```
hmhealey: This could be simplified to `return !gif;`

src/reducers/entities/gifs.js
```diff
@@ -0,0 +1,268 @@
...
+    [GifTypes.CATEGORIES_LIST_FAILURE]: (state) => ({
+        ...state,
+        isFetching: false,
+    }),
+};
+
+const TERM_SELECTOR = {
+    [GifTypes.REQUEST_SEARCH]: (state, action) => ({
+        ...state,
+        [action.searchText]: {
```
hmhealey: What happens if the user makes a lot of different searches? Will the memory usage of this part of the store ever cause a problem, particularly on mobile where we write the store to disk?

src/utils/gfycatSdk.js
hmhealey: If we treat it like the YouTube previews, an admin would be able to set up an ID and secret in the System Console under Integrations > External Services

src/utils/gif_utils.js
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function generateUUID() {
```
hmhealey: We've got a similar `generateId` function in `utils/helpers.js` to this. Would this be able to use that instead?

src/actions/gifs.js
jasonblais: Good catch. @kennethjeremyau We have pretty strict agreements with our users on what tracking is allowed, and what data is collected - if any.

We'll need to tie this with our [Diagnostics and Error Reporting feature](https://docs.mattermost.com/administration/telemetry.html#error-and-diagnostics-reporting-feature) if we want to collect data.

src/actions/gifs.js
kennethjeremyau: Done.

src/actions/gifs.js
kennethjeremyau: I've changed to `return !state.cache.gifs[gfyId]`

src/reducers/entities/gifs.js
```diff
@@ -0,0 +1,268 @@
...
+    [GifTypes.REQUEST_CATEGORIES_LIST]: (state) => ({
+        ...state,
+        isFetching: true,
+    }),
+    [GifTypes.CATEGORIES_LIST_RECEIVED]: (state, action) => {
+        const {cursor, tags} = action;
+        const {tagsList: oldTagsList = []} = state;
+        const tagsDict = {};
+        const newTagsList = tags.filter((item) => {
+            if (item && item.gfycats[0] && item.gfycats[0].width) {
```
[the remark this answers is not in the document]
kennethjeremyau: Done.

src/utils/gfycatSdk.js
kennethjeremyau: I've brought this up to my team.

src/utils/gif_utils.js
kennethjeremyau: Yes.  I've refactored to use.

src/reducers/entities/gifs.js
kennethjeremyau: If there are a lot of different searches, then this part of the store would grow accordingly with results for each search text until refresh or store is destroyed.  We don't think this is going to be a problem because it just holds text.  Our other integrations have not seen any problems.

enahum: The store has to be manually destroyed on mobile so we should set a cap on this

hmhealey: @enahum What if we just make it so that this isn't persisted to disk on mobile? As long as the app goes into the background at some point, this should be fine

hmhealey:
I just did another review pass, and everything looks good except for having the client credentials passed from the server which I understand is in progress

src/reducers/entities/gifs.js
enahum: Yes this can be done, but it should be done as soon as this PR gets merged

src/utils/gfycatSdk.js
kennethjeremyau: I've added it to System Console under "GIF (Beta)".

src/actions/gifs.js
kennethjeremyau: I've removed our analytics and am using `client.trackEvent()`.

src/utils/gif_utils.js
kennethjeremyau: I've removed the cookie utilities.

src/utils/gfycat-sdk.umd.bundle.min.js
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable */
```
lieut-data: Is this available as an NPM package? Would be nice to be able to reference the original source and minify it as part of our build. I don't think this is a blocker.

src/utils/gfycatSdk.js
```diff
@@ -3,7 +3,11 @@
...
 
-export default new Gfycat({
-    client_id: '2_KtH_W5',
-    client_secret: '3wLVZPiswc3DnaiaFoLkDvB4X0IV6CpMkj4tf2inJRsBY6-FnkT08zGmppWFgeof',
-});
+let instance = null;
+
+export default function (key, secret) {
+    if (!instance) {
+        instance = new Gfycat({client_id: key, client_secret: secret});
```
hmhealey: Does this cause any errors if you have this enabled but haven't configured the API keys?

hmhealey: Also, this doesn't let you change the keys without restarting the client. Would doing something like this work?
```javascript
let instance = null;
let instanceKey;
let instanceSecret;

export default function (key, secret) {
    if (!instance || key !== instanceKey || secret !== instanceSecret) {
        instance = new Gfycat({client_id: key, client_secret: secret});
        instanceKey = key;
        instanceSecret = secret;
    }
    return instance;
}
```

kennethjeremyau: I updated this to work the way you say, but I'm not sure I needed to.  Switching from the System Console seems to refresh the page.

kennethjeremyau: Any blank API key/secret now defaults to specific values so this case shouldn't happen.  Normally, the SDK throws a single, custom exception that doesn't affect operation if it isn't given credentials.

hmhealey:
Looks good. I'll merge this now along with the server changes, so if you can `npm add mattermost/mattermost-redux#hash` in the web app, it'll pick up these changes and make it available for testing on a test server

PR #549, opened 2018-06-22
New selectors for sidebar reorganization

csduarte:
#### Summary
Adds support for sidebar reorganization.
PR for webapp: https://github.com/mattermost/mattermost-webapp/pull/1374

Selectors have been modified to support 2 changes.
Grouping:
* by_type, never, unreads_at_top, favorites_at_top
Sorting:
* recents, alphabetical order

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -32,7 +32,6 @@ import {
     completeDirectChannelInfo,
     completeDirectChannelDisplayName,
     getUserIdFromChannelName,
-    isChannelMuted,
     getDirectChannelName,
     isAutoClosed,
     isDirectChannelVisible,
@@ -70,6 +69,44 @@ export const getDirectChannelsSet = createSelector(
 
 export function getChannelMembersInChannels(state) {
     return state.entities.channels.membersInChannel;
+}
+
+export const mapAndSortChannelIds = (channels, currentUser, myMembers, lastPosts, sorting) => {
+    const locale = currentUser.locale || General.DEFAULT_LOCALE;
+
+    if (sorting === 'recent') {
+        channels.sort((a, b) => {
+            const aLastPostAt = (lastPosts[a.id] && lastPosts[a.id].update_at) || a.last_post_at;
+            const bLastPostAt = (lastPosts[b.id] && lastPosts[b.id].update_at) || b.last_post_at;
+
+            const aDate = new Date(aLastPostAt);
+            const bDate = new Date(bLastPostAt);
+
+            return bDate.getTime() - aDate.getTime();
+        });
+    } else {
+        channels.sort(sortChannelsByDisplayNameAndMuted.bind(null, locale, myMembers));
+    }
+
+    return channels.map((c) => c.id);
+};
+
+export function filterChannels(unreadIds, favoriteIds, channelIds, unreadsAtTop, favoritesAtTop) {
+    let channels = channelIds;
+
+    if (unreadsAtTop) {
+        channels = channels.filter((id) => {
+            return !unreadIds.includes(id);
+        });
+    }
+
+    if (favoritesAtTop) {
+        channels = channels.filter((id) => {
+            return !favoriteIds.includes(id);
+        });
+    }
+
+    return channels;
 }
 
 export function makeGetChannel() {
@@ -198,7 +235,7 @@ export const getChannelsNameMapInCurrentTeam = createSelector(
 );
 
 // Returns both DMs and GMs
-export const getDirectChannels = createSelector(
+export const getAllDirectChannels = createSelector(
     getAllChannels,
     getDirectChannelsSet,
     (state) => state.entities.users,
@@ -232,7 +269,7 @@ export const getGroupChannels = createSelector(
 
 export const getMyChannels = createSelector(
     getChannelsInCurrentTeam,
-    getDirectChannels,
+    getAllDirectChannels,
     getMyChannelMemberships,
     (channels, directChannels, myMembers) => {
         return [...channels, ...directChannels].filter((c) => myMembers.hasOwnProperty(c.id));
@@ -387,7 +424,7 @@ export const canManageChannelMembers = createSelector(
     }
 );
 
-export const getDirectChannelIds = createIdsSelector(
+export const getAllDirectChannelIds = createIdsSelector(
     getDirectChannelsSet,
     (directIds) => {
         return Array.from(directIds);
@@ -404,7 +441,7 @@ export const getChannelIdsInCurrentTeam = createIdsSelector(
 
 export const getChannelIdsForCurrentTeam = createIdsSelector(
     getChannelIdsInCurrentTeam,
-    getDirectChannelIds,
+    getAllDirectChannelIds,
     (channels, direct) => {
         return [...channels, ...direct];
     }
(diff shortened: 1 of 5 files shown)


miguelespinoza:
Hoping I can get some assistance on a failing test.

The test is failing on private channel displayName change to `abc`.
I'm fairly new to reselect, from my understanding it does not recompute the selection if the arguments are the same.  

Does this not account for deeply nested objects?

Based on the test it does not sortPrivateChannelIds properly, but I tested this from the webapp and this is the behavior.
Sorts just fine.

private channel to abc

miguelespinoza:
I see my issue.
When I was using a selector for `getPublicChannels`, `getPrivateChannels`, `getDirectMessages` I was using `createIdsSelector`.  Which only does a shallowEqual check.

Using `createSelector` fixes that issue 😥

src/selectors/entities/sidebar.js
```diff
@@ -0,0 +1,43 @@
...
+import {getBool as getBoolPreference, get as getPreference} from 'selectors/entities/preferences';
+
+const defaultSidebarPrefs = {
+    grouping: 'by_type',
+    unreads_at_top: 'true',
+    favorite_at_top: 'true',
+    sorting: 'alpha',
+};
+
+export function getSidebarPreferences(state) {
```
jwilander: Any reason this isn't a selector?

hmhealey: It looks like it should probably be a selector since it's returning an object. Something like this should work
```javascript
export const getSidebarPreferences = createSelector(
    (state) => {
        const config = getConfig(state);
        return config.ExperimentalGroupUnreadChannels !== General.DISABLED && getBoolPreference(
            state,
            Preferences.CATEGORY_SIDEBAR_SETTINGS,
            'show_unread_section',
            config.ExperimentalGroupUnreadChannels === General.DEFAULT_ON
       );
    },
    (state) => getPreference(state, Preferences.CATEGORY_SIDEBAR_SETTINGS, '', null),
    (showUnreadSection, sidebarPreferencesJson) => {
        // ...
    }
);
```

src/utils/channel_utils.js
```diff
@@ -546,6 +546,16 @@ export function sortChannelsByDisplayNameAndMuted(locale, members, a, b) {
     return -1;
 }
 
+export function sortChannelsByRecency(lastPosts, a, b) {
+    const aLastPostAt = (lastPosts[a.id] && lastPosts[a.id].update_at) || a.last_post_at;
```
hmhealey: I'm not sure I understand the logic here. Is this supposed to be equivalent to `lastPosts[a.id] ? lastPosts[a.id].update_at : a.last_post_at`

test/selectors/channels.test.js
```diff
@@ -15,35 +15,49 @@ describe('Selectors.Channels', () => {
 
     const channel1 = TestHelper.fakeChannelWithId(team1.id);
     channel1.display_name = 'Channel Name';
+    channel1.last_post_at = (new Date()).getTime();
```
hmhealey: Minor nit, this could also be `Date.now()` if you wanted to simplify it slightly

src/selectors/entities/sidebar.js
miguelespinoza: Thanks for the feedback.  I'll look at making this a selector

src/selectors/entities/channels.js
```diff
@@ -198,7 +259,7 @@ export const getChannelsNameMapInCurrentTeam = createSelector(
 );
 
 // Returns both DMs and GMs
-export const getDirectChannels = createSelector(
+export const getAllDirectChannels = createSelector(
```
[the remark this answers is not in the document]
miguelespinoza: Agreed, we're working on adding support on the mobile side as well

src/selectors/entities/channels.js
```diff
@@ -670,3 +769,117 @@ export function getGroupOrDirectChannelVisibility(state, channelId) {
         getLastPostPerChannel(state)
     );
 }
+
+const getAllActiveChannels = createSelector(
```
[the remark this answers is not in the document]
miguelespinoza: One thing I noticed is that getAllChannels gives you all of them.  Meaning, archived channels even DMs with deactivated users.

The way I get around that is to get all "active" channels meaning Public/Private channels and Direct/Group messages

src/utils/channel_utils.js
miguelespinoza: Yes, that's the idea.

If a post exists for that channel get the `updated_at` timestamp.
If not get the timestamp of the channel property `last_post_at `

I noticed that `updated_at` is not updated if a new post is added to the channel.

test/selectors/channels.test.js
miguelespinoza: Cool, thanks I'll make that change

src/selectors/entities/channels.js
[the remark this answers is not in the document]
miguelespinoza: I don't see how this would block Redux PRs.

If you look closely getDirectChannels is still a function.  It's just a different implementation than previous.

If I'm not mistaken test would break if `getDirectChannels` was not included in the selector.


You can take a look at `getDirectChannelIds`.

I think what's confusing is the order in which the functions are implemented.

It was hard to navigate at first when implementing this feature.  Since there's multiple selectors for different type of channels.
So what I did, was to reorganize the order of the functions which makes the git diff look like it's renaming functions

src/utils/channel_utils.js
```diff
@@ -546,6 +546,16 @@ export function sortChannelsByDisplayNameAndMuted(locale, members, a, b) {
     return -1;
 }
 
+export function sortChannelsByRecency(lastPosts, a, b) {
+    const aLastPostAt = lastPosts[a.id] ? lastPosts[a.id].update_at : a.last_post_at;
```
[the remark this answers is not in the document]
miguelespinoza: Yeh, I'll take a look at refactoring so it respects the latest out of the 2

src/utils/channel_utils.js
```diff
@@ -546,6 +546,16 @@ export function sortChannelsByDisplayNameAndMuted(locale, members, a, b) {
...
+    const bLastPostAt = lastPosts[b.id] ? lastPosts[b.id].update_at : b.last_post_at;
+
+    const aDate = new Date(aLastPostAt);
+    const bDate = new Date(bLastPostAt);
+
+    return bDate.getTime() - aDate.getTime();
```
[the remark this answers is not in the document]
miguelespinoza: Just tested and you're right gong to make a change to this function.

miguelespinoza:
@hmhealey thanks for the feedback.  I've made an update to the recency function to reflect what you mentioned.

Let me know if you have other questions about the reply I made for `getDirectChannels`.  I'm hoping I'm not missing anything that would cause all Redux PRs to be blocked

hmhealey:
@miguelespinoza Regarding the renaming, my main concern is if the new getDirectChannels works differently than the old one in a way that would break either client since that would prevent anyone else from making redux changes on top of it while the web and mobile PRs are still pending. If it still behaves the same, then I'm totally fine merging

miguelespinoza:
@hmhealey I had to run `master` against this branch to ensure it's not breaking anything and forgot that all selectors such as `getSortedPublicChannelWithUnreadsIds` are removed. 

Your concern is valid,  I can add those selectors back for backward compatibility and have a followup where it removes them.

After that it should be good to merge so Eric can test the webapp

miguelespinoza:
@hmhealey @jwilander I added 
* getSortedPublicChannelWithUnreadsIds
* getSortedPrivateChannelWithUnreadsIds
* getSortedFavoriteChannelWithUnreadsIds
* getSortedDirectChannelWithUnreadsIds

So it doesn't break the existing webapp behavior.  Let me know if this looks good and we can merge

PR #559, opened 2018-07-05
If the websocket url is malformed do not attempt to connect

enahum:
By preventing the WebSocket to try and connect when the url is malformed we avoid the Signal 6 crashes seen on iOS


src/client/websocket_client.js
```diff
@@ -89,6 +89,7 @@ class WebSocketClient {
             } else {
                 // If we're unable to set the origin header, the websocket won't connect, but the URL is likely malformed anyway
                 console.warn('websocket failed to parse origin from ' + connectionUrl); // eslint-disable-line no-console
+                return;
```
hmhealey: Do we need to return something or call an event handler here to show this in the UI, or will this just show the "connecting" bar?

enahum: it will show the connecting bar

PR #561, opened 2018-07-08
GH-8994 added flow assertions to roles js

pradeepmurugesan:
#### Summary
Added flow types to role.js

#### Ticket Link
[#8994](https://github.com/mattermost/mattermost-server/issues/8994)


src/types/roles.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+// @flow
+
+export type Role = {
```
jespino: Here I think is better to use exact object (with `{|...|}`).

PR #568, opened 2018-07-13
Migrate actions/helpers.js to flow

pepf:
#### Summary
Add flow typing to actions/helpers.js

#### Ticket Link
[MM-11041](https://mattermost.atlassian.net/browse/MM-11041)
https://github.com/mattermost/mattermost-server/issues/8997

#### Checklist
- [ ] ~Added or updated unit tests (required for all new features)~
- [ ] ~All new/modified APIs include changes to the drivers~

#### Test Information
This PR was tested on: [Mac OS High Sierra, 10.13.5]

### DIFF
--- a/src/actions/helpers.js
+++ b/src/actions/helpers.js
@@ -1,21 +1,27 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// @flow
 
 import {batchActions} from 'redux-batched-actions';
 import {Client4} from 'client';
 import {UserTypes} from 'action_types';
 import {logError} from './errors';
+
+import type {Client4Error} from '../types/errors';
+import type {GenericAction, ActionResult, DispatchFunc, GetStateFunc} from '../types/actions';
+type ActionType = string;
+
 const HTTP_UNAUTHORIZED = 401;
 
-export function forceLogoutIfNecessary(err, dispatch, getState) {
+export function forceLogoutIfNecessary(err: Client4Error, dispatch: DispatchFunc, getState: GetStateFunc) {
     const {currentUserId} = getState().entities.users;
-    if (err.status_code === HTTP_UNAUTHORIZED && err.url.indexOf('/login') === -1 && currentUserId) {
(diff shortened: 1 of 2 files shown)


src/actions/helpers.js
```diff
@@ -1,50 +1,58 @@
...
 // See LICENSE.txt for license information.
+// @flow
 
 import {batchActions} from 'redux-batched-actions';
 import {Client4} from 'client';
 import {UserTypes} from 'action_types';
 import {logError} from './errors';
+
+import type {Client4Error} from '../types/errors';
+import type {GenericAction, ActionResult, DispatchFunc, GetStateFunc} from '../types/actions';
```
jwilander: Can you switch these to using non-relative paths, e.g. `from 'types/actions';`? Makes it easier to refactor/move files around

jespino:
Sorry, but the types import using absolute path have implications for the future use of it as part of other projects. Probably we will need to change it in the future, but for now we are using relative path for types import everywhere. My proposal here is to keep doing that, and when we start to include it in other projects (probably in mattermost-mobile) look for the best way to do it.

jwilander:
Is this something specific to flow?

jespino:
Is not flow specific, is more related to how we compile an distribute mattermost-redux, and at the same time, how we distribute our flow code (using the copy approach). Is something that we can solve with the first mattermost-mobile/mattermost-webapp flow integration. Until then, I think is better to use only 1 approach, and the current one is use the relative paths.

jwilander:
I'm not sure I understand, we use absolute paths relative to the project root for almost all the imports in mattermost-redux. If we're going for consistency then that is what we should do.

jespino:
Consistency with the rest of the flow imports. Let my try to integrate flow in mattermost-mobile to revisit the problem by myself.

pepf:
ok, so we might be able to fix the "flow issue" using [this suggestion](https://github.com/facebook/flow/issues/4186#issuecomment-347078321), but what shall I do with the current code? Just change it to relative for now?..

jespino:
For now, lets back to the relative path, merge it, and then decide what to do with the relative/absolute path problem.

pepf:
done 👌

jespino:
Thanks :sparkles:

PR #573, opened 2018-07-17
MM-11070: Adds post ID filter selector.

mkraft:
#### Summary
Adds new selector to filter post IDs using conditions involving associated channels.

#### Ticket Link
Need as part of [MM-11070](https://mattermost.atlassian.net/browse/MM-11070).

#### Checklist

#### Test Information
iPhone 6, iOS 11.4 emulator

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -926,6 +926,9 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
 // Filters post IDs by the given condition.
 // The condition function receives as parameters the associated channel object and the post object.
 export const filterPostIds = (condition) => {
+    if (typeof condition !== 'function') {
+        throw new TypeError(`${condition} is not a function`);
+    }
     return createSelector(
         getAllChannels,
         getAllPosts,
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -934,7 +934,7 @@ export const filterPostIds = (condition) => {
         getAllPosts,
         (state, postIds) => postIds,
         (channels, posts, postIds) => {
-            return postIds.filter((postId) => {
+            return postIds.filter((postId) => { // eslint-disable-line
                 const post = posts[postId];
                 if (post) {
                     const channel = channels[post.channel_id];
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -934,14 +934,13 @@ export const filterPostIds = (condition) => {
         getAllPosts,
         (state, postIds) => postIds,
         (channels, posts, postIds) => {
-            return postIds.filter((postId) => { // eslint-disable-line
+            return postIds.filter((postId) => {
                 const post = posts[postId];
+                let channel;
                 if (post) {
-                    const channel = channels[post.channel_id];
-                    if (condition(channel, post)) {
-                        return postId;
-                    }
+                    channel = channels[post.channel_id];
                 }
+                return post && channel && condition(channel, post);
             });
         }
     );
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -934,17 +934,15 @@ export const filterPostIds = (condition) => {
         getAllPosts,
         (state, postIds) => postIds,
         (channels, posts, postIds) => {
-            const ids = [];
-            postIds.forEach((postId) => {
+            return postIds.filter((postId) => {
                 const post = posts[postId];
                 if (post) {
                     const channel = channels[post.channel_id];
                     if (condition(channel, post)) {
-                        ids.push(postId);
+                        return postId;
                     }
                 }
             });
-            return ids;
         }
     );
 };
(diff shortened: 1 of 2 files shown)


src/selectors/entities/channels.js
```diff
@@ -923,3 +923,25 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
...
+        getAllChannels,
+        getAllPosts,
+        (state, postIds) => postIds,
+        (channels, posts, postIds) => {
+            const ids = [];
+            postIds.forEach((postId) => {
+                const post = posts[postId];
+                if (post) {
+                    const channel = channels[post.channel_id];
+                    if (condition(channel, post)) {
```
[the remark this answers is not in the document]
mkraft: `condition` is not a runtime parameter.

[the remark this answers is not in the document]
enahum: thanks for adding the check @mkraft

src/selectors/entities/channels.js
```diff
@@ -923,3 +923,28 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
...
+    if (typeof condition !== 'function') {
+        throw new TypeError(`${condition} is not a function`);
+    }
+    return createSelector(
+        getAllChannels,
+        getAllPosts,
+        (state, postIds) => postIds,
+        (channels, posts, postIds) => {
+            const ids = [];
+            postIds.forEach((postId) => {
```
jespino: Why a forEach and not a `for ... in` loop, or a `.filter`?

src/selectors/entities/channels.js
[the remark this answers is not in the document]
mkraft: @enahum Actually I was wrong `condition` *could* be a runtime parameter (if you were creating the factory at runtime). Anyhow, I tested the behaviour of `Array.prototype.filter()` and it throws a `TypeError` if you try to invoke the method without a function parameter so I have added that error handling. However, it leaves the error handling of the function parameter to the user, so I have added a test that proves this code works the same way.

test/selectors/channels.test.js
```diff
@@ -753,4 +753,35 @@ describe('Selectors.Channels', () => {
...
+        } catch (e) {
+            filterErrorMessage = e.message;
+        }
+
+        const postIDs = Object.keys(posts);
+
+        assert.deepEqual(filterPostIDsByArchived(testStateC, postIDs), ['b']);
+        assert.deepEqual(filterPostIDsByUserB(testStateC, postIDs), ['c']);
+
+        assert.throws(() => Selectors.filterPostIds(), TypeError);
```
mkraft: Test added to prove `TypeError` if invoked without the `condition` function.

test/selectors/channels.test.js
```diff
@@ -753,4 +753,35 @@ describe('Selectors.Channels', () => {
...
+
+        assert.throws(() => filterPostIDsInvalid(testStateC, postIDs), ReferenceError, filterErrorMessage);
```
mkraft: Test proving that the error handling is the same as `Array.prototype.filter()`.

src/selectors/entities/channels.js
mkraft: No particular reason... are you thinking what I'm thinking, namely:
a) `for ... in` would be faster, and
b) `filter` would avoid an extra variable declaration?

mkraft:
I have disabled the `consistent-return` ESLint warning at

https://github.com/mattermost/mattermost-redux/pull/573/files#diff-f48970fd598b41ca30a1892864e3f2aeR937

because if I didn't then something like this:

```js
[1, 'b'].filter(item => {
	if (typeof item === 'string') {
		return item;
    }
});
```
would need to be changed to something like this:
```js
[1, 'b'].filter(item => {
	if (typeof item === 'string') {
		return item;
    }
    return null;
}).filter(item => item);
```
which is a useless extra iteration.

src/selectors/entities/channels.js
```diff
@@ -923,3 +923,26 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
...
+        throw new TypeError(`${condition} is not a function`);
+    }
+    return createSelector(
+        getAllChannels,
+        getAllPosts,
+        (state, postIds) => postIds,
+        (channels, posts, postIds) => {
+            return postIds.filter((postId) => { // eslint-disable-line
+                const post = posts[postId];
+                if (post) {
```
jespino: Filter expect a function that returns a boolean (it is working because the id is like a boolean in this case).

I think you can replace it with something like
```javascript
const post = posts[postId];
return post && condition(channels[post.channel_id], post)
```

jespino: And besides, you don't need to remove the es-lint rule this way.

mkraft: Seems to me that filter just handles no return correctly, try this:

```js
['hello', 1, null, {}, 'b'].filter(item => {
	if (typeof item === 'string') {
		return item;
    }
});
```
// ["hello", "b"]

jespino: Yes, filter handle no-return properly because no-returns is equal to undefined, and undefined is false. and filter expect a function that returns truish o falsish values. You are returning the item, which is at least, not what is expected by filter. Works, but is not the expected usage.

mkraft: Yeah it's an implicit return... I guess being explicit is probably better.

PR #581, opened 2018-07-26
Mattermost redux rethreading

kkekkonen:
#### Summary
Adds a new feature: rethreading. Posts not included in threads can be moved to existing threads, or a new thread can be created with another post not in threads. Rethreading can be done by the posts creator (and any system admin or team admin if he/she has permissions to edit the post).

#### Ticket Link
None. 

There is the Mattermost forum post:
https://mattermost.uservoice.com/forums/306457-general/suggestions/34136164-add-ability-to-move-mistaken-thread-replies-in-a-c

and mattermost feature channel
https://pre-release.mattermost.com/core/pl/13zruhcrfidtxqi9s6m8tu54fe

#### Checklist

#### Test Information


kkekkonen:
new changes: added stuff to reducer to trigger post-rethread replycount changes for users after rethread is complete

jwilander:
Moving this and the server PR to dev review so that the PM/designers can test the web app PR

src/constants/preferences.js
```diff
@@ -67,6 +67,7 @@ export default {
             mentionHighlightBg: '#ffe577',
             mentionHighlightLink: '#166de0',
             codeTheme: 'github',
+            postRethreadHighlight: '#cfebfd',
```
lieut-data: I have little experience with new theme preferences. Should we aim to use an existing colour here instead? Do we need a migration strategy otherwise? (Probably questions for you @hmhealey.)

hmhealey: Migration isn't a problem any more, but I feel like we should be careful about adding one-off colours like this since it's an extra option in the theme UI specifically for a this feature. Most of the other colours are reused in at least a few places, so it seems odd to have this one very specific one that people may configure

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
     return {posts: nextPosts, postsInChannel: nextPostsForChannel, postsInThread: nextPostsForThread || postsInThread};
 }
 
+function handleRethreadPost(postsMap, action) {
+    const posts = postsMap.posts ? postsMap.posts : {};
+    const postsInChannel = postsMap.postsInThread ? postsMap.postsInChannel : {};
```
hmhealey: Should this be
```
const postsInChannel = postsMap.postsInChannel ? postsMap.postsInChannel : {};
```
? You could also shorthand these as
```
const postsInChannel = postsMap.postsInChannel || {};
```

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
...
+    const postsInThread = postsMap.postsInThread ? postsMap.postsInThread : {};
+    var post = action.data.post;
```
hmhealey: Can you use `const` or `let` instead of var?

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
...
+    const nextPosts = posts;
+    nextPosts[post.id] = post;
```
hmhealey: This is making changes to the existing data in the store which we shouldn't be doing anywhere. You should be making a copy of posts like
```
const nextPosts = {...posts}; // create a shallow copy of posts
nextPosts[post.id] = post; // since nextPosts is a copy, this isn't changing the existing stored data
```

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
...
+    const rootId = action.data.original_root_id;
+
+    const nextPostsForChannel = postsInChannel;
+    let nextPostsForThread = postsInThread;
+
+    if (nextPosts[post.id]) {
```
hmhealey: This outer if statement isn't necessary since you set `nextPosts[post.id]` above.

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
...
+function handleRethreadPost(postsMap, action) {
+    const posts = postsMap.posts ? postsMap.posts : {};
+    const postsInChannel = postsMap.postsInThread ? postsMap.postsInChannel : {};
+    const postsInThread = postsMap.postsInThread ? postsMap.postsInThread : {};
+    var post = action.data.post;
+    const nextPosts = posts;
+    nextPosts[post.id] = post;
+    const rootId = action.data.original_root_id;
+
+    const nextPostsForChannel = postsInChannel;
```
hmhealey: Is something supposed to be changing with `nextPostsInChannel`? The code that affects it just changes its name repeatedly
```
function handleRethreadPost(postsMap, action) {
    const postsInChannel = postsMap.postsInThread ? postsMap.postsInChannel : {};
    ...
    const nextPostsForChannel = postsInChannel;
    ...
    return {..., postsInChannel: nextPostsForChannel, ...};
}
```

hmhealey: Also, can you rename these to `nextPostsInChannel` and `nextPostsInThread` to keep the names consistent? `postsFor` usually refers to the post IDs for a single channel like `const postsForChannel = postsInChannel[channelId]`

src/reducers/entities/posts.js
```diff
@@ -384,6 +411,8 @@ function handlePosts(posts = {}, postsInChannel = {}, postsInThread = {}, action
         return {posts, postsInChannel, postsInThread};
     case PostTypes.REMOVE_POST:
         return handleRemovePost(posts, postsInChannel, postsInThread, action);
+    case PostTypes.RETHREAD_POST_UPDATE:
```
hmhealey: This isn't the action you've called attached to the `rethreadPost` action above, so this code is never actually ran

src/reducers/entities/posts.js
```diff
@@ -359,6 +359,33 @@ function handleRemovePost(posts = {}, postsInChannel = {}, postsInThread = {}, a
...
+    nextPosts[post.id] = post;
+    const rootId = action.data.original_root_id;
+
+    const nextPostsForChannel = postsInChannel;
+    let nextPostsForThread = postsInThread;
+
+    if (nextPosts[post.id]) {
+        if (postsInThread[rootId]) {
+            nextPostsForThread = nextPostsForThread || {...postsInThread};
+            const threadPosts = [...postsInThread[rootId]];
```
hmhealey: And here's where you could use the name `nextPostsForThread` since that's usually how it's used 😛

src/constants/preferences.js
kkekkonen: I didnt have much experience with the themes and preferences so I just set the values to such that they could be tested and that the themes werent entirely horrible. I didnt spend much time on the feature

esethna:
@hmhealey @lieut-data looks like there have been updates since last review, have your comments been addressed here?

hmhealey:
Unit tests still need to be added, and there's also been a bad merge that needs to be resolved since there's a bunch of unrelated commits from master included in here as well

esethna:
cc// @kkekkonen on the above when you have a chance ^

jasonblais:
After discussing with @kkekkonen, I will close this and the related pull requests, but keep the branches open so work can be easily continued later on.

The work halted here as a result of our process failures. In particular,
 - there was no owner for these PRs, leading to slow response times from the Mattermost team
 - there was no ticket, nor no design, prior to submitting the PRs, leading to 
   - Mattermost team giving confusing feedback (too high-level)
   - uncertainty and inconsistency about the direction of the feature
   - conflicting reviewer feedback, in some cases leading to contributor reverting his changes back

We've taken notes and started discussions on how to address the above - we'll start by having a JIRA ticket for every contributor pull request that details the expected behaviour and design, and have an owner for each.

PR #585, opened 2018-08-01
[MM-11051]: Add new param to displayUsername to circumvent defaultName

sudheerDev:
#### Summary
Adding a new param(useDefaultUserName) to avoid someone string from returning for cases where we need to show loader.

#### Ticket Link
[MM-11051](https://mattermost.atlassian.net/browse/MM-11051)

#### Checklist

### DIFF
--- a/src/utils/user_utils.js
+++ b/src/utils/user_utils.js
@@ -21,9 +21,9 @@ export function getFullName(user: UserProfile): string {
 export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: string,
-    usernameWithPrefix: boolean,
+    useDefaultUserName: boolean = true,
 ): string {
-    let name = localizeMessage('channel_loader.someone', 'Someone');
+    let name = useDefaultUserName ? localizeMessage('channel_loader.someone', 'Someone') : '';
 
     if (user) {
         if (teammateNameDisplay === Preferences.DISPLAY_PREFER_NICKNAME) {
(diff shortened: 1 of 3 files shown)


src/utils/user_utils.js
```diff
@@ -21,21 +21,21 @@ export function getFullName(user: UserProfile): string {
 export function displayUsername(
     user: UserProfile,
     teammateNameDisplay: string,
-    usernameWithPrefix: boolean,
+    useDefaultUserName: boolean = true,
```
enahum: this name is misleading, cause I'm seeing this returning `Someone` instead of the username, also why are we removing the `usernameWithPrefix` ?

src/utils/user_utils.js
```diff
@@ -21,21 +21,21 @@ export function getFullName(user: UserProfile): string {
...
         } else if (teammateNameDisplay === Preferences.DISPLAY_PREFER_FULL_NAME) {
             name = getFullName(user);
         } else {
-            name = usernameWithPrefix ? `@${user.username}` : user.username;
+            name = user.username;
         }
 
         if (!name || name.trim().length === 0) {
-            name = usernameWithPrefix ? `@${user.username}` : user.username;
+            name = user.username;
```
enahum: what in the entire product is affected by this change?

sudheerDev: There is no function that uses this param in the codebase anymore.

I could have used another param but using two flag params as arguments didn't feel right at all because I would have to change every place this function is used(That is a lot!) and pass the third params value even if it is default value to keep the fourth param optional.

The other option is to use another function.

src/utils/user_utils.js
sudheerDev: i  can change this to `fallbackUsername` if that makes sense to you. If you have anything in mind please do let me know

enahum: that works for me, or something like `allowAnonymous`

src/utils/user_utils.js
enahum: cool, just wanted to make sure, if that is the case we can leave the change

enahum:
cool, merge when ready and build has passed

PR #588, opened 2018-08-02
MM-11484 Add check for undefined channel objects

sudheerDev:
#### Summary
Add check for undefined channel objects

#### Ticket Link
[MM-11484](https://mattermost.atlassian.net/browse/MM-11484)

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Ran `make check-style` to check for style errors (required for all pull requests)
- [x] Ran `make test` to ensure unit tests passed
- [x] Ran `make flow` to ensure type checking passed

PR #590, opened 2018-08-03
MM-11556: introduce a getDirectAndGroupChannels

lieut-data:
#### Summary
This is a simplified variant of `getDirectChannels` that does does not exclude manually or automatically closed direct and group channels, nor archived channels. The downstream webapp needs the whole set to display the channels in the "Other" section of the channel quick switcher.

Modifying `getDirectChannels` directly was out of scope:
* it is being used by the mobile app, and thus risks changing semantics unexpectedly
* it is extremely difficult to unravel the business logic. This is a candidate for significant refactoring.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-11556

#### Checklist

#### Test Information

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -653,6 +653,31 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
     }
 );
 
+// getDirectAndGroupChannels returns all direct and group channels, even if they have been manually
+// or automatically closed.
+//
+// This is similar to the getDirectChannels above (which actually also returns group channels,
+// but suppresses manually closed group channels but not manually closed direct channels.) This
+// method does away with all the suppression, since the webapp client downstream uses this for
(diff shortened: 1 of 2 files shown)


src/selectors/entities/channels.js
```diff
@@ -653,6 +653,31 @@ export const getSortedDirectChannelWithUnreadsIds = createIdsSelector(
...
+    getCurrentUser,
+    getUsers,
+    getAllChannels,
+    getTeammateNameDisplaySetting,
+    (currentUser, profiles, channels, settings) => {
+        if (!currentUser) {
+            return [];
+        }
+
+        return Object.values(channels).filter((channel) =>
```
sudheerDev: @lieut-data Can you add a minor check for channelObj similar to https://github.com/mattermost/mattermost-redux/pull/588. we still are not sure why undefined channel objects get into the state this is just as a precaution

PR #592, opened 2018-08-03
Jenkins build improvements

lieut-data:
#### Summary
Update Jenkinsfile to build inside `node:8` docker container, add support for generating and consuming junit output from Mocha to show up in Jenkins pipeline, and clean up the stage/steps.

This is part of a learning experiment to figure out how to containerize the server builds, but I'm starting with an easy repository. By specifying the docker container here, we can also choose to upgrade to building on a different version of node independently of the underlying Jenkins slave AMI. For now I'm using the `node:8` image which of course will change with minor versions, but we could choose to pin this or even build our own.

When tests pass:

image

When a test fails:

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-11539

### DIFF
--- a/build/.gitignore
+++ b/build/.gitignore
@@ -1,0 +1,1 @@
+test-results.xml
--- a/build/Jenkinsfile
+++ b/build/Jenkinsfile
@@ -36,4 +36,9 @@ pipeline {
             }
         }
     }
+    post {
+        always {
+            junit 'build/test-results.xml'
+        }
+    }
 }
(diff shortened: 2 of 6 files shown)


lieut-data:
This does have the disadvantage of having to `docker pull node:8` on slave machines that have been recycled. We could choose to build a smaller image than the whole of Debian Jesse (alpine doesn't have enough to git checkout mattermost-redux, so this was the easy path.)

cpanato:
this is pretty cool!

PR #595, opened 2018-08-07
MM-11477 Add message to all doFetchWithResponse errors

hmhealey:
Sentry has better handling for non-error exceptions with a `message` field, so make sure to provide one in all cases. Also removes an unused method that didn't need to be part of Client4 since I think it was replaced by some mobile-specific code.

#### Ticket Link


hmhealey:
@enahum Yeah, it should go into 1.11

enahum:
Ok I’ll make sure it does

PR #597, opened 2018-08-10
[MM-11170] Fix unreads for deleted users

sudheerDev:
#### Summary
Fixes unread for deleted users

#### Ticket Link
[MM-11170](https://mattermost.atlassian.net/browse/MM-11170)

#### Checklist

#### Test Information

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -23,7 +23,7 @@ import {
 import {getLastPostPerChannel, getAllPosts} from 'selectors/entities/posts';
 import {getCurrentTeamId, getCurrentTeamMembership} from 'selectors/entities/teams';
 import {haveICurrentChannelPermission} from 'selectors/entities/roles';
-import {isCurrentUserSystemAdmin} from 'selectors/entities/users';
+import {isCurrentUserSystemAdmin, getCurrentUserId} from 'selectors/entities/users';
 
 import {
     buildDisplayableChannelList,
@@ -306,20 +306,33 @@ export const getMembersInCurrentChannel = createSelector(
 export const getUnreads = createSelector(
     getAllChannels,
     getMyChannelMemberships,
-    (channels, myMembers) => {
+    getUsers,
+    getCurrentUserId,
+    (channels, myMembers, users, currentUserId) => {
         let messageCount = 0;
         let mentionCount = 0;
         Object.keys(myMembers).forEach((channelId) => {
             const channel = channels[channelId];
             const m = myMembers[channelId];
+
             if (channel && m) {
+                let otherUserId;
                 if (channel.type === 'D') {
-                    mentionCount += channel.total_msg_count - m.msg_count;
+                    otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
+                    if (users[otherUserId]) {
+                        mentionCount += channel.total_msg_count - m.msg_count;
+                    }
                 } else if (m.mention_count > 0) {
                     mentionCount += m.mention_count;
                 }
                 if (m.notify_props && m.notify_props.mark_unread !== 'mention' && channel.total_msg_count - m.msg_count > 0) {
-                    messageCount += 1;
+                    if (channel.type === 'D') {
+                        if (users[otherUserId]) {
+                            messageCount += 1;
+                        }
+                    } else {
+                        messageCount += 1;
+                    }
                 }
             }
         });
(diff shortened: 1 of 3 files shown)


src/selectors/entities/channels.js
```diff
@@ -306,20 +306,33 @@ export const getMembersInCurrentChannel = createSelector(
...
         Object.keys(myMembers).forEach((channelId) => {
             const channel = channels[channelId];
             const m = myMembers[channelId];
+
             if (channel && m) {
+                let otherUserId;
                 if (channel.type === 'D') {
-                    mentionCount += channel.total_msg_count - m.msg_count;
+                    otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
+                    if (users[otherUserId]) {
```
enahum: don't you need to also check if the other user is not deleted? you may have it in the store cause it was previously loaded or something...

src/selectors/entities/channels.js
```diff
@@ -306,20 +306,33 @@ export const getMembersInCurrentChannel = createSelector(
...
+                        mentionCount += channel.total_msg_count - m.msg_count;
+                    }
                 } else if (m.mention_count > 0) {
                     mentionCount += m.mention_count;
                 }
                 if (m.notify_props && m.notify_props.mark_unread !== 'mention' && channel.total_msg_count - m.msg_count > 0) {
-                    messageCount += 1;
+                    if (channel.type === 'D') {
+                        if (users[otherUserId]) {
```
enahum: same here

src/selectors/entities/channels.js
sudheerDev: >don't you need to also check if the other user is not deleted

I didn't get you which `other user` you are referring to? . DM is between two users so, I am checking for the user other than the currentUser here.

Reagrding store:
Web: Deleted user will be part of redux store till a refreash. User will continue to see unread undicator and deleted user in LHS so, they can visit the DM and it clear automatically or they can refreash in this case deleted user will no longer be available in LHS and in store along with decreased unread count as it is supposed to be.

Mobile: Same case with mobile but here it is till clear cache.

enahum: Sorry I needed to make this clearer, other user means the second user in the DM that is not the current user.

Just for the fact that you need to refresh or clear the cache means that the user might still be in the redux state and the unread counts will include messages from this user, so you need to check that the user is in the store and deleted_at equals to 0

sudheerDev: > other user means the second user in the DM that is not the current user.

Yes, i am referring to the other user(Not the current one). We only look for other user data in store.

> Just for the fact that you need to refresh or clear the cache means that the user might still be in the redux state and the unread counts will include messages.

Yes, Please read the above comment once again. Deleted user is not removed from the store immediately. So till a refreash or clear cache you will see the user in LHS. But i dont see how this messes up with unread Logic.

If user is present in redux store or in cache you will see them in LHS and unread count appropriately i.e (Unread count)  so i dont see a descripanacy because it adds up to the count.

After refreash or clear cache unread count still adds up because `otherUser` is not part of LHS or in cache so - (unread count - deleted users count)

Regarding `deleted_at` i think you meant to say check for !== 0 but this probably is a bug in CLI as deleted_at is always zero for DM channels objects.

src/selectors/entities/channels.js
enahum: ok I can see what you mean..

2 things:
1. needs rebase
2. should we use the release-5.2 branch for this?

src/selectors/entities/channels.js
enahum: ok I can see what you mean..

2 things:
1. needs rebase
2. should we use the release-5.2 branch for this?

sudheerDev: I don't know if we should consider it for 5.2 as this is not a regression and it touches unread logic.

If this should be considered then i can rebase and raise it against 5.2

saturninoabril: > Regarding deleted_at i think you meant to say check for !== 0 but this probably is a bug in CLI as deleted_at is always zero for DM channels objects.

I believed it's not to DM channel but to the other user, like:
```javascript
if (users[otherUserId] && users[otherUserId].delete_at > 0)
```

sudheerDev: `but to the other user, like:` are you saying this needs to be added for channel?

I noticed one bug when testing this. For a just archived channel with unreads the count is not in sync with LHS. I will fix this and push a commit.

saturninoabril: Oh no, sorry.  I was thinking something like: 
```javascript
if (users[otherUserId] && users[otherUserId].delete_at === 0)
```

enahum:
@sudheerDev this needs a rebase

enahum:
and I think it should point to the release-5.2 branch?

src/selectors/entities/channels.js
```diff
@@ -306,20 +306,33 @@ export const getMembersInCurrentChannel = createSelector(
...
             const channel = channels[channelId];
             const m = myMembers[channelId];
+
             if (channel && m) {
+                let otherUserId;
                 if (channel.type === 'D') {
-                    mentionCount += channel.total_msg_count - m.msg_count;
-                } else if (m.mention_count > 0) {
+                    otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
+                    if (users[otherUserId] && users[otherUserId].delete_at === 0) {
```
mkraft: you could instantiate a variable for `users[otherUserId]` instead of accessing the array twice

sudheerDev:
Closing this as it is already merged

PR #602, opened 2018-08-14
New sidebar feature fixed

csduarte:
#### Summary
Bringing back the sidebar feature after a regression with the unread channels indicator on the web side.
Old PR: https://github.com/mattermost/mattermost-webapp/pull/1374

Nothing has changed in the redux library.

#### Checklist
- [x] Ran `make test` to ensure unit and component tests passed
- [x] Has redux changes (please link)
- [x] Has UI changes
- [x] Includes text changes and localization file ([.../i18n/en.json](https://github.com/mattermost/mattermost-webapp/blob/master/i18n/en.json)) updates


src/selectors/entities/channels.js
```diff
@@ -754,3 +823,172 @@ export const filterPostIds = (condition) => {
...
+export const getAllSortedChannelIds = createIdsSelector(
+    getUnreadChannelIds,
+    getFavoritesPreferences,
+    getAllChannelIds,
+    (state, lastUnreadChannel, unreadsAtTop = true) => unreadsAtTop,
+    (state, lastUnreadChannel, unreadsAtTop, favoritesAtTop = true) => favoritesAtTop,
+    filterChannels,
+);
+
+export const getOrderedChannelIds = (state, lastUnreadChannel, grouping, sorting, unreadsAtTop, favoritesAtTop) => {
```
hmhealey: This needs to be memoized since the Sidebar component in the webapp and the ChannelsList in the mobile app are calling this in mapStateToProps, but that's understandably hard because of how much goes into this function to make the output possibly changed.

While `createSelector` won't work because this requires the whole state be passed in, you could attempt to memoize it after the sections have been created since `getSortedPublicChannelIds` and such appear to be memoized correctly. Something like this would probably work to return the same channel arrays even if something unrelated in the store has changed.
```javascript
export const getOrderedChannelIds = () => {
    let lastChannels;

    const hasChannelsChanged = (channels) => {
        if (!lastChannels || lastChannels.length !== channels.length) {
            return true;
        }

        for (let i = 0; i < channels.length; i++) {
            if (channels[i].type !== lastChannels[i].type || channels[i].items !== lastChannels[i].items) {
                return true;
            }
        }

        return false;
    };

    return (state, lastUnreadChannel, grouping, sorting, unreadsAtTop, favoritesAtTop) => {
        const channels = ...; // Do everything like it's currently done to construct channels

        if (hasChannelsChanged(channels)) {
            lastChannels = channels;
        }

        return lastChannels;
    };
}();
```

src/selectors/entities/channels.js
```diff
@@ -754,3 +823,172 @@ export const filterPostIds = (condition) => {
...
+    const channels = [];
+
+    if (grouping === 'by_type') {
+        channels.push({
+            type: 'public',
+            name: 'PUBLIC CHANNELS',
```
hmhealey: It doesn't look like this `name` field is being used anywhere, so it can probably be removed. If it is being used, can you make sure that wherever is using it is localizing it properly?

src/selectors/entities/sidebar.js
```diff
@@ -0,0 +1,48 @@
...
+        );
+    },
+    (state) => {
+        const sidebarPreference = getPreference(
+            state,
+            Preferences.CATEGORY_SIDEBAR_SETTINGS,
+            '',
+            null
+        );
+        return JSON.parse(sidebarPreference);
```
hmhealey: `JSON.parse` will return a new object each time, so it should be moved into the result function to prevent this from being memoized properly

src/selectors/entities/sidebar.js
```diff
@@ -0,0 +1,48 @@
...
+import {getBool as getBoolPreference, get as getPreference} from 'selectors/entities/preferences';
+
+const defaultSidebarPrefs = {
+    grouping: 'by_type',
+    unreads_at_top: 'true',
+    favorite_at_top: 'true',
+    sorting: 'alpha',
+};
+
+export const getSidebarPreferences = createSelector(
```
hmhealey: Could you move this into `selectors/entities/preferences` since `sidebar` isn't an entity?

miguelespinoza:
Thanks for the feedback @hmhealey. I'm hoping to take a look at it today and tomorrow

miguelespinoza:
Just a heads up, we had to put this on hold.  Currently we have some top priority tasks for the upcoming weeks.
I'm hoping I can squeeze sometime next week to address this feedback but that's not certain

hmhealey:
Thanks for the heads up. Definitely understand since we're doing the same for some other features.

src/selectors/entities/channels.js
miguelespinoza: It's used in the webapp and it's being localized

src/selectors/entities/channels.js
```diff
@@ -463,29 +495,21 @@ export const getUnreadChannelIds = createIdsSelector(
...
+    (currentUser, profiles, channels, myMembers, unreadIds, settings) => {
         // If we receive an unread for a channel and then a mention the channel
         // won't be sorted correctly until we receive a message in another channel
         if (!currentUser) {
             return [];
         }
 
-        const locale = currentUser.locale || General.DEFAULT_LOCALE;
-        const allUnreadChannels = unreadIds.map((id) => {
+        const allUnreadChannels = unreadIds.filter((u) => u).map((id) => {
```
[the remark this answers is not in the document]
miguelespinoza: Last I check I was getting falsy ids.  Some where null.  I could check if that's still a problem

miguelespinoza:
Hey @hmhealey, just a heads up.  We plan on getting this in shape for merging.  I'm addressing your feedback and would greatly appreciate any input for getting this in master soon.  Thanks

src/selectors/entities/channels.js
```diff
@@ -198,7 +230,7 @@ export const getChannelsNameMapInCurrentTeam = createSelector(
 );
 
 // Returns both DMs and GMs
-export const getDirectChannels = createSelector(
+export const getAllDirectChannels = createSelector(
```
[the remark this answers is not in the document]
miguelespinoza: Hey @sudheerDev we have a mobile PR for sidebar preference support. We should ensure that it's good to go there

src/selectors/entities/channels.js
```diff
@@ -525,30 +550,34 @@ export const getSortedUnreadChannelIds = createIdsSelector(
...
+);
+
+export const getSortedUnreadChannelIds = createIdsSelector(
+    getUnreadChannelIds,
+    getFavoritesPreferences,
+    (state, lastUnreadChannel, unreadsAtTop, favoritesAtTop, sorting = 'alpha') => {
+        return getMapAndSortedUnreadChannelIds(state, lastUnreadChannel, sorting);
+    },
+    (unreadChannelIds, favoritePreferences, mappedAndSortedUnreadChannelIds) => {
+        return filterChannels(unreadChannelIds, favoritePreferences, mappedAndSortedUnreadChannelIds, false, false);
```
hmhealey: Since we're always passing false into these last two, can we skip that call entirely?

src/selectors/entities/channels.js
```diff
@@ -491,28 +524,20 @@ export const getUnreadChannelIds = createIdsSelector(
...
+export const getUnreadChannels = createIdsSelector(
     getCurrentUser,
     getUsers,
     getAllChannels,
     getMyChannelMemberships,
     getUnreadChannelIds,
     getTeammateNameDisplaySetting,
-    (state, lastUnreadChannel = null) => lastUnreadChannel,
-    (currentUser, profiles, channels, myMembers, unreadIds, settings, lastUnreadChannel) => {
+    (currentUser, profiles, channels, myMembers, unreadIds, settings) => {
```
hmhealey: I don't think `myMembers` is needed here since the filtering has been removed

src/selectors/entities/channels.js
```diff
@@ -525,30 +550,34 @@ export const getSortedUnreadChannelIds = createIdsSelector(
             }
 
             return c;
-        }).filter((c) => c).sort((a, b) => {
```
sudheerDev: Filter here is used for removing deleted channels re: line 544:546 

```
if (c.delete_at !== 0) {
     return false;
}
```

hmhealey: Really good catch. I didn't see that before. We should actually make it so that a filter is done to remove deleted channels before the map call instead of returning false sometimes. Having mixed return types from the map statement is not something I expected at all, and it definitely causes problems like the one you caught

sudheerDev: I am not a fan of this logic here. I will let miguelespinoza decide if he wants to fix this or leave it as two maps. 
Having a reduce would have done the job rather than mixed return types and two `maps`

I recently almost broke this https://github.com/mattermost/mattermost-redux/pull/680.
and thought about refactoring it myself so it obvious on first look so we don't run into issues like this but coudn't make the time.

hmhealey: A reduce would do it as well, but this only needs one map
```javascript
const allUnreadChannels = unreadIds.filter((id) => channels[id] && channels[id].delete_at > 0).map(...);
```

sudheerDev: 👍 Aah ya that is more readable. I tend to resort to reduce whenever i need to filter and iterate.

miguelespinoza: Nice, thanks for the feedback @hmhealey and nice catch @sudheerDev.  I'll address this hopefully today.  If not for sure tomorrow. I'm glad you found that catch @sudheerDev 👍

miguelespinoza:
@hmhealey @sudheerDev thanks for the feedback.  I've addressed this and will rebase the webapp PR tomorrow.  Everything is good to go for the redux PR

hmhealey:
Looks good. We're going to hold off on merging until the web app and mobile PRs are ready to avoid any compatibility issues

PR #612, opened 2018-08-23
add getChannelTimezone to redux

cpanato:
#### Summary
Add the getChannelTimezone to redux 

#### Ticket Link
N/a

#### Other Prs:
webapp: https://github.com/mattermost/mattermost-webapp/pull/1607
server: https://github.com/mattermost/mattermost-server/pull/9286
api reference: https://github.com/mattermost/mattermost-api-reference/pull/387

#### Checklist

#### Test Information
This PR was tested on: mac ox 10.13.6


jasonblais:
@lieut-data @jwilander Here is the associated Redux PR for https://github.com/mattermost/mattermost-server/pull/9286

src/actions/channels.js
```diff
@@ -467,6 +467,35 @@ export function getChannelAndMyMember(channelId) {
     };
 }
 
+export function getChannelTimezones(channelId) {
+    return async (dispatch, getState) => {
+        dispatch({type: ChannelTypes.GET_CHANNELS_TIMEZONE_REQUEST}, getState);
```
hmhealey: `dispatch` doesn't actually take a second argument, so we don't need to pass `getState` into it

src/actions/channels.js
```diff
@@ -467,6 +467,35 @@ export function getChannelAndMyMember(channelId) {
...
+        let channelTimezones;
+        try {
+            const channelTimezonesRequest = Client4.getChannelTimezones(channelId);
+
+            channelTimezones = await channelTimezonesRequest;
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: ChannelTypes.GET_CHANNELS_TIMEZONE_FAILURE, error},
+                logError(error)(dispatch),
```
hmhealey: Similarly, we don't need to explicitly pass `dispatch` into this since it'll be handled by the outer `dispatch`

src/actions/channels.js
```diff
@@ -467,6 +467,35 @@ export function getChannelAndMyMember(channelId) {
...
+        }
+
+        dispatch(batchActions([
+            {
+                type: ChannelTypes.GET_CHANNELS_TIMEZONE_SUCCESS,
+                data: channelTimezones,
+            },
+        ]), getState);
+
+        return {data: channelTimezones};
```
hmhealey: I'm not sure how I feel about a Redux action that doesn't persist it's results to the store, but at the same time, I don't think it necessarily makes sense to add this data to the store since it's only used for this one-off action.

I think I'm good with how this is done for now since we can always move it into the store if we need to later

cpanato:
@hmhealey PTAL

src/actions/channels.js
cpanato: done

src/actions/channels.js
cpanato: done

PR #615, opened 2018-08-27
Feature/search after before on

dmitrysamuylovpharo:
#### Summary
Modified Search Posts api to accept an object with parameters to support passing of additional parameters and specifically of time zone offset value for date based flag searches to allow the server to perform those in the context of the client's time zone

#### Ticket Link
N/A

#### Checklist
N/A

#### Test Information
Mac OSX desktop client

### DIFF
--- a/src/actions/search.js
+++ b/src/actions/search.js
@@ -64,7 +64,8 @@ export function searchPostsWithParams(teamId, params, includeDeletedChannels = f
                 type: SearchTypes.RECEIVED_SEARCH_TERM,
                 data: {
                     teamId,
-                    params,
+                    terms: params.terms,
+                    isOrSearch: params.is_or_search,
                 },
             },
             {
--- a/src/actions/search.js
+++ b/src/actions/search.js
@@ -34,13 +34,13 @@ function getMissingChannelsFromPosts(posts) {
     };
 }
 
-export function searchPosts(teamId, terms, isOrSearch = false, includeDeletedChannels = false) {
+export function searchPostsWithParams(teamId, params, includeDeletedChannels = false) {
     return async (dispatch, getState) => {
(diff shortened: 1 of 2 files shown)


src/client/client4.js
```diff
@@ -1476,15 +1476,19 @@ export default class Client4 {
...
 
         return this.doFetch(
             `${this.getTeamRoute(teamId)}/posts/search?include_deleted_channels=${includeDeletedChannels}`,
-            {method: 'post', body: JSON.stringify({terms, is_or_search: isOrSearch})}
+            {method: 'post', body: JSON.stringify(params)}
         );
     };
 
+    searchPosts = async (teamId, terms, isOrSearch, includeDeletedChannels = false) => {
+        this.searchPostsWithParams(teamId, {terms, is_or_search: isOrSearch}, includeDeletedChannels);
```
hmhealey: This should be `return this.searchPostsWithParams(...);` in case anyone wants to wait on this request

src/actions/search.js
```diff
@@ -64,8 +64,7 @@ export function searchPosts(teamId, terms, isOrSearch = false, includeDeletedCha
                 type: SearchTypes.RECEIVED_SEARCH_TERM,
                 data: {
                     teamId,
-                    terms,
-                    isOrSearch,
+                    params,
```
hmhealey: The reducer that handles this action still expects to the `terms` and `isOrSearch` fields, so you should either update the reducer to use `params` or just pass them with their previous names like
```javascript
data: {
    teamId,
    terms: params.terms,
    isOrSearch: params.is_or_search,
}
```

hmhealey:
As for the tests failing, I think it's due to the missing `return` statement noted above

dmitrysamuylovpharo:
@enahum it defaults to UTC on the server if you don't supply it here

PR #617, opened 2018-08-28
MM-11696 Factor out marking GMs open to a separate action

jwilander:
#### Summary
Factor out marking GMs open to a separate action so it can be called by the web app when needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-11696

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -159,6 +159,19 @@ export function createDirectChannel(userId, otherUserId) {
     };
 }
 
+export function markGroupChannelOpen(channelId) {
+    return async (dispatch, getState) => {
+        const {currentUserId} = getState().entities.users;
+
+        const preferences = [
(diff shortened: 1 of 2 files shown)


src/actions/channels.js
```diff
@@ -209,10 +217,6 @@ export function createGroupChannel(userIds) {
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: member,
             },
-            {
```
lieut-data: I'm guessing `savePreferences` already effected this action?

enahum: That action is being received by the websocket if i’m not mistaken

jwilander: Yes, it already effects this through the redux offline dispatch

PR #618, opened 2018-08-29
[MM-11717] Exclude deleted users from favourites list

sudheerDev:
#### Summary
Exclude deleted users from the favourites list

#### Ticket Link
[MM-11717](https://mattermost.atlassian.net/browse/MM-11717)

#### Checklist

#### Test Information
Mac os high sierra

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -549,6 +549,11 @@ export const getSortedFavoriteChannelWithUnreadsIds = createIdsSelector(
                 return false;
             }
 
+            // Deleted users from CLI will not have a profiles entry
+            if (channel.type === General.DM_CHANNEL && !profiles[otherUserId]) {
+                return false;
+            }
+
             if (channel.type === General.DM_CHANNEL && !isDirectChannelVisible(profiles[otherUserId] || otherUserId, config, prefs, channel, null, null, currentChannelId)) {
                 return false;
             } else if (channel.type === General.GM_CHANNEL && !isGroupChannelVisible(config, prefs, channel)) {
--- a/test/selectors/channels.test.js
+++ b/test/selectors/channels.test.js
@@ -744,5 +744,57 @@ describe('Selectors.Channels', () => {
                 {...channel13, display_name: profiles.fakeUserId.username},
             ]);
         });
+
+        it('Should not include deleted users in favorites', () => {
+            const newDmChannel = TestHelper.fakeDmChannel(user.id, 'newfakeId');
+            newDmChannel.total_msg_count = 1;
+            newDmChannel.display_name = '';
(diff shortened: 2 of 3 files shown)


jespino:
I removed @mkraft as reviewer because is on holidays all the week :)

src/selectors/entities/channels.js
```diff
@@ -549,6 +549,11 @@ export const getSortedFavoriteChannelWithUnreadsIds = createIdsSelector(
                 return false;
             }
 
+            // Deleted users from CLI will not have a profiles entry
```
lieut-data: Are there other reasons why a user might not have a profiles entry? (e.g. async loading still pending or some such).

Is it worth updating the CLI tool to "unmark" such favourites vs. requiring the client-side to handle this?

sudheerDev: Yes, it can happen but for those scenarios we request profile and we hide in the sidebar till we get the info of the user profile but as this selector returns the ID and we hide in the sidebar the title persists. 

> Is it worth updating the CLI tool to "unmark" such favourites vs. requiring the client-side to handle this?

Definetely but not just for deleted users i would say we need to this for all CLI commands i.e delete, deactivate and archive etc. 

One we update the CLI command we wont see these issues anymore so we dont have to hide or decreases counts manually on clients 

Any idea on how we can handle these for existing deleted users? because i would eventually like to remove all this if we correct CLI commands

jespino: For existing deleted users we can search all the favorites for all the users and remove that which doesn't match with any user. And for a normal user remove process. we can search for all the favorites of all users to remove that ones which are direct messages with the user.

I think that would be very expensive en performance, and I think the solution of solving it from the front-end is not a huge deal.

sudheerDev: I am sorry i am like a newbie when it comes to server stuff. 
> For existing deleted users we can search all the favorites for all the users and remove that which doesn't match with any user.

How would this work? would be like a cleanup job at the start of update? or should it be a script sys admin has to run?

I think we need to have this till we fix our cli commands anyway though

jespino: In that case, normally is executed as a migration, it means, executes automatically during the upgrade. But I'm still not sure we want that. Is something expensive in performance I think, we can think about it, but for me is out of the scope of this fix/pr.

@lieut-data What do you think about it?

lieut-data: I think this PR is fine for the scope as currently defined.

In general, I'm pushing towards "getting the data model right" vs. trying to encode deep knowledge like this at the client-level, unless there's a use case where the information is still needed. I'd love to see a migration clean up the existing data in the same release that we fix the CLI when deleting users, but 0/5 on timing right now.

cpanato:
@sudheerDev please change to `release-5.3`

PR #626, opened 2018-09-06
Fix when opengraph returns null

enahum:
#### Summary
Action was failing when doing a request to get the opengraph data returned `null`


hmhealey:
When can the server return null for this? As far as I'm aware, that endpoint always returns at least an empty object

enahum:
@hmhealey good question I do not know when, what I do know is that when making a request for a permalink I end up having an error saying `TypeError: null is not an object (evaluating 'data.url')` so to me that means **data** is either null or undefined

PR #635, opened 2018-09-18
Run babel on 'npm install' and remove .npminstall

hmhealey:
I've always found the fact that you can't just `npm install mattermost/mattermost-redux` kind of annoying, but moving the build step to `prepublish` or `prepare` (both are very similar, but `prepare` happens after `prepublish`) instead of `prepublishOnly` so that it triggers automatically on `npm install` should fix that.

This also lets us remove the `.npminstall` files on each repo since we can just check if `node_modules` is up to date rather than creating a new file.

Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/1717
Mobile PR: https://github.com/mattermost/mattermost-mobile/pull/2135

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: check-style clean pre-run test install flow
 
-.npminstall: package.json
+node_modules: package.json
 	@if ! [ $(shell which npm) ]; then \
 		echo "npm is not installed"; \
 		exit 1; \
@@ -10,9 +10,7 @@
 
 	npm install --ignore-scripts
 
-	touch $@
-
-check-style: | pre-run .npminstall
+check-style: | pre-run node_modules
 	@echo Checking for style guide compliance
 
 	npm run check
@@ -21,7 +19,6 @@ clean:
 	@echo Cleaning app
 
 	rm -rf node_modules
-	rm -f .npminstall
 
 pre-run:
 	@echo Make sure no previous build are in the folder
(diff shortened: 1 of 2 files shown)


enahum:
Will this work with babel 7?

hmhealey:
@enahum What's changing with babel 7? Regardless, I don't think that should affect this since I'm just making `npm build` automatic instead of having it done manually in the webapp/mobile Makefiles

enahum:
a lot of stuff changing in babel 7, RN 57 actually uses babel 7 and the only way I was able to make it work was actually running `npm i` inside the mattermost-redux module before building it, so we should make sure it does build or find a way to have the module_resolver babel plugin actually let us use the lib without building it.

Another approach could be publishing a new version to npm on each merge to master or something like that.

enahum:
you can try the `rn57` branch of the mobile app and see

hmhealey:
Looks good to me. I took your branch, cherry-picked the mobile changes for this PR onto it, removed [this line](https://github.com/mattermost/mattermost-mobile/commit/f27dc95cf0b2a6fc65f382be110273321c9bb962#diff-b67911656ef5d18c4ae36cb6741b7965L85), and then ran `make test` successfully. My branch for it is [here](https://github.com/mattermost/mattermost-mobile/commits/hh-prepare-rn57)

enahum:
@hmhealey that sounds great, what about actually running the app ?

enahum:
actually just tested it, is working great

PR #636, opened 2018-09-19
[MM-9490] Add a selector to get all user IDs from the reactions of a post

saturninoabril:
#### Summary
Add a selector to get all user IDs from the reactions of a post.

Related PR on mobile https://github.com/mattermost/mattermost-mobile/pull/2125/files

#### Ticket Link
Jira ticket: [MM-9490](https://mattermost.atlassian.net/browse/MM-9490)

#### Checklist

#### Test Information
This PR was tested on: [Webapp, iOS simulator, Android emulator]

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -32,9 +32,20 @@ export function makeGetReactionsForPost() {
         getReactionsForPosts,
         (state, postId) => postId,
         (reactions, postId) => {
-            return Object.values(reactions[postId] || {});
+            return Object.values(reactions[postId] || []);
         }
     );
+}
+
+const getReactionsForPost = makeGetReactionsForPost();
+
+export function getUserIdsFromReactions(state, postId) {
+    const reactions = getReactionsForPost(state, postId);
+
+    return reactions.reduce((acc, reaction) => {
+        acc.push(reaction.user_id);
+        return acc;
+    }, []);
 }
 
 export function getOpenGraphMetadata(state) {
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -42,10 +42,7 @@ const getReactionsForPost = makeGetReactionsForPost();
 export function getUserIdsFromReactions(state, postId) {
     const reactions = getReactionsForPost(state, postId);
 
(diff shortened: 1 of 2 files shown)


src/selectors/entities/posts.js
```diff
@@ -32,11 +32,22 @@ export function makeGetReactionsForPost() {
         getReactionsForPosts,
         (state, postId) => postId,
         (reactions, postId) => {
-            return Object.values(reactions[postId] || {});
+            return Object.values(reactions[postId] || []);
```
saturninoabril: Corrected this since it returns an array of object/s.

src/selectors/entities/posts.js
```diff
@@ -32,11 +32,22 @@ export function makeGetReactionsForPost() {
...
         }
     );
 }
 
+const getReactionsForPost = makeGetReactionsForPost();
+
+export function getUserIdsFromReactions(state, postId) {
+    const reactions = getReactionsForPost(state, postId);
+
+    return reactions.reduce((acc, reaction) => {
```
ThiefMaster: Isn't that reduce() in the end doing the same as `reactions.map(reaction => reaction.user_id)`?

saturninoabril: Oh right, I'll update that.  Thanks!

src/selectors/entities/posts.js
hmhealey: If that's the case, do we even need the `Object.values` here?

src/selectors/entities/posts.js
```diff
@@ -32,11 +32,19 @@ export function makeGetReactionsForPost() {
...
         (reactions, postId) => {
-            return Object.values(reactions[postId] || {});
+            return Object.values(reactions[postId] || []);
         }
     );
 }
 
+const getReactionsForPost = makeGetReactionsForPost();
+
+export function getUserIdsFromReactions(state, postId) {
```
hmhealey: Since this is taking the postId as an argument, it should be done as a factory function that returns an instance of the selector

src/selectors/entities/posts.js
saturninoabril: Hmm, my bad, it's not correct.  I'll revert the change.

PR #637, opened 2018-09-20
MM-12254 Removing legacy client4 behavour.

crspeller:


hmhealey:
Is this the only way that `this.token` is set? If so, we still need that since it's used for websocket code and some other stuff on mobile

crspeller:
@hmhealey No it's not. The mobile app sets it itself as well. I have tested it and it works.

PR #641, opened 2018-09-21
MM-12205: manage mySession.expired

lieut-data:
#### Summary
Manage a new reducer tracking whether or not the current session has expired.

A session expiry event occurs whenever an HTTP request fails with 401 unauthorized and we force logout. An intentional logout can triggers this same code path, however, so the expiration bit is only toggled if a logout request hasn't already occurred.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12205

#### Checklist

#### Test Information
This PR was tested on: OSX Chrome

### DIFF
--- a/src/actions/helpers.js
+++ b/src/actions/helpers.js
@@ -17,6 +17,7 @@ export function forceLogoutIfNecessary(err: Client4Error, dispatch: DispatchFunc
     const {currentUserId} = getState().entities.users;
     if (err.status_code === HTTP_UNAUTHORIZED && err.url && err.url.indexOf('/login') === -1 && currentUserId) {
         Client4.setToken('');
+        dispatch({type: UserTypes.SESSION_EXPIRED});
         dispatch({type: UserTypes.LOGOUT_SUCCESS, data: {}});
     }
 }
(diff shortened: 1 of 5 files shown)


src/actions/helpers.js
```diff
@@ -17,6 +17,7 @@ export function forceLogoutIfNecessary(err: Client4Error, dispatch: DispatchFunc
     const {currentUserId} = getState().entities.users;
     if (err.status_code === HTTP_UNAUTHORIZED && err.url && err.url.indexOf('/login') === -1 && currentUserId) {
         Client4.setToken('');
+        dispatch({type: UserTypes.SESSION_EXPIRED});
```
enahum: actually can you combined this with a `batchActions` ?

enahum: nvm don't do that as it will break something in the mobile middleware, forget what I said

lieut-data:
I ended up pivoting away from tracking this in redux so as to be able to detect session expirations even if the page is refreshed (i.e. on desktop app restart only to discover the session expiration then vs. automatic login).

PR #642, opened 2018-09-23
[#MM-12130] Custom Terms of Service

chetanyakan:
#### Summary
- Create a new selector for showCustomTerms.
- Create an action for updateServiceTermsStatus.
- Updated Client4.js to make API call for updateServiceTermsStatus.
- Refactor buildQueryString from Client4.js to utils to be able to use it in the web-app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12130

#### Checklist
- [x] Ran `make check-style` to check for style errors
- [x] Ran `npm test` to ensure unit tests passed

#### Test Information
This PR was tested on: Ubuntu 18.04


src/actions/users.js
```diff
@@ -610,6 +610,22 @@ export function getMe() {
...
+        await bindClientFunc(
+            Client4.updateServiceTermsStatus,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_REQUEST,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_FAILURE,
+            userId,
+            serviceTermsId,
+            accepted
+        )(dispatch, getState);
+        await getMe()(dispatch, getState);
```
enahum: This should be `dispatch(getMe())` but why are we getting the user profile again agter the ToS was accepted or denied?

src/actions/users.js
```diff
@@ -610,6 +610,22 @@ export function getMe() {
     };
 }
 
+export function updateServiceTermsStatus(userId, serviceTermsId, accepted) {
```
enahum: Why passing an userId, shouldn’t this be for the current user?

src/actions/users.js
```diff
@@ -610,6 +610,22 @@ export function getMe() {
...
+    return async (dispatch, getState) => {
+        await bindClientFunc(
```
enahum: `dispatch(bindClientFunc(...))`

src/client/client4.js
```diff
@@ -416,6 +416,13 @@ export default class Client4 {
         );
     }
 
+    updateServiceTermsStatus = async (userId, serviceTermsId, accepted) => {
```
enahum: Shouldn’t this also be for the current user?

src/reducers/requests/users.js
```diff
@@ -318,6 +318,16 @@ function switchLogin(state: RequestStatusType = initialRequestState(), action: G
     );
 }
 
+function updateServiceTermsStatus(state: RequestStatusType = initialRequestState(), action: GenericAction): RequestStatusType {
+    return handleRequest(
+        UserTypes.UPDATE_SERVICE_TERMS_STATUS_REQUEST,
+        [UserTypes.RECEIVED_ME, UserTypes.UPDATE_ME_SUCCESS, UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS],
```
enahum: Why RECEIVED_ME?

src/actions/users.js
chetanyakan: We need to update the current user object in the store, so that the selector, which reads `!user.latest_service_terms_accepted` works as expected.

src/actions/users.js
chetanyakan: If we don't use `await` here, the get call for the user object is completed before the `latest_service_terms_accepted` can be updated. It leads to getting stale data.

src/actions/users.js
chetanyakan: Also, if we don't use `await` here, the user data is not updated before the update action in the webapp returns a response without the user object being updated. It leads to `showCustomTerms` selector to behave incorrectly. So, we need to `await` in both.

src/reducers/requests/users.js
chetanyakan: It was a mistake. Removed.

(Initially, we were planning to send the updated user as the API response. We had changed that behaviour and I should have removed this.)

src/client/client4.js
chetanyakan: Updated.

src/actions/users.js
chetanyakan: Updated

src/actions/users.js
```diff
@@ -610,6 +610,21 @@ export function getMe() {
     };
 }
 
+export function updateServiceTermsStatus(serviceTermsId, accepted) {
+    return async (dispatch, getState) => {
+        await bindClientFunc(
```
enahum: Ok so let me explain you can do

```js
await dispatch(bindClientFunc(
    Client4.updateServeTermsStatus,
    UserTypes.UPDATE_SERVICE_TERMS_STATUS_REQUEST,
    UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS,
    UserTypes.UPDATE_SERVICE_TERMS_STATUS_FAILURE,
    serviceTermsId,
    accepted));
```

src/actions/users.js
```diff
@@ -610,6 +610,21 @@ export function getMe() {
...
+            Client4.updateServiceTermsStatus,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_REQUEST,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_FAILURE,
+            serviceTermsId,
+            accepted
+        )(dispatch, getState);
+        await getMe()(dispatch, getState);
```
enahum: the same as above here, BUT, I'm not convinced you need to get the user profile yet again after logging in just to update 1 property, you could dispatch an action just to update that property in the current user, avoiding what is seems an unnecessary request to the server

enahum:
Also one other worry.

The custom terms of service seems to be part of the config that is sent to the client, meaning that every time the config is fetch we will get the text of the terms which can be very large increasing the payload, I think its going to be better if we have an endpoint to fetch the text for the custom terms and retrieve it only when we need to show them so the user can Agree.

thoughts?

jasonblais:
@jwilander You were proposed as a second reviewer for this PR

jwilander:
Took a quick look and it looks fine other than @enahum's comment. I'm going to wait until the server PR is approved before doing a proper review here as this PR is very dependent on the server PR.

lieut-data:
@enahum, we ended up converging on a discrete endpoint in the server PR around similar concerns. I believe @chetanyakan will update this PR to reflect these changes soon.

chetanyakan:
As @enahum suggested, I have added an action to make API call for `getServiceTerms`.

src/actions/users.js
```diff
@@ -610,6 +610,37 @@ export function getMe() {
...
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_FAILURE,
+            serviceTermsId,
+            accepted
+        ));
+        if (data && data.status === 'OK') {
+            const currentUser = getCurrentUser(getState());
+            dispatch({
+                type: UserTypes.RECEIVED_ME,
+                data: Object.assign({}, currentUser, {accepted_service_terms_id: JSON.parse(accepted) ? serviceTermsId : null}),
```
jwilander: `accepted` is just true or false, right? I don't think it needs a `JSON.parse()` around it

test/test_helper.js
```diff
@@ -79,6 +79,20 @@ class TestHelper {
...
+            accepted_service_terms_id: 1,
+        };
+    };
+
+    fakeUserWithServiceTermsNotAccepted = () => {
+        return {
+            ...this.fakeUserWithId(),
+            accepted_service_terms_id: 0,
+        };
+    };
```
jwilander: I don't think these really need to be functions in the test helper since they're so specific. I'd suggest just doing this in-line where needed or create helper functions right in the test file itself if you still want functions

src/actions/users.js
```diff
@@ -610,6 +610,37 @@ export function getMe() {
...
+    return async (dispatch, getState) => {
+        const {data, error} = await dispatch(bindClientFunc(
+            Client4.updateServiceTermsStatus,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_REQUEST,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_SUCCESS,
+            UserTypes.UPDATE_SERVICE_TERMS_STATUS_FAILURE,
+            serviceTermsId,
+            accepted
+        ));
+        if (data && data.status === 'OK') {
```
enahum: can we make 'OK' a constant? is there any other way to check for success? like a `status_code`?

jwilander:
Looks good other than the style error:

```
/home/ubuntu/jenkins/workspace/rdx_mattermost-redux_PR-642-4UXLIHBBDRCT446TNR7JQXBDEP4IHEGKUISDOYSFXTGBZDZVA2AA/src/selectors/entities/users.js

  351:20  error  'license' is defined but never used  no-unused-vars



```

PR #643, opened 2018-09-24
MM-12027 Adding paging for elasticsearch results.

crspeller:


src/actions/search.js
```diff
@@ -78,7 +83,19 @@ export function searchPostsWithParams(teamId, params, includeDeletedChannels = f
 }
 
 export function searchPosts(teamId, terms, isOrSearch, includeDeletedChannels) {
-    return searchPostsWithParams(teamId, {terms, is_or_search: isOrSearch}, includeDeletedChannels);
+    return searchPostsWithParams(teamId, {terms, is_or_search: isOrSearch, include_deleted_channels: includeDeletedChannels, page: 0, per_page: 20});
```
hmhealey: Is there a constant for this per page default?

src/actions/search.js
```diff
@@ -78,7 +83,19 @@ export function searchPostsWithParams(teamId, params, includeDeletedChannels = f
...
+}
+
+export function getMorePostsForSearch() {
+    return async (dispatch, getState) => {
+        const teamId = getCurrentTeamId(getState());
+        const {params, isEnd} = getState().entities.search.current[teamId];
+        if (!isEnd) {
+            const newParams = Object.assign({}, params);
```
hmhealey: It's more common for us to use the spread operator `{...params}` instead of Object.assign to copy objects lately

src/actions/search.js
```diff
@@ -78,7 +83,19 @@ export function searchPostsWithParams(teamId, params, includeDeletedChannels = f
...
+            newParams.page = newParams.page + 1;
+            await searchPostsWithParams(teamId, newParams)(dispatch, getState);
```
hmhealey: Should this be returning the results of searchPostsWithParams?

src/reducers/entities/search.js
```diff
@@ -171,4 +220,13 @@ export default combineReducers({
     // Object where every key is a team composed with
     // an object where the key is the term and the value indicates is "or" search
     recent,
+
+    // Object holding the current searches for every team
+    current,
+
+    // Boolean true if we are are searching initally
+    isSearchingTerm,
```
hmhealey: Do we need both isSearchingTerm and isSearchGettingMore since they're opposites of each other? 0/5 though

src/actions/search.js
crspeller: No, but I can add one.

src/actions/search.js
crspeller: Ya I guess it should.

src/reducers/entities/search.js
crspeller: They are not opposites. They are both false if nothing is happening.

src/actions/search.js
crspeller: Is that really better? It seems less clear what is happening. (not that Object.assign is particularly clear)

src/reducers/entities/search.js
hmhealey: Ah right. I missed that part

src/actions/search.js
hmhealey: It's not particularly better, but it would be more consistent. Also you could extend that to
```
const newParams = {
    ...params,
    page: params.page + 1,
};
```
to combine it with the next line if you wanted. 0/5 though

PR #647, opened 2018-09-25
Fix "{yyyy} {name of copyright owner}" in LICENSE.txt

jasonblais:


jasonblais:
@enahum It's 2016 in the License header.

Though that said, I'm actually not sure if it should be 2016 or "2016-present"

enahum:
@jasonblais it should be `Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.` as well as in the header

jasonblais:
Changed it to "2016-present" since this project was created in 2016,

enahum:
This is a file in this project https://github.com/mattermost/mattermost-redux/blob/master/src/client/client4.js#L1 as any other file in this project is 2015-present

jasonblais:
Sure, though I still think it should be 2016-present in all headers then,

PR #653, opened 2018-09-28
Fixing flow errors

jespino:
#### Summary
Fixing flow errors

#### Checklist


hmhealey:
Are these not being checked automatically? Shouldn't those errors have failed the build on Jenkins?

jespino:
Yes, I though that, is part of the Makefile test task requirements, but maybe jenkins is not using directly the Make tasks.

PR #654, opened 2018-09-28
Enabling flow check in the jenkins pipeline

jespino:
#### Summary
Enabling flow check in the jenkins pipeline

### DIFF
--- a/build/Jenkinsfile
+++ b/build/Jenkinsfile
@@ -20,6 +20,12 @@ pipeline {
             }
         }
 
+        stage('Check types') {
+            steps {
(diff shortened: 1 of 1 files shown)


jespino:
I don't know if this change is enough to enable the flow checking in Jenkins. Actually it doesn't work in the branch, I don't know if this need to be in master.

cpanato:
if you open this pr in upstream branch it will get that change, otherwise just when we merge it

jespino:
Ok, then I close it and create a new one with the upstream branch.

PR #659, opened 2018-10-01
MM-11044 Add flow annotations to preferences actions

anchepiece:
#### Summary
Flow type definitions added to preferences

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-11044
https://github.com/mattermost/mattermost-server/issues/9300

#### Checklist
- [X] Ran `make check-style` to check for style errors (required for all pull requests)
- [X] Ran `make test` to ensure unit tests passed
- [X] Ran `make flow` to ensure type checking passed

#### Test Information
This PR was tested on: Ubuntu 18.04.1 LTS

### DIFF
--- a/src/actions/preferences.js
+++ b/src/actions/preferences.js
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// @flow
 
 import {Client4} from 'client';
 import {Preferences} from 'constants';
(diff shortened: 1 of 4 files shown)


src/actions/preferences.js
```diff
@@ -12,8 +13,10 @@ import {bindClientFunc} from './helpers';
 import {getProfilesByIds, getProfilesInChannel} from './users';
 import {getChannelAndMyMember, getMyChannelMember} from './channels';
 
-export function deletePreferences(userId, preferences) {
-    return async (dispatch, getState) => {
+import type {GetStateFunc, DispatchFunc, ActionFunc} from '../types/actions';
+
+export function deletePreferences(userId: string, preferences: Object): ActionFunc {
```
jespino: In general the PR is ok, but you are using `Object` and we prefer to avoid that. If you want we can define the Preferences type. In this case, the preferences is an array of "Preference" type, which is
```
{|
  category: string
  name: string
  user_id: string
  value: string
|}
```

jespino:
Maybe, if you want, you can add that to the `src/types/store.js` file, to check the `myPreferences` field, which is a `{[string]: PreferenceType}` (I think).

PR #661, opened 2018-10-03
Remove duplicates search results

enahum:
#### Summary
With the new mechanism to get more search results we are actually retrieving results that we already have, so this PR will eliminate those duplicates from the reducer.

The best way to fix this is actually make sure that the server does not return duplicates, but this will definitely be an improvement if we ever get duplicate results. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12399

### DIFF
--- a/src/reducers/entities/search.js
+++ b/src/reducers/entities/search.js
@@ -9,7 +9,7 @@ function results(state = [], action) {
     switch (action.type) {
     case SearchTypes.RECEIVED_SEARCH_POSTS: {
         if (action.isGettingMore) {
-            return state.concat(action.data.order);
+            return [...new Set(state.concat(action.data.order))];
         }
         return action.data.order;
     }


src/reducers/entities/search.js
```diff
@@ -9,7 +9,7 @@ function results(state = [], action) {
     switch (action.type) {
     case SearchTypes.RECEIVED_SEARCH_POSTS: {
         if (action.isGettingMore) {
-            return state.concat(action.data.order);
+            return [...new Set(state.concat(action.data.order))];
```
crspeller: It's a bit strange to use a set here. The semantics of sets in general don't preserve order. I am assuming JS preserves order here.

enahum: Yes it preserves the order

hmhealey: It's definitely a bit weird to look at, but in JS, Sets are iterated in insertion order, so this is correct.

enahum:
cherry-picked 5.4

PR #666, opened 2018-10-04
 MM-11477 Changes required to intercept leaked network errors on mobile

hmhealey:
Somewhere, we're not properly catching the error objects thrown by Client4 which has been causing them to be logged by Sentry for months, but due to how Sentry handles things that are thrown that aren't actual errors, it provides no information about where we're not handling those.

By using a modified version of the thunk middleware on mobile, we'll be able to intercept those events and replace them with more useful logs on Sentry.

#### Ticket Link

### DIFF
--- a/src/store/configureStore.dev.js
+++ b/src/store/configureStore.dev.js
@@ -28,6 +28,17 @@ import initialState from './initial_state';
 import {offlineConfig, createReducer} from './helpers';
 import {createMiddleware} from './middleware';
 
+/**
+ * Configures and constructs the redux store. Accepts the following parameters:
+ * preloadedState - Any preloaded state to be applied to the store after it is initially configured.
+ * appReducer - An object containing any app-specific reducer functions that the client needs.
+ * userOfflineConfig - Any additional configuration data to be passed into redux-offline aside from the default values.
+ * getAppReducer - A function that returns the appReducer as defined above. Only used in development to enable hot reloading.
+ * clientOptions - An object containing additional options used when configuring the redux store. The following options are available:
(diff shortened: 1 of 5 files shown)


src/store/middleware.js
```diff
@@ -0,0 +1,48 @@
...
+    enableThunk: true,
+};
+
+/***
+clientOptions object - This param allows users to configure the store from the client side.
+It has the following properties:
+additionalMiddleware - func | array - Allows for single or multiple additional middleware functions to be passed in from the client side.
+enableBuffer - bool - default = true - If true, the store will buffer all actions until offline state rehydration occurs.
+enableThunk - bool - default = true - If true, include the thunk middleware automatically. If false, thunk must be provided as part of additionalMiddleware.
+***/
```
jwilander: Does this comment make sense here?

src/store/configureStore.dev.js
```diff
@@ -28,40 +25,21 @@ const devToolsEnhancer = (
...
-import {defaultOptions, offlineConfig, createReducer} from './helpers';
+import {offlineConfig, createReducer} from './helpers';
+import {createMiddleware} from './middleware';
 
-/***
-clientOptions object - This param allows users to configure the store from the client side.
-It has two properties currently:
-enableBuffer - bool - default = true - If true the store will buffer all actions until offline state rehydration occurs.
-additionalMiddleware - func | array - Allows for single or multiple additional middleware functions to be passed in from the client side.
-***/
```
jwilander: Why remove the comments?

hmhealey:
@jwilander I moved those comments back to the configureStore methods and I added the other parameters to them so that they document more than just the settings used to configure middleware

PR #668, opened 2018-10-05
GH-8347 added the selector for getting channels with user profiles

pradeepmurugesan:
#### Summary
Selectors for getting the channels with the user profile.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/8347

#### Checklist

#### Test Information
This PR was tested on: Mac OS Sierra, 10.12.6

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -23,7 +23,7 @@ import {
 import {getLastPostPerChannel, getAllPosts} from 'selectors/entities/posts';
 import {getCurrentTeamId, getCurrentTeamMembership} from 'selectors/entities/teams';
 import {haveICurrentChannelPermission} from 'selectors/entities/roles';
-import {isCurrentUserSystemAdmin, getCurrentUserId} from 'selectors/entities/users';
+import {isCurrentUserSystemAdmin, getCurrentUserId, makeGetProfilesInChannel} from 'selectors/entities/users';
 
 import {
     buildDisplayableChannelList,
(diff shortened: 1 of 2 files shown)


src/selectors/entities/channels.js
```diff
@@ -764,3 +764,30 @@ export const filterPostIds = (condition) => {
...
 };
+
+const doGetProfilesInChannel = makeGetProfilesInChannel();
+const getOtherProfilesInChannel = (state) => {
+    return (channelId, currentUserId) => {
+        return doGetProfilesInChannel(state, channelId).filter((profile) => profile.id !== currentUserId);
+    };
+};
+export const getChannelsWithUserProfiles = createSelector(
+    getOtherProfilesInChannel,
```
hmhealey: I'm not sure how familiar you are with [reselect selectors](https://github.com/reduxjs/reselect), but this one doesn't work correctly because `getOtherProfilesInChannel` returns a new function every time it's called, meaning that `getChannelsWithUserProfiles` recalculates its results each time it is ran. A better way to do this would be to get rid of `getOtherProfilesInChannel` and instead pass `getUserIdsInChannels` (which returns an object mapping channel ID to user IDs) and `getUsers` (which returns an object mapping user IDs to users).

pradeepmurugesan:
@hmhealey I have made the necessary changes. I understand that we can replace matchExistsInChannelProfiles with filterProfilesMatchingTerm.

I will work on the webapp changes once the redux changes gets merged.

Kindly review

PR #669, opened 2018-10-06
MM-11042: Migrate to Flow file src/actions/users.js in mattermost-redux

cometkim:
#### Summary
Migrate to Flow file src/actions/users.js in mattermost-redux

#### Ticket Link
- [Jira Ticket](https://mattermost.atlassian.net/browse/MM-11042)
- [GitHub Issue](https://github.com/mattermost/mattermost-server/issues/9302)

#### Checklist
- ~~Ran `make test` to ensure unit tests passed~~
- ~~Added or updated unit tests (required for all new features)~~
- ~~All new/modified APIs include changes to the drivers~~

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -214,7 +214,7 @@ function completeLogin(data: UserProfile): ActionFunc {
             }
         }
         if (roles.size > 0) {
-            dispatch(loadRolesIfNeeded([...roles]));
+            dispatch(loadRolesIfNeeded(roles));
         }
 
         return {data: true};
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 // @flow
 
-import type {ActionFunc, DispatchFunc, GetStateFunc} from 'types/actions';
+import type {ActionFunc, ActionResult, DispatchFunc, GetStateFunc} from 'types/actions';
 import type {UserProfile} from 'types/users';
 import type {TeamMembership} from 'types/teams';
 
@@ -38,7 +38,7 @@ import {getConfig} from 'selectors/entities/general';
 import {getCurrentUser, getCurrentUserId} from 'selectors/entities/users';
 
 export function checkMfa(loginId: string): ActionFunc {
-    return async (dispatch, getState) => {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         dispatch({type: UserTypes.CHECK_MFA_REQUEST, data: null}, getState);
         try {
             const data = await Client4.checkUserMfa(loginId);
@@ -65,10 +65,10 @@ export function generateMfaSecret(userId: string): ActionFunc {
 }
 
 export function createUser(user: UserProfile, token: string, inviteId: string): ActionFunc {
-    return async (dispatch, getState) => {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         dispatch({type: UserTypes.CREATE_USER_REQUEST, data: null}, getState);
 
-        let created: UserProfile;
+        let created = null;
         try {
             created = await Client4.createUser(user, token, inviteId);
         } catch (error) {
@@ -83,8 +83,9 @@ export function createUser(user: UserProfile, token: string, inviteId: string): 
             return {error};
         }
 
-        const profiles: { [string]: UserProfile } = {};
-        profiles[created.id] = created;
+        const profiles: { [userId: string]: UserProfile } = {
+            [created.id]: created,
+        };
         dispatch({type: UserTypes.RECEIVED_PROFILES, data: profiles});
 
         return {data: created};
@@ -92,12 +93,12 @@ export function createUser(user: UserProfile, token: string, inviteId: string): 
 }
 
 export function login(loginId: string, password: string, mfaToken: string = '', ldapOnly: boolean = false): ActionFunc {
-    return async (dispatch, getState) => {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         dispatch({type: UserTypes.LOGIN_REQUEST, data: null}, getState);
 
         const deviceId = getState().entities.general.deviceToken;
 
-        let data: UserProfile;
+        let data = null;
         try {
             data = await Client4.login(loginId, password, mfaToken, deviceId, ldapOnly);
         } catch (error) {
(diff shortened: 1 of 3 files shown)


src/actions/users.js
```diff
@@ -323,12 +330,12 @@ export function getMissingProfilesByIds(userIds) {
     };
 }
 
-export function getMissingProfilesByUsernames(usernames) {
+export function getMissingProfilesByUsernames(usernames: Array<string>): ActionFunc {
     return async (dispatch, getState) => {
         const {profiles} = getState().entities.users;
 
         const usernameProfiles = Object.values(profiles).reduce((acc, profile) => {
```
cometkim:
```
Cannot cast profile to UserProfile because inexact mixed [1] is incompatible with exact UserProfile [2].

     src/actions/users.js
     335│         const {profiles} = getState().entities.users;
     336│
     337│         const usernameProfiles = Object.values(profiles).reduce((acc, profile) => {
 [2] 338│             acc[(profile: UserProfile).username] = profile;
     339│             return acc;
     340│         }, {});
     341│

     /tmp/flow/flowlib_1bc5a954/core.js
 [1]  62│     static values(object: any): Array<mixed>;
```

I don't think this is wrong...

Any workaround for Object.values()?

cometkim:
Currently, the file has 76.3% type coverage and 9 type errors. I will fix.

cometkim:
76.0% and no errors. I think it's ready to review.

cometkim:
rebased

src/actions/users.js
```diff
@@ -208,15 +214,15 @@ function completeLogin(data) {
             }
         }
         if (roles.size > 0) {
-            dispatch(loadRolesIfNeeded(roles));
+            dispatch(loadRolesIfNeeded([...roles]));
```
lieut-data: I don't understand this change: this forces a duplication of `roles`. What makes this necessary?

src/actions/users.js
```diff
@@ -323,11 +331,11 @@ export function getMissingProfilesByIds(userIds) {
...
 }
 
-export function getMissingProfilesByUsernames(usernames) {
-    return async (dispatch, getState) => {
+export function getMissingProfilesByUsernames(usernames: Array<string>): ActionFunc {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         const {profiles} = getState().entities.users;
 
-        const usernameProfiles = Object.values(profiles).reduce((acc, profile) => {
+        const usernameProfiles = Object.values(profiles).reduce((acc, profile: any) => {
```
lieut-data: Can we use type `UserProfile` here instead of `any`?

src/actions/users.js
```diff
@@ -593,15 +601,15 @@ export function getProfilesNotInChannel(teamId, channelId, page, perPage = Gener
...
+export function getMe(): ActionFunc {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         const getMeFunc = bindClientFunc(
             Client4.getMe,
             UserTypes.USER_REQUEST,
             [UserTypes.RECEIVED_ME, UserTypes.USER_SUCCESS],
             UserTypes.USER_FAILURE
         );
-        const me = await getMeFunc(dispatch, getState);
+        const me: $Subtype<ActionResult> = await getMeFunc(dispatch, getState);
```
lieut-data: When `Client4.getMe` eventually gets annotated, how do we reflects its type accurately for clients of this action creator?

src/actions/users.js
cometkim: Yep. this part is necessary for:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/actions/users.js:217:40

Cannot call loadRolesIfNeeded with roles bound to roles because Set [1] is incompatible with array type [2].

     src/actions/users.js
 [1] 207│         const roles = new Set();
        :
     214│             }
     215│         }
     216│         if (roles.size > 0) {
     217│             dispatch(loadRolesIfNeeded(roles));
     218│         }
     219│
     220│         return {data: true};

     src/actions/roles.js
 [2]  62│ export function loadRolesIfNeeded(roles: Array<string>): ActionFunc {
```

src/actions/users.js
cometkim: To avoid the issue: https://github.com/facebook/flow/issues/2221

Flow infers that could be `mixed`... IDK, probably something better workaround there.

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/actions/users.js:339:25

Cannot get profile.username because property username is missing in mixed [1].

     src/actions/users.js
     336│         const {profiles} = getState().entities.users;
     337│
     338│         const usernameProfiles = Object.values(profiles).reduce((acc, profile) => {
     339│             acc[profile.username] = profile;
     340│             return acc;
     341│         }, {});
     342│

     /tmp/flow/flowlib_113cee6b/core.js
 [1]  62│     static values(object: any): Array<mixed>;
```

src/actions/users.js
cometkim: Currently, can use it only if it is compatible with the `ActionResult: {data, error}` type. If the client's return type is compatible with `{data}`, this is not a problem.

But later ActionResult's data needs to be improved to infer the "actual" type of data.

src/actions/users.js
lieut-data: Aha, I missed the nuance that `roles` was a `Set`. I suppose flow doesn't model a supertype that both an `Array` and `Set` could match against for use on `loadRolesIfNeeded`?

src/actions/users.js
lieut-data: Cool, thanks for teaching me something!

src/actions/users.js
lieut-data: I like it! Your call on whether to do it now or in the future :)

lieut-data:
Looks like my reviewer changes got lost in the GitHub outage -- actually added @saturninoabril now :)

src/actions/users.js
saturninoabril:
```suggestion
        const usernameProfiles = (Object.values(profiles): any).reduce((acc, profile: UserProfile) => {
```

Above works but maybe not necessary.

PR #670, opened 2018-10-08
MM-12521 Fix custom emoji not returning results

sudheerDev:
#### Summary
Fixes custom emoji returning empty results

If serverVersion is not stored client4 class isMinimumServerVersion utilty returns false. So, adding back version info into the class.

Reverts part of this commit: https://github.com/mattermost/mattermost-redux/commit/2514cbdff6cc3d6e177202b5f5e37049e0a27eb5#diff-6655fe9b9bb92382c9e8084066aa895e

#### Ticket Link
[MM-12521](https://mattermost.atlassian.net/browse/MM-12521)

#### Checklist

#### Test Information
This PR was tested on: web


sudheerDev:
@enahum Yes, I think we use this in a few other places so we need it for now. Down the lane, we should depend only the redux store version or this class value than having two entries in with same value.

PR #673, opened 2018-10-10
MM-12346: Checks permissions to edit post regardless of license.

mkraft:
#### Summary
Checking permissions regardless of whether user has a license or not.

Basically this just removes the outer `license.IsLicensed === 'true'` conditional which I don't think is necessary since permissions must be verified regardless of license status.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12346

#### Checklist

#### Test Information
iPhone 6, iOS 12.0 (emulator)


grundleborg:
Yeah, the *timelimit* specifically has always been an Enterprise feature, and remains so, I believe.

PR #685, opened 2018-10-24
fix default language import

zetaab:
Please make sure you've read the [pull request](http://docs.mattermost.com/developer/contribution-guide.html#preparing-a-pull-request) section of our [code contribution guidelines](http://docs.mattermost.com/developer/contribution-guide.html).

When filling in a section please remove the help text and the above text.

#### Summary
Currently default language is not loaded correctly. See screenshots in issue link.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/9733

#### Checklist

#### Test Information

This PR is not tested. How I can compile this to inside mattermost-webapp? Usually when building mattermost-webapp I just write `make clean && make build` and then compile mattermost-server again. How I can use custom mattermost-redux there? (now its coming from node_moduels)

### DIFF
--- a/src/selectors/entities/i18n.js
+++ b/src/selectors/entities/i18n.js
@@ -3,9 +3,9 @@
 
 import {getCurrentUser} from 'selectors/entities/common';
 
-import {DEFAULT_LOCALE} from 'constants/general';
+import {General} from 'constants';
 
-export function getCurrentUserLocale(state, defaultLocale = DEFAULT_LOCALE) {
+export function getCurrentUserLocale(state, defaultLocale = General.DEFAULT_LOCALE) {
     const currentUser = getCurrentUser(state);
 
     if (!currentUser) {
--- a/src/selectors/entities/teams.js
+++ b/src/selectors/entities/teams.js
@@ -3,7 +3,7 @@
 
 import {createSelector} from 'reselect';
 
-import {DEFAULT_LOCALE} from 'constants/general';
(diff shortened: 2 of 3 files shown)


zetaab:
I am not sure does this solve this problem at all, I do not see any difference. Still I have locale some id instead of default locale. How to reproduce this problem: update your user Locale to `''` in database and upload many files in same post. It might need few posts and then UI will crash

zetaab:
now the default locale things are working. However, still something is wrong because weird locale is passed from `makeGetFilesForPost` to `sortFileInfos`. 

screen shot 2018-10-25 at 1 18 56

I solved this issue in our mattermost setup with following
```
MariaDB [mattermost]> update Users set Locale='en' where locale='';
Query OK, 652 rows affected (0.03 sec)
Rows matched: 652  Changed: 652  Warnings: 0
MariaDB [mattermost]> alter table Users alter Locale SET DEFAULT 'en';
Query OK, 0 rows affected (0.06 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

After that all channels which were broken works like should.

hmhealey:
Great catch. I'm pretty sure those bad imports were my fault.

You're saying you still get the error sometimes even with these changes though? If that's the case, maybe take a look at `getCurrentUserLocale` itself to see where else it might be coming from. I haven't been able to reproduce this error myself, but I have got some reports where it's affecting the mobile apps as well

PR #687, opened 2018-10-26
Selectors to return files and reactions metadata if present

enahum:
#### Summary
If the post metadata if present and it has reactions and files make the selectors use the metadata available

### DIFF
--- a/src/selectors/entities/files.js
+++ b/src/selectors/entities/files.js
@@ -4,6 +4,7 @@
 import {createSelector} from 'reselect';
 
 import {getCurrentUserLocale} from 'selectors/entities/i18n';
+import {getPost} from 'selectors/entities/posts';
 
 import {sortFileInfos} from 'utils/file_utils';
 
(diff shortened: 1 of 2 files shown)


src/selectors/entities/files.js
```diff
@@ -25,9 +26,13 @@ export function getFilePublicLink(state) {
 
 export function makeGetFilesForPost() {
     return createSelector(
-        [getAllFiles, getFilesIdsForPost, getCurrentUserLocale],
-        (allFiles, fileIdsForPost, locale) => {
-            const fileInfos = fileIdsForPost.map((id) => allFiles[id]).filter((id) => Boolean(id));
+        [getPost, getAllFiles, getFilesIdsForPost, getCurrentUserLocale],
+        (post, allFiles, fileIdsForPost, locale) => {
+            let fileInfos = post.metadata && post.metadata.files && [...post.metadata.files];
```
sudheerDev: Do we need this? I thought we are making changes to have files in the same key as before.

enahum: 0/5 on how to do it, this is definitely the easiest path

sudheerDev: This is definitely easy. I am confused on the approach we are talking with respect to the having files in state.

From last discussion offline we decided to have files in the same redux key as before so i am curios as to which direction we are going?

@hmhealey thoughts?

enahum:
closing in favor of #689

PR #689, opened 2018-10-29
MM-12827 Populate emoji and file state from post metadata 

hmhealey:
As discussed on Friday, this makes it so that `post.metadata.emojis` and `post.metadata.files` are removed from the post object when it is stored in the redux state, and then they are added to their respective stores.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12827

#### Checklist

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -938,11 +938,15 @@ export function getPostsAfterWithRetry(channelId, postId, page = 0, perPage = Po
 }
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
-export async function getProfilesAndStatusesForPosts(postsArrayOrMap, dispatch, getState) {
+export function getProfilesAndStatusesForPosts(postsArrayOrMap, dispatch, getState) {
+    if (!postsArrayOrMap) {
+        // Some API methods return {error} for no results
+        return Promise.resolve();
+    }
+
     const posts = Object.values(postsArrayOrMap);
 
-    if (!posts || posts.length === 0) {
-        // Some API methods return {error} for no results
+    if (posts.length === 0) {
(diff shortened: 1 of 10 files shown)


src/reducers/entities/posts.js
```diff
@@ -158,6 +160,23 @@ function handleReceivedPosts(posts = {}, postsInChannel = {}, postsInThread = {}
...
+export function removeUnneededMetadata(post) {
+    if (!post.metadata || (!post.metadata.emojis && !post.metadata.files)) {
+        return post;
+    }
+
+    const metadata = {...post.metadata};
+
+    // These fields are stored separately
+    Reflect.deleteProperty(metadata, 'emojis');
+    Reflect.deleteProperty(metadata, 'files');
```
enahum: do not forget about reactions here

src/actions/posts.js
```diff
@@ -938,15 +938,18 @@ export function getPostsAfterWithRetry(channelId, postId, page = 0, perPage = Po
 }
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
-export async function getProfilesAndStatusesForPosts(posts, dispatch, getState) {
-    if (!posts) {
+export async function getProfilesAndStatusesForPosts(postsArrayOrMap, dispatch, getState) {
+    const posts = Object.values(postsArrayOrMap);
```
sudheerDev: `postsArrayOrMap` will this be only object?

hmhealey: It can be an array of posts or an object mapping post IDs to posts. `Object.values` converts it into an array of posts either way. It's always done that, but it used to do it later in the process.

I did just realize this doesn't null check any more though so I'll fix that

PR #690, opened 2018-10-29
MM-12687 Shallow clone data to avoid state mutation in analytics reducer

jwilander:
#### Summary
Shallow clone data to avoid state mutation in analytics reducer.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12687

#### Checklist

### DIFF
--- a/src/reducers/entities/admin.js
+++ b/src/reducers/entities/admin.js
@@ -126,20 +126,21 @@ function samlCertStatus(state = {}, action) {
 
 function convertAnalyticsRowsToStats(data, name) {
     const stats = {};
+    const clonedData = [...data];
 
     if (name === 'post_counts_day') {
-        data.reverse();
-        stats[Stats.POST_PER_DAY] = data;
+        clonedData.reverse();
+        stats[Stats.POST_PER_DAY] = clonedData;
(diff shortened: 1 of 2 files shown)


hmhealey:
Is this something we can add a test for? At the very least, it can just call the function but wrap the state with deepFreeze since that'll throw an error if it mutates the state

jwilander:
Good call @hmhealey, test added

src/reducers/entities/admin.js
```diff
@@ -124,22 +124,23 @@ function samlCertStatus(state = {}, action) {
...
 
-function convertAnalyticsRowsToStats(data, name) {
+export function convertAnalyticsRowsToStats(data, name) {
     const stats = {};
+    const clonedData = [...data];
 
     if (name === 'post_counts_day') {
-        data.reverse();
-        stats[Stats.POST_PER_DAY] = data;
+        clonedData.reverse();
```
sudheerDev: Nit: As this has hardly has any performance overhead.

Looks like the mutation is caused by just by reverse. Could we just instead do `[...data].reverse()` so we don't clone if we are not reversing the order?

PR #693, opened 2018-10-30
MM-12233: local user mention results

lieut-data:
#### Summary
Populate `profilesInChannel` with posts received for the channel. As noted in the comments, this is not proof that the user is still in the channel: they may have left. However, if we are viewing a post as such, they are likely to still be "in context", and are useful to mention as such.

This is currently a work in progress, and not necessarily ready for review.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12233

#### Checklist

#### Test Information
This PR was tested on: Chrome, OSX Mojave

### DIFF
--- a/src/reducers/entities/users.js
+++ b/src/reducers/entities/users.js
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {combineReducers} from 'redux';
-import {UserTypes} from 'action_types';
+import {UserTypes, PostTypes} from 'action_types';
 import {profileListToMap} from 'utils/user_utils';
 
 function profilesToSet(state, action) {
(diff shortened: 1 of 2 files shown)


lieut-data:
@jasonblais, I've marked this as PM review specifically to get your feedback on the choice to populate profiles in channel using post data, but recognizing the limitation that the user might have left/been kicked from the channel above. This PR is an addition to https://github.com/mattermost/mattermost-webapp/pull/1974, allowing for results on first load vs. after the first at mention query, but isn't strictly required.

@enahum, @sudheerDev, marking this for dev review since it's ready as such.

lieut-data:
After some discussion, we agreed that pre-flighting the data here causes more problems than it solves, since other parts of the UI rely on this reducer. Closing in favour of the base webapp changes.

PR #704, opened 2018-11-07
MM-11477 Wrap objects thrown by Client4 in a proper error type

hmhealey:
This is another attempt at stopping the "non-error exceptions" that Sentry keeps getting because we're not catching the errors thrown by Client4 correctly, and Sentry doesn't provide any helpful debugging information unless you're throwing an actual Error object.

#### Ticket Link

#### Checklist

### DIFF
--- a/test/actions/helpers.test.js
+++ b/test/actions/helpers.test.js
@@ -1,0 +1,129 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import assert from 'assert';
+
+import {UserTypes} from 'action_types';
+import {forceLogoutIfNecessary} from 'actions/helpers';
+import {Client4} from 'client';
+import {ClientError} from 'client/client4';
+import configureStore, {mockDispatch} from 'test/test_store';
+
+describe('Actions.Helpers', () => {
+    describe('forceLogoutIfNecessary', () => {
+        const token = 'token';
+
+        beforeEach(() => {
+            Client4.setToken(token);
+        });
+
+        it('should do nothing when passed a client error', async () => {
+            const store = await configureStore({
+                entities: {
+                    users: {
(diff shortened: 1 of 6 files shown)


test/actions/helpers.test.js
```diff
@@ -0,0 +1,129 @@
...
+describe('Actions.Helpers', () => {
+    describe('forceLogoutIfNecessary', () => {
+        const token = 'token';
+
+        beforeEach(() => {
+            Client4.setToken(token);
+        });
+
+        it('should do nothing when passed a client error', async () => {
+            const store = await configureStore({
```
jespino: Maybe we can move the store and dispatch initialization to the beforeEach, and handle the unique difference in the `'should do nothing when not logged in'` test case configuring again the store.

test/utils/sentry.test.js
```diff
@@ -0,0 +1,41 @@
...
+import {cleanUrlForLogging} from 'utils/sentry';
+
+describe('utils/sentry', () => {
+    describe('cleanUrlForLogging', () => {
+        const baseUrl = 'https://mattermost.example.com/subpath';
+
+        const client = new Client4();
+        client.setUrl(baseUrl);
+
+        const tests = [{
```
jespino: I love to incorporate the table testing in javascript part too :)

test/actions/helpers.test.js
hmhealey: I'm kind of on the fence about how this is done. On one hand, I like that this keeps all of the code for the test in one place, and it ensures that the tests don't depend on each other, but on the other, it is a lot of duplication in some of these cases.

I've wanted to write a more declarative way to construct the store for a while now (something like `const store = TestStore.addUser({current: true}).configure()`) that would kind of solve both problems, but last time I tried that, it got really complicated. Maybe it would work for simple cases like this though

PR #706, opened 2018-11-12
Ldap groups phase1

mkraft:
#### Summary
LDAP groups sync epic

#### Checklist

#### Test Information
Chrome browser, MacOS


src/actions/groups.js
```diff
@@ -0,0 +1,163 @@
...
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: GroupTypes.LINK_GROUP_SYNCABLE_FAILURE, error, data: {groupID, syncableID, type: syncableType}},
+                logError(error),
+            ]));
+            return {error};
+        }
+
+        var type;
```
grundleborg: 0/5 but any reason to use `var` over `let` here?

src/actions/groups.js
```diff
@@ -0,0 +1,163 @@
...
+            {type, data},
+        ]));
+
+        return {data: true};
+    };
+}
+
+export function getGroupSyncables(groupID: string, syncableType: SyncableType): ActionFunc {
+    return async (dispatch, getState) => {
+        dispatch({type: GroupTypes.GET_GROUP_SYNCABLES_REQUEST, data: {groupID, syncableType}});
```
grundleborg: In the action above, it's `type: syncableType` but here it's just `syncableType`. Is there a reason for the inconsistency?

src/client/client4.js
```diff
@@ -1143,6 +1143,13 @@ export default class Client4 {
 
     // Channel Routes
 
+    getAllChannels = async () => {
```
grundleborg: Should there be `page` and `perPage` parameters here?

src/store/initial_state.js
```diff
@@ -617,6 +624,28 @@ const state: GlobalState = {
                 error: null,
             },
         },
+        groups: {
```
grundleborg: It looks like the default request state for `getAllChannels` and the 3 ldap/admin requests are missing.

src/types/requests.js
```diff
@@ -164,3 +164,11 @@ export type SchemesRequestsStatuses = {|
...
     getSchemeChannels: RequestStatusType
 |};
+
+export type GroupsRequestsStatuses = {|
+    linkGroupSyncable: RequestStatusType,
+    unlinkGroupSyncable: RequestStatusType,
+    getGroupSyncables: RequestStatusType,
+    getGroupMembers: RequestStatusType,
+    getGroup: RequestStatusType,
+|};
```
grundleborg: Could you add the new channel and admin requests here too?

src/actions/groups.js
grundleborg: Also could ask the question about a few other places, but since it's 0/5 I'm not going to be pedantic - more just curious about the choice.

src/actions/admin.js
```diff
@@ -188,6 +188,74 @@ export function syncLdap(): ActionFunc {
     );
 }
 
+export function getLdapGroups(page: number = 0, perPage: number = General.LOGS_PAGE_SIZE_DEFAULT): ActionFunc {
```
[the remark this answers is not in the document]
jespino: @mkraft :point_up:

src/client/client4.js
jespino: Yes, I think so. I going to review that.

src/store/initial_state.js
jespino: Good point.

src/client/client4.js
jespino: Fixed

src/store/initial_state.js
jespino: Fixed

src/types/requests.js
jespino: Fixed

src/reducers/entities/channels.js
```diff
@@ -37,6 +37,17 @@ function currentChannelId(state = '', action) {
     }
 }
 
+function allChannels(state = {}, action) {
+    switch (action.type) {
+    case ChannelTypes.RECEIVED_ALL_CHANNELS:
+        return action.data;
```
jwilander: The `getAllChannels` action has paging, so if you were to page through a couple pages this would clobber previous results

src/reducers/entities/channels.js
```diff
@@ -437,6 +448,9 @@ export default combineReducers({
     // object where every key is the channel id and has and object with the channel detail
     channels,
 
+    // object where every key is the channel id and has and object with the channel detail for all channels list
+    allChannels,
```
jwilander: Why the new reducer? Why not just use the `channels` reducer?

src/reducers/entities/channels.js
mkraft: @jespino ☝️

src/reducers/entities/channels.js
mkraft: @jespino ☝️

src/reducers/entities/channels.js
jespino: Yes, that is the intention here. Actually, we show the "last" getAllChannels or searchAllChannels results to show in the interface. And the idea is to not interfere with the normal channels used in other places (because if I wipe out that channels can generate problems in other places).

src/reducers/entities/channels.js
jespino: The same reason here, is to have a different place to list or iterate over a bunch of all the channels, not the "channels loaded to use allover the application"

src/reducers/entities/channels.js
jwilander: Hmm if it's so ephemeral do we need to store it in redux at all? Would local state in the component be better?

Adding this reducer really adds a lot of potential for confusion. If I'm not familiar with exactly why this reducer exists how do I know when to look at `channels` versus`allChannels`? When should I use the `getAllChannels` action versus `getChannels`? How do I know the `allChannels` reducer is going to get purged before every new `getAllChannels` call, when none of our other actions follow that pattern?

jwilander: If it does need to be in redux, I think it needs to mesh better with our existing reducers and be clear as to what it's for

hmhealey: Adding to this discussion, the intent of `channels` is to hold a list of information about any channel that the client knows about, not necessarily the user's current channels. Anything else that would store a channel just stores an ID to look up the entry in `channels` so that we aren't duplicating this data. Perhaps any of these other channels should be stored in `channels` and then we store the IDs that we're showing on that screen elsewhere (like in the webapp-specific `views` reducer)?

jespino: Yes, I think you are right, I going to move this to the views, and try to reuse the channels.

src/selectors/entities/groups.js
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {createSelector} from 'reselect';
+
+export function getAllGroups(state) {
+    return state.entities.groups.groups;
+}
+
+export const getGroup = createSelector(
```
hmhealey: Most of these selectors should be plain functions since they have arguments and they're mostly returning objects straight from the store.

jespino:
@jwilander I'm not 5/5 about that, because I don't want to interfere with the current channels usage. I going to check the channels usage in mattermost-webapp and mattermost-mobile and check if we can store that data there.

jespino:
@jwilander Now is ready, I'm using now the channels reducer to store the data in channels, and then store apart the ids only for show the results, and use the channels selector to build the new list of channels.

src/reducers/entities/channels.js
jespino: Done!

src/selectors/entities/groups.js
jespino: Fixed.

jespino:
@hmhealey PTAL

src/selectors/entities/groups.js
```diff
@@ -0,0 +1,58 @@
...
+export function getGroupMemberCount(state, id) {
+    const memberData = state.entities.groups.members;
+    const groupMemberData = memberData[id];
+    if (!groupMemberData) {
+        return 0;
+    }
+    return memberData[id].totalMemberCount;
+}
+
+const getGroupSyncables = createSelector(
```
hmhealey: `getGroupSyncables`, `getGroupTeams`, `getGroupChannels`, and `getGroupMembers` should be factory functions that return selectors, otherwise they'll overwrite each others' memoization. Either that or we can just remove the usage of `createSelector` since they're returning primitives without doing much work.

jespino: Fixed, PTAL

PR #709, opened 2018-11-14
MM-13020 Make ClientError.message enumerable

hmhealey:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13020

#### Checklist

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -2604,6 +2604,8 @@ export class ClientError extends Error {
         this.server_error_id = data.server_error_id;
         this.status_code = data.status_code;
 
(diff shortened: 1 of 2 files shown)


src/client/client4.js
```diff
@@ -2603,5 +2603,7 @@ export class ClientError extends Error {
         this.intl = data.intl;
         this.server_error_id = data.server_error_id;
         this.status_code = data.status_code;
+
```
lieut-data: Is it worth adding a comment here explaining the context? Perhaps something like:
```suggestion

// Ensure message is treated as a property of this class when object spreading.
```

hmhealey: Good call. I'll add that

PR #710, opened 2018-11-15
MM-12844/MM-13001 Add interactive dialog actions

jwilander:
#### Summary
Add support for interactive dialogs. See ticket for more details.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12844
https://mattermost.atlassian.net/browse/MM-13001

#### Checklist


src/actions/websocket.js
```diff
@@ -808,6 +812,11 @@ function handlePluginStatusesChangedEvent(msg, dispatch) {
     });
 }
 
+function handleOpenDialogEvent(msg, dispatch) {
```
enahum: 👍

test/actions/websocket.test.js
```diff
@@ -702,4 +702,28 @@ describe('Actions.Websocket', () => {
...
+
+            const dialog = state.entities.integrations.dialog;
+            assert.ok(dialog);
+            assert.ok(dialog.url === 'someurl');
+            assert.ok(dialog.trigger_id === 'sometriggerid');
+            assert.ok(dialog.dialog);
+            done();
+        }
+
+        test();
```
enahum: seems that this test is failing

jwilander: Grrr it passed locally

jwilander: Ahh it's a stlye issue

jwilander:
@enahum fixed

enahum:
@jwilander weird but is still failing
image

jwilander:
I have no idea why, the other build passed and it passes locally

enahum:
@saturninoabril would you know?

saturninoabril:
I'll take a look.

saturninoabril:
Looks like not related to PR.  Might be on PR build server -> `mocha: not found`

saturninoabril:
Weird, it suddenly passed.

jwilander:
FYI @enahum pushed a new commit moving some utils code over from the webapp

src/utils/integration_utils.js
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function checkDialogElementForError(elem, value) {
+    if (!value && !elem.optional) {
+        return {
+            id: 'interactive_dialog.error.required',
```
enahum: do we have this id's defined somewhere in the web and mobile apps? and the ones below? cause if we don't they are not going to get picked up by @jespino tool.

Also not sure if this localization stuff should be in redux, perhaps return a constant and use that constant in the apps?

jwilander: The same IDs will be used in both the webapp and RN. I don't really see a difference between returning a localization ID and creating a separate constant that would probably be the same thing

jwilander: Also the translations are already stored in redux state, right?

enahum: sounds good, about the id instead of the constant, but the translations are NOT in the redux store, perhaps for the webapp but NOT in the mobile app for sure

PR #711, opened 2018-11-15
WebSocket actions improvement

enahum:
#### Summary
I was looking to get posts as soon as the WS reconnects and fixing up a bit the reconnect logic, and I ended up modifying a bit of code to make the WebSocket actions a bit better.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12832
https://mattermost.atlassian.net/browse/MM-13039

### DIFF
--- a/src/client/websocket_client.js
+++ b/src/client/websocket_client.js
@@ -14,18 +14,17 @@ class WebSocketClient {
         this.token = null;
         this.sequence = 1;
         this.connectFailCount = 0;
-        this.eventAction = null;
-        this.firstConnectAction = null;
-        this.reconnectAction = null;
-        this.errorAction = null;
-        this.closeAction = null;
-        this.connectingAction = null;
-        this.dispatch = null;
+        this.eventCallback = null;
+        this.firstConnectCallback = null;
+        this.reconnectCallback = null;
+        this.errorCallback = null;
+        this.closeCallback = null;
+        this.connectingCallback = null;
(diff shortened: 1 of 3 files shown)


enahum:
Ticket to move the dispatcher here https://mattermost.atlassian.net/browse/MM-13039

src/client/websocket_client.js
```diff
@@ -172,35 +174,35 @@ class WebSocketClient {
...
+                } else if (this.eventAction) {
+                    dispatch(this.eventAction(msg));
                 }
             };
         });
     }
 
-    setConnectingCallback(callback) {
-        this.connectingCallback = callback;
+    setConnectingAction(action) {
```
hmhealey: This is going to break anyone else using the websocket client. I'm pretty sure we want backwards compatibility here

enahum: this is going to break anyway as we changed the way the previous callbacks now actions are called

enahum: But it should not be a problem as in reality people should be using the WS actions and not the client directly

enahum:
@jwilander @hmhealey refactor this a bit more to remove the `dispatch` from the client

enahum:
rebased with the latest master that includes the dialogs

enahum:
I'm not QA but I did test it using the mobile app and it looked good, but yeah, I can submit a PR to the mobile repo and build it so QA can test

PR #712, opened 2018-11-16
Fix fetchMyChannelsAndMembers error handler

enahum:
#### Summary
By having the client calls outside of the try/catch the app was reporting a promise not being handled correctly

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -515,8 +515,10 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
         let channels;
         let channelMembers;
         try {
-            channels = await Client4.getMyChannels(teamId);
-            channelMembers = await Client4.getMyChannelMembers(teamId);
+            const channelRequest = Client4.getMyChannels(teamId);
+            const memberRequest = Client4.getMyChannelMembers(teamId);
+            channels = await channelRequest;
+            channelMembers = await memberRequest;
         } catch (error) {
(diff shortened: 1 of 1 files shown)


enahum:
@jwilander if I make that change then we'll need to batch the action to mark both as failed, that should work, right?

jwilander:
Yeah I think that should work 👍

enahum:
@jwilander done, we were already dispatching both errors

src/actions/channels.js
```diff
@@ -512,25 +512,11 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
...
-                logError(error),
-            ]), getState);
-            return {error};
-        }
-
         let channelMembers;
         try {
-            channelMembers = await channelMembersRequest;
+            channels = await Client4.getMyChannels(teamId);
+            channelMembers = await Client4.getMyChannelMembers(teamId);
```
jwilander: I think you need to do:

```javascript
const channelRequest = Client4.getMyChannels(teamId);
const memberRequest = Client4.getMyChannelMembers(teamId);
await channelRequest;
await memberRequest
```
So that they happen in parallel

enahum: @jwilander @hmhealey changed so the requests are executed in parallel

enahum:
This thing is failing on a totally unrelated test and the strange thing is that is passing locally

PR #720, opened 2018-11-22
Keep autocomplete generated display names for DMs

jespino:
#### Summary
When you get the channels by autocompleteChannels api, it returns a
generated display name Private channels, because they doesn't have it by
default (normally is the other user username, which depends on caller of
the API).

But when you load channels in any other way, or receive channels updates
from the websocket, this is overriden. This simply do not override it
with an empty value.

#### Checklist

### DIFF
--- a/src/reducers/entities/channels.js
+++ b/src/reducers/entities/channels.js
@@ -40,15 +40,20 @@ function currentChannelId(state = '', action) {
 function channels(state = {}, action) {
     switch (action.type) {
     case ChannelTypes.RECEIVED_CHANNEL:
+        if (state[action.data.id] && action.data.type === General.DM_CHANNEL) {
+            action.data.display_name = action.data.display_name || state[action.data.id].display_name;
+        }
         return {
             ...state,
             [action.data.id]: action.data,
         };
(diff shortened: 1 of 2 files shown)


jespino:
Yes, you are right, i change it in the last second and it was wrong. (Copy pasting). Yes, the idea is DMs. Fixed with a push -f.

enahum:
@jespino this is fine now BUT I do believe the selectors will try and complete the display_name anyway, can you please take a look at that as it is totally related to this PR

jespino:
@enahum yes, but it relays in the fact that you must have all the users data which have some conversation with you in the store. and that is not necessarily true during the autocomplete.

jespino:
Doesn't matter, because that display name (private channel display name for the search autocomplete) is the username. That is not reflected in real time, but at least solve the problem of "removing the name" when you are autocompleting and for some reason you receive an update of the channels from the server.

The point is to not remove the visual data from the channel autocomplete (on search). Maybe the information (in some corner cases) can be out of date, but is not a big deal. I can explain it better if you want or you need.

PR #722, opened 2018-11-27
MM-12574 Track events for interactive messages and dialogs

jwilander:
#### Summary
Track events for interactive messages and dialogs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12574

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -1553,6 +1553,12 @@ export default class Client4 {
     };
 
     doPostAction = async (postId, actionId, selectedOption = '') => {
+        if (selectedOption) {
+            this.trackEvent('api', 'api_interactive_messages_menu_selected');
+        } else {
+            this.trackEvent('api', 'api_interactive_messages_button_clicked');
+        }
+
         return this.doFetch(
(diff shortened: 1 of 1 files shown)


src/client/client4.js
```diff
@@ -2552,9 +2559,9 @@ export default class Client4 {
         });
     };
 
-    trackEvent(category, event, props) {
-        // Temporary change to only track one event to reduce data rate - see MM-13062
-        if (event !== 'api_posts_create') {
+    trackEvent(category, event, props, allow = false) {
+        // Temporary change to allow only certain events to reduce data rate - see MM-13062
+        if (!allow) {
```
grundleborg: I'd prefer we have an explicit whitelist in this function, because otherwise it's going to be annoying to track down what's whitelisted/not whitelisted by having to look all over the code in this and the web/mobile repos.

Also, did @jasonblais review this and sign off on tracking these events?

jasonblais: Yes, I signed off on this one.

jwilander: Makes sense, I'll make it more explicit

jwilander:
@grundleborg updated

PR #723, opened 2018-11-27
MM-12400:  Call get my terms of service only if the terms of service feature is enabled

chetanyakan:
#### Summary
Make API request only when terms are enabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-12400

#### Checklist

#### Test Information
This PR was tested on Ubuntu 18.04 / Chrome

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -225,6 +225,7 @@ function completeLogin(data: UserProfile): ActionFunc {
 export function loadMe(): ActionFunc {
     return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         const state = getState();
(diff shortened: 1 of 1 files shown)


jwilander:
@jasonblais is this meant for 5.6?

jasonblais:
@jwilander As far as I know, this is just a JavaScript error with no impact to end users, so not critical for 5.6.

jwilander:
Oh this is more than just a JS error, it's sending a bad request on every load of the clients when terms of service is not enabled. I definitely think we should fix this for 5.6, I don't want to have the servers spammed with bad requests

PR #724, opened 2018-11-28
[MM-13174] Add emoji utils and selectors to identify non-existing custom emoji

saturninoabril:
#### Summary
Add emoji utils and selectors to identify non-existing custom emoji.

We have this similar implementation in the mobile RN recently.  I'll update that after this is merged.

#### Ticket Link
Jira ticket: [MM-13174](https://mattermost.atlassian.net/browse/MM-13174)

#### Checklist

#### Test Information
This PR was tested on: [Chrome/FF, iOS simulator /Android emulator]

### DIFF
--- a/src/selectors/entities/emojis.js
+++ b/src/selectors/entities/emojis.js
@@ -3,6 +3,8 @@
 // @flow
 
 import {createSelector} from 'reselect';
+
+import {doesMatchNamedEmoji} from 'utils/emoji_utils';
 import {createIdsSelector} from 'utils/helpers';
 
 import type {GlobalState} from 'types/store';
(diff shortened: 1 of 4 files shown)


src/selectors/entities/emojis.js
```diff
@@ -44,3 +46,11 @@ export const getCustomEmojiIdsSortedByName: (state: GlobalState) => Array<string
         );
     }
 );
+
+export const existsInCustomEmojis: (state: GlobalState, emojiName: string) => boolean = createSelector(
+    getCustomEmojisByName,
+    (_, emojiName) => emojiName,
+    (customEmojisByName, emojiName) => {
+        return doesMatchNamedEmoji(`:${emojiName}:`) && customEmojisByName.has(emojiName);
```
jwilander: Curious, what's the point of the regex here? Wouldn't checking `customEmojisByName.has(emojiName)` do everything you need and have better performance than matching a regex?

Also making this a selector isn't going to help performance-wise as the "emojiName" argument is going to change often. If you care about the memoization you would need to follow the `makeExistsInCustomEmojis` pattern where it returns a function that creates a selector

saturninoabril: Thanks Joram, I have looked into this again and this PR is not necessary.  I'm closing this now, and will explain more on the webapp PR.

PR #728, opened 2018-12-03
MM-13195: Enforce at least one load of roles for the initial load

jespino:
#### Summary
This enforces at least call the load roles after the serverVersion
arrive once. This way we can be sure we don't leave the roles not loaded
during the initial page load.

#### Ticket Link
[MM-13195](https://mattermost.atlassian.net/browse/MM-13195)

#### Checklist

#### Test Information


jespino:
Yes, I don't love it either, but is the simplest, less disruptive solution I found.

PR #730, opened 2018-12-05
[MM-13174] Consider the result of getNeededCustomEmojis as non-existent emojis if post_metadata is enabled

saturninoabril:
#### Summary
Consider the result of getNeededCustomEmojis as non-existent emojis if post_metadata is enabled

#### Ticket Link
Jira ticket: [MM-13174](https://mattermost.atlassian.net/browse/MM-13174)
- added emoji action type of LOAD_NONEXISTENT_EMOJIS
- added emoji reducer to directly load non-existent emojis

#### Checklist

#### Test Information
This PR was tested on: [Chrome/FF, iOS emulator]

### DIFF
--- a/src/reducers/entities/emojis.js
+++ b/src/reducers/entities/emojis.js
@@ -9,7 +9,7 @@ import {
     UserTypes,
 } from 'action_types';
 
-import type {CustomEmoji} from '../../types/emojis';
+import type {CustomEmoji, NonExistentEmoji} from '../../types/emojis';
 import type {Post} from '../../types/posts';
 import type {GenericAction} from '../../types/actions';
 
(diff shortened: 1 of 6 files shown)


src/reducers/entities/emojis.js
```diff
@@ -81,6 +81,20 @@ function nonExistentEmoji(state = new Set(), action) {
         }
         return state;
     }
+    case EmojiTypes.LOAD_NONEXISTENT_EMOJIS: {
```
jwilander: Can we match the other type's name, so update to:
```suggestion
    case EmojiTypes. CUSTOM_EMOJIS_DO_NOT_EXIST: {
```

saturninoabril: Sure, now updated, thanks

jwilander:
We'd also want to skip the requests to the server for unknown emojis right? Which would still require some changes here. As for just ignoring nonExistentEmoji when post metadata is available and adding it to the client config, I'm OK with that.

saturninoabril:
I agree. I think if totally ignore nonExistentEmoji for enabled post metadata, then we'll effectively skip the network request for those unknown emoji.  There might be no change needed here, but I'll check.
Will submit PR to server and webapp (and redux if needed).

hmhealey:
Hopefully it's just the server and web app for that one

PR #732, opened 2018-12-06
MM-8662: Added `force` flag to uploadPlugin methods

levb:
#### Summary
Added a `force` flag to uploadPlugin action and client methods to facilitate in-place updating of plugins.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-8662

#### Checklist
- [ ] Added or updated unit tests (required for all new features) - NEED TO FIGURE OUT
- [ ] All new/modified APIs include changes to the drivers - NOT SURE WHAT THIS MEANS

#### Test Information
This PR was tested on: Chrome, Mac OS X

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -2419,13 +2419,14 @@ export default class Client4 {
         this.trackEvent('api', 'api_plugin_upload');
 
         const formData = new FormData();
-        formData.append('plugin', fileData);
         if (force) {
             formData.append('force', 'true');
         }
+        formData.append('plugin', fileData);
 
         const request = {
             method: 'post',
+
             body: formData,
         };
 
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -2426,7 +2426,6 @@ export default class Client4 {
 
         const request = {
             method: 'post',
-
             body: formData,
         };
 
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -2415,11 +2415,14 @@ export default class Client4 {
 
     // Plugin Routes - EXPERIMENTAL - SUBJECT TO CHANGE
 
-    uploadPlugin = async (fileData) => {
+    uploadPlugin = async (fileData, force = false) => {
         this.trackEvent('api', 'api_plugin_upload');
 
         const formData = new FormData();
         formData.append('plugin', fileData);
+        if (force) {
+            formData.append('force', 'true');
+        }
 
         const request = {
             method: 'post',
--- a/test/actions/admin.test.js
+++ b/test/actions/admin.test.js
@@ -864,17 +864,21 @@ describe('Actions.Admin', () => {
             return;
         }
 
-        const data1 = fs.createReadStream('test/assets/images/test.png');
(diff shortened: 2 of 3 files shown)


test/actions/admin.test.js
```diff
@@ -871,7 +871,11 @@ describe('Actions.Admin', () => {
             post('/plugins').
             reply(200, testPlugin);
 
-        await Actions.uploadPlugin(testFileData)(store.dispatch, store.getState);
+        await Actions.uploadPlugin(testFileData, false)(store.dispatch, store.getState);
+
+        // TODO: not sure how to add a test for Actions.uploadPlugin(testFileData, true)
```
lieut-data: Naively myself, I'd duplicate this test case, call it `uploadPlugin, forced`, and invoke with `, true`, but configuring the `nock` mock to expect `post('/plugins', '?force=true')`.

levb: I did now, not sure how meaningful this test is, but probably better than nothing. The flag is passed in the (multipart) form, so the POST URL is the same.

test/actions/admin.test.js
```diff
@@ -858,6 +858,37 @@ describe('Actions.Admin', () => {
...
+        const data2 = fs.createReadStream('test/assets/images/test.png');
+        const testPlugin = {id: 'testplugin', webapp: {bundle_path: '/static/somebundle.js'}};
+
+        nock(Client4.getBaseRoute()).
+            post('/plugins').
+            reply(200, testPlugin);
+        await Actions.uploadPlugin(data1, false)(store.dispatch, store.getState);
+
+        nock(Client4.getBaseRoute()).
+            post('/plugins').
```
lieut-data: While the URL does indeed stay the same, I think the second parameter to `post` accepts the body of the request, and can be used to verify the `force` flag as such. It might be tricky to build the body given the stream, but you can also pass a function to verify, e.g. as per example on https://www.npmjs.com/package/nock#specifying-request-body:
```js
nock('http://www.example.com')
  .post('/login', function(body) {
    return body.username && body.password;
  })
  .reply(200, { id: '123ABC' });
```

This might make the test a little more "useful" (test utility is marginal whenever mocks are involved, but /shrug).

lieut-data: @levb, I saw you re-requested a review, but I'm not seeing any changes here: do you think it's worth adding the extra validation to the mock?

levb: @lieut-data: my apologies, the changes didn't get added and I missed it. Should be there now. Thanks for the suggestion!

src/client/client4.js
```diff
@@ -2419,13 +2419,14 @@ export default class Client4 {
...
         const formData = new FormData();
-        formData.append('plugin', fileData);
         if (force) {
             formData.append('force', 'true');
         }
+        formData.append('plugin', fileData);
 
         const request = {
             method: 'post',
+
```
lieut-data: nit: intentional?

levb: Done

PR #742, opened 2018-12-18
[MM 13161] Update getJoinableTeams to return array of teams

migbot:
#### Summary
Since getJoinableTeams returned an object, the mobile app had to transform the response in mapStateToProps which lead to multiple re-renders when the redux state changed. This pull request includes a change to have getJoinableTeams return an array of teams which the mobile app can consume as is.

#### Ticket Link
[https://github.com/mattermost/mattermost-server/issues/9951](https://github.com/mattermost/mattermost-server/issues/9951)

#### Checklist

#### Test Information
This PR was tested on:
* Android emulator
* LG K20V running Android 7.0
* iOS emulator
* iPhone 8 running iOS 12.1

### DIFF
--- a/src/selectors/entities/teams.js
+++ b/src/selectors/entities/teams.js
@@ -158,7 +158,7 @@ export const getJoinableTeams = createSelector(
             openTeams[id] = teams[id];
         }
 
-        return openTeams;
+        return Object.values(openTeams);
     }
 );
 
--- a/src/selectors/entities/teams.js
+++ b/src/selectors/entities/teams.js
@@ -152,13 +152,7 @@ export const getJoinableTeams = createSelector(
     getTeams,
     getJoinableTeamIds,
     (teams, joinableTeamIds) => {
-        const openTeams = {};
-
-        for (const id of joinableTeamIds) {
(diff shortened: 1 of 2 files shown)


enahum:
@migbot thinking more about this, would it make sense to remove this one then and use `getSortedJoinableTeams` in the mobile app as well?

src/selectors/entities/teams.js
```diff
@@ -158,7 +158,7 @@ export const getJoinableTeams = createSelector(
             openTeams[id] = teams[id];
         }
 
-        return openTeams;
+        return Object.values(openTeams);
```
hmhealey: Instead of constructing an object and then using `Object.values` on it, you could replace the body of this with a `.map` call to simplify it a bit and skip creating the intermediate object
```
(teams, joinableTeamIds) -> {
    return joinableTeamIds.map((id) => teams[id]);
}
```

migbot: Yes, much cleaner! I've made the change.

hmhealey:
@enahum Just to confirm, is this safe to merge? If so, it's good to merge whenever

PR #747, opened 2018-12-23
[GH-7494] Added the filters in selectors and get users request

pradeepmurugesan:
#### Summary
Adds the filtering ability to the redux selectors

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/7494

#### Checklist

#### Test Information
This PR was tested on: [mac, ubuntu 18]

### DIFF
--- a/src/selectors/entities/users.js
+++ b/src/selectors/entities/users.js
@@ -7,8 +7,8 @@ import {
     getCurrentChannelId,
     getCurrentUser,
     getCurrentUserId,
-    getUsers,
     getMyCurrentChannelMembership,
+    getUsers,
 } from 'selectors/entities/common';
 
 import {getConfig, getLicense} from 'selectors/entities/general';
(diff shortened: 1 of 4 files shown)


src/selectors/entities/users.js
```diff
@@ -212,11 +212,32 @@ function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
...
+        if (filters.role && filters.role !== '') {
+            if (profile.roles && profile.roles.indexOf(filters.role) !== -1) {
+                filteredProfiles[id] = profile;
+            }
+        } else if (filters.inactive) {
+            if (profile.delete_at !== 0) {
+                filteredProfiles[id] = profile;
+            }
+        }
+    }
```
jwilander: This looks like it will only work for a single filter at a time. E.g. if I supply a filters of `{role: 'system_admin', inactive: true}` I won't get inactive sysadmins, I'll just get sysadmins. By default I'd expect that both filters I provide work

pradeepmurugesan: I did it this way keeping in my mind the UI in the ticket which didn't allow multi select. 

But looking it as a generic selector method it makes sense to support multiple filters here, as the consumers of this selector will expect it. Will change now.. 👍

PR #748, opened 2018-12-26
MM-13432 Remove unused _REQUEST, _SUCCESS and _FAILURE actions

sudheerDev:
#### Summary
This PR removes any unused status states and requests from the following entities

Channels
Posts
teams
Users
Files
Emoji

This is part of the performance improvement for channel switch/ postList mount and update. PR does not completely remove all of the request entities as most of the flags left does not impact the performance for the channel switch. Moreover moving away from storing flags might not be al that possible and needed in few cases so, we might have to circle back again to a similar pattern. Moving the flags to react components can cause problems as state is lost when a component is unmounted.

Going forward we have to make sure that we don't introduce these state flags for request if we don't use

#### Ticket Link
[MM-13544](https://mattermost.atlassian.net/browse/MM-13544)
related mobile PR: https://github.com/mattermost/mattermost-mobile/pull/2490

#### Checklist


src/actions/helpers.js
```diff
@@ -62,29 +62,37 @@ export function requestFailure(type: ActionType, error: Client4Error) {
  * @param {...Array<any>} args
  * @returns {ActionFunc} ActionFunc
  */
-export function bindClientFunc(clientFunc: () => Promise<mixed>, request: ActionType,
-    success: ActionType | Array<ActionType>, failure: ActionType, ...args: Array<any>): ActionFunc {
+export function bindClientFunc({
```
lieut-data: Should we update the `@param` comments here? Do we need to capture any flow types?

sudheerDev: Oh yes, I will update the params here. I need to push another commit as i saw few unused flags in mobile repo. So will be removing them and updating this PR along with it.

sudheerDev: @lieut-data done 👍 Have a look at the flow commit changes. Let me know if you have any comments

jwilander:
@sudheerDev is this ready for review?

sudheerDev:
@jwilander Forgot to change the label back. It is up for review

sudheerDev:
Yes both mobile and webapp do still use few flags. 
This PR removes the flags from few files which effect the perf and not at all used in both repos. 

There is a related mobile PR  in i removed few unused status flags so, i removed them as well and removed them from redux files in the last commit here.

Rest of the flags left here in the PR for the files above are being used and luckily they are not all that frequent actions.

src/actions/helpers.js
sudheerDev: Aah forgot to change the comment of @param here :(

PR #752, opened 2019-01-07
MM-13627 Fix DM and GM mention counts missing from browser tab

sudheerDev:
#### Summary
  * Update unreadCount selector to take into account other teams count values

#### Ticket Link
[MM-13627](https://mattermost.atlassian.net/browse/MM-13627)

#### Checklist

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -435,7 +435,7 @@ export const getUnreads: (GlobalState) => {messageCount: number, mentionCount: n
     getCurrentTeamId,
     getMyTeams,
     getTeamMemberships,
-    (channels: IDMappedObjects<Channel>, myMembers: RelationOneToOne<Channel, ChannelMembership>, users: IDMappedObjects<UserProfile>, currentUserId: string, currentTeamId: string, myTeams: Array<Team>, myTeamMemberships: Array<TeamMembership>): {messageCount: number, mentionCount: number} => {
+    (channels: IDMappedObjects<Channel>, myMembers: RelationOneToOne<Channel, ChannelMembership>, users: IDMappedObjects<UserProfile>, currentUserId: string, currentTeamId: string, myTeams: Array<Team>, myTeamMemberships: RelationOneToOne<Team, TeamMembership>): {messageCount: number, mentionCount: number} => {
         let messageCountForCurrentTeam = 0; // Includes message count from channels of current team plus all GM'S and all DM's across teams
         let mentionCountForCurrentTeam = 0; // Includes mention count from channels of current team plus all GM'S and all DM's across teams
         Object.keys(myMembers).forEach((channelId) => {
@@ -443,7 +443,7 @@ export const getUnreads: (GlobalState) => {messageCount: number, mentionCount: n
             const m = myMembers[channelId];
 
             if (channel && m && (channel.team_id === currentTeamId || channel.type === General.DM_CHANNEL || channel.type === General.GM_CHANNEL)) {
-                let otherUserId;
+                let otherUserId = '';
                 if (channel.type === 'D') {
                     otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
                     if (users[otherUserId] && users[otherUserId].delete_at === 0) {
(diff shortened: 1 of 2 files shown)


src/selectors/entities/channels.js
```diff
@@ -428,24 +431,42 @@ export const getUnreads = createSelector(
                 if (channel.type === 'D') {
                     otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
                     if (users[otherUserId] && users[otherUserId].delete_at === 0) {
```
sudheerDev: We do a lot of calc here in this selector to get the values of count based on archived channels and deleted channel. So far it works fine because we have the information of the archived channels for the current team.

Now here with this change we are trying to add count from all teams which might have count from archived channels which can pose an issue with unread count being out of sync. This probably will be considered edge case but i want to point that out as to think of a better solution around this.

sudheerDev:
@hmhealey 

Previously the tab title used to use this selector from redux which takes into account present team unreads alone.

When making the enhancement for getting unread counts of all teams on tab title we made a selector in webapp(Refer the removed code in the above webapp PR) where we were iterating the memberships for count which would exclude DM's and GM's hence the present issue.

With this PR wen are taking into account the present team counts along with DM's and GM's and other teams channel's count from memberships

> I'm pretty sure if you have the channels for two teams loaded, this will make it count unreads for channels the not current team twice.

Aah thats true and it is a bug for a long time because it counts based on the loaded channel. 

let me fix the loaded channel case

sudheerDev:
@hmhealey @saturninoabril No matter what we will have issues with archived channels here because we don't know the channels that are archived in other teams. is that an acceptable edge case to go ahead for the release?

should we be reverting https://mattermost.atlassian.net/browse/MM-13224 for the release and figure out a better way for deleted users and archived channels?

src/selectors/entities/channels.js
```diff
@@ -416,36 +416,57 @@ export const getUnreads = createSelector(
...
+    getTeamMemberships,
+    (channels, myMembers, users, currentUserId, currentTeamId, myTeams, myTeamMemberships) => {
+        let messageCountForCurrentTeam = 0; // Includes message count from channels of current team plus all GM'S and all DM's across teams
+        let mentionCountForCurrentTeam = 0; // Includes mention count from channels of current team plus all GM'S and all DM's across teams
         Object.keys(myMembers).forEach((channelId) => {
             const channel = channels[channelId];
             const m = myMembers[channelId];
 
-            if (channel && m) {
+            if (channel && m && (channel.team_id === currentTeamId || channel.type === 'D')) {
```
hmhealey: Does this check need to also include group channels as well?

Also, maybe we could remove `channel.team_id === currentTeamId` and just get the unread count for the current team the same way that we do for the other teams in the loop below? I'm not sure if we need to treat the current team differently than the other teams.

sudheerDev: we have all those checks for removing archived channels and deleted users unread count to be in sync with what is in LHS. 

If we remove that and do the same for all teams then we will end up with regressions as it will cause unreads to go out of sync with teams with archived channels or unread users

we cannot fix the case for archived channels unread count for other teams here and the best way to fix it would be by doing it server side and hence my question if we should revert the above PR in my comment so we don't need this PR.

sudheerDev: > Does this check need to also include group channels as well?

Aah i think we should have GM's as well here

sudheerDev: Tests fails. I will have a look at this tomorrow if i can do this in any better way.

hmhealey: What if you have an archived channel or deleted user DM on another team. Will that not cause the count to get messed up?

Since that's an edge case, I'd be fine taking this PR for now, but I want to make a ticket to at least test this more thoroughly. The tests for that selector are so closely tied to the other selector tests, so it's really hard to ensure this is tested properly in different cases.

sudheerDev: What if you have an archived channel or deleted user DM on another team. Will that not cause the count to get messed up?
Deleted users will work as they are independent of team but archived channels will cause issues
as i was sating before in other comments

> we cannot fix the case for archived channels unread count for other teams here and the best way to fix it would be by doing it server side

sudheerDev:
Will be talking to eric about the shortcomings later today before merging this.

PR #753, opened 2019-01-07
[MM-10346] CSRF Token Implementation

DSchalla:
#### Summary
This PR implements CSRF protection tokens for additional hardening compared to the currently used custom header. For additional explanation, please check the linked server PR below.

The way the CSRF cookie currently is read appears not idiomatic / ideal to me, I'd appreciate some feedback there in general, also how the initial parsing when the client is initiated / after the login is done is currently implemented. I'll update the tests afterwards accordingly, currently they fail because of the `document` usage.

#### Ticket Link
https://mattermost.atlassian.net/projects/MM/issues/MM-10346

#### Checklist

#### Test Information
This PR was tested on: Linux, Fedora 29, Google Chrome 71.0.3578.98

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -43,7 +43,6 @@ export default class Client4 {
             unknownError: 'We received an unexpected status code from the server.',
         };
 
-        this.setCSRFFromCookie();
     }
 
     getUrl() {
@@ -70,17 +69,8 @@ export default class Client4 {
         this.token = token;
     }
 
-    setCSRFFromCookie() {
-        if (typeof document !== 'undefined' && typeof document.cookie !== 'undefined') {
-            const cookies = document.cookie.split(';');
-            for (let i = 0; i < cookies.length; i++) {
-                const cookie = cookies[i].trim();
-                if (cookie.startsWith('MMCSRF=')) {
-                    this.csrf = cookie.replace('MMCSRF=', '');
-                    break;
-                }
-            }
-        }
+    setCSRF(csrfToken) {
+        this.csrf = csrfToken
     }
 
     setAcceptLanguage(locale) {
@@ -505,8 +495,6 @@ export default class Client4 {
             {method: 'post', body: JSON.stringify(body)}
         );
 
-        this.setCSRFFromCookie();
-
         return data;
     };
 
@@ -524,8 +512,6 @@ export default class Client4 {
             `${this.getUsersRoute()}/login`,
             {method: 'post', body: JSON.stringify(body)}
         );
-
-        this.setCSRFFromCookie();
 
         return data;
     };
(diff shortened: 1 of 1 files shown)


src/client/client4.js
```diff
@@ -66,6 +70,17 @@ export default class Client4 {
         this.token = token;
     }
 
+    setCSRFFromCookie() {
+        const cookies = document.cookie.split(';');
```
hmhealey: I don't know if we can assume we'll have the cookie here depending on what environment this is running on. I don't know if this is needed outside of the web though, so maybe we just add a null check for `document.cookie` before doing this?

DSchalla: Good call, yeah. On other environments token-based authentication is used, I suppose? Is the reason the tests fail with `document.cookie` undefined because the direct access to `document.cookie` isn't supposed to happen with Redux or is that rather a mocking problem, since it works fine in the application itself?

hmhealey: I haven't thought too much about whether or not we just shouldn't use `document.cookie` in this code. If we're really trying to make this code platform-agnostic, we wouldn't access it at all since it only exists on the browser and not on mobile/node.js, but I don't think it's worth that extra effort when we could just null check it. As for the tests, that's just an example of an environment without `document.cookie`, so while mocking it would fix the tests, it doesn't fix the root problem.

DSchalla: Absolutely, makes sense. Updated the code to check if `document` and `document.cookie` are defined.

src/client/client4.js
```diff
@@ -66,6 +70,19 @@ export default class Client4 {
         this.token = token;
     }
 
+    setCSRFFromCookie() {
```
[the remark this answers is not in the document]
hmhealey: As long as we're properly checking that the cookie exists before doing this, I think it's fine to have here since other people may need this for their own browser-based integrations as well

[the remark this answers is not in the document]
DSchalla: Not having a particular opinion here, reason why it was placed here was to parse the cookie directly from `loginById` - If you think we should implement a method ala `setCSRF` and call that in the webapp after the redux login method was called, that sounds good to me as well. Happy about any proposals. Thoughts @enahum @hmhealey @crspeller ?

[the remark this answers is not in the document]
enahum: I’m 0/5 after HH comment

DSchalla:
@hmhealey After a meeting today with @enahum we decided it makes sense to leave the set mechanism to the client itself, since aside from the webapp there are not many cases where you got access to document.cookie itself, and the callback got triggered more often than necessary. 

A separate mobile PR will be necessary however, the webapp PR got already updated. This PR is save to merge, since it doesn't have any negative impact on current installations / tests.

PR #755, opened 2019-01-09
Migrate to flow selectors and utils for posts

jespino:
#### Summary
Migrate to flow selectors and utils for posts.

This time I'm extensively used the `$ID` utility function to be super explicit about which kind of parameter is passed. I would appreciate comments about this.

#### Checklist

### DIFF
--- a/src/selectors/entities/posts.js
+++ b/src/selectors/entities/posts.js
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// @flow
 
 import {createSelector} from 'reselect';
 
@@ -11,26 +12,34 @@ import {Posts, Preferences} from 'constants';
 import {isPostEphemeral, isSystemMessage, shouldFilterJoinLeavePost, comparePosts, isPostPendingOrFailed, isPostCommentMention} from 'utils/post_utils';
 import {getPreferenceKey} from 'utils/preference_utils';
 
-export function getAllPosts(state) {
+import type {GlobalState} from 'types/store';
+import type {Post, PostWithFormatData} from 'types/posts';
+import type {Reaction} from 'types/reactions';
+import type {UserProfile} from 'types/users';
+import type {Channel} from 'types/channels';
+
+import type {$ID, IDMappedObjects, RelationOneToOne, RelationOneToMany} from '../../types/utilities';
+
+export function getAllPosts(state: GlobalState) {
     return state.entities.posts.posts;
 }
 
-export function getPost(state, postId) {
+export function getPost(state: GlobalState, postId: $ID<Post>): Post {
     return getAllPosts(state)[postId];
 }
 
-export function getPostsInThread(state) {
+export function getPostsInThread(state: GlobalState): RelationOneToMany<Post, Post> {
     return state.entities.posts.postsInThread;
 }
 
-export function getReactionsForPosts(state) {
+export function getReactionsForPosts(state: GlobalState): RelationOneToOne<Post, Array<Reaction>> {
     return state.entities.posts.reactions;
 }
 
-export function makeGetReactionsForPost() {
+export function makeGetReactionsForPost(): (GlobalState, $ID<Post>) => ?Array<Reaction> {
     return createSelector(
         getReactionsForPosts,
-        (state, postId) => postId,
+        (state: GlobalState, postId) => postId,
         (reactions, postId) => {
             if (reactions[postId]) {
                 return reactions[postId];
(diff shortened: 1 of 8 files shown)


src/selectors/entities/posts.js
```diff
@@ -260,12 +276,8 @@ export function makeGetPostsAroundPost() {
                     continue;
                 }
 
-                const previousPost = allPosts[slicedPostIds[i + 1]] || {create_at: 0};
-                const formattedPost = formatPostInChannel(post, previousPost, i, allPosts, postsInThread, slicedPostIds, currentUser);
-
-                if (post.id === focusedPostId) {
-                    formattedPost.highlight = true;
-                }
+                const previousPost = allPosts[slicedPostIds[i + 1]] || null;
```
cometkim: @jespino Could you explain logical change here?

jespino: Sure, the previous post is now A Post or null, not a Post or a custom object to fullfil some requirement elsewhere.

On the other hand, we have the `formatedPost`, generated using `formatPostIChannel` which now receive the `focusedPostId` variable to decide if the `formattedPost` is highlight or not.

src/types/users.js
```diff
@@ -56,3 +58,13 @@ export type UserTimezone = {|
     automaticTimezone: string,
     manualTimezone: string,
 |};
+
+export type UserActivity = {
+    [PostType]: {
+        [string]: {|
```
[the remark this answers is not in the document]
jespino: Good point.

src/types/posts.js
```diff
@@ -59,13 +60,32 @@ export type Post = {|
...
+    isCommentMention: boolean,
+    highlight: boolean,
+|};
+
 export type PostsState = {|
     posts: IDMappedObjects<Post>,
     postsInChannel: RelationOneToMany<Channel, Post>,
     postsInThread: RelationOneToMany<Post, Post>,
+    reactions: {[string]: Array<Reaction>},
+    openGraph: {[string]: Object},
```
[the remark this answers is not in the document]
jespino: Done

src/types/users.js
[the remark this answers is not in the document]
jespino: Done

src/selectors/entities/posts.js
hmhealey: Do we need to include `|| null` here? I don't know if the difference between null and undefined are meaningful in this code

src/selectors/entities/posts.js
```diff
@@ -278,11 +290,11 @@ export function makeGetPostsAroundPost() {
 // Returns a function that creates a creates a selector that will get the posts for a given thread.
 // That selector will take a props object (containing a rootId field) as its
 // only argument and will be memoized based on that argument.
-export function makeGetPostsForThread() {
+export function makeGetPostsForThread(): (GlobalState, {rootId: $ID<Post>}) => Array<Post> {
     return createSelector(
         getAllPosts,
-        (state, {rootId}) => state.entities.posts.postsInThread[rootId] || [],
-        (state, {rootId}) => state.entities.posts.posts[rootId],
+        (state: GlobalState, {rootId}) => state.entities.posts.postsInThread[rootId] || [],
```
hmhealey: Not something for this PR, but these selectors aren't memoized properly because of the array literal here. I'll file a ticket for that

src/utils/post_utils.js
```diff
@@ -192,24 +202,35 @@ export const postTypePriority = {
     [Posts.POST_TYPES.ADD_TO_CHANNEL]: 5,
     [Posts.POST_TYPES.LEAVE_CHANNEL]: 6,
     [Posts.POST_TYPES.REMOVE_FROM_CHANNEL]: 7,
+    [Posts.POST_TYPES.PURPOSE_CHANGE]: 8,
```
hmhealey: Why add these values? This array is only used for sorting posts as part of combined system messages, and the types you've added are not combined into those

src/selectors/entities/posts.js
jespino: Not really, but I normally prefer an explicit `null` than `undefined` which is normally implicit.

src/utils/post_utils.js
jespino: Is not needed as part of the "Code" itself, but is needed to be consistent and complete with the available post types. Basically is for make flow pass, other way is to define an specific type for "combinable post types", or something like that.

hmhealey: Alright, I guess this is fine. I'll likely take another pass through this anyway since I'll be rewriting part of that anyway

hanzei:
Hey @hmhealey, Could you give this a round of review, when you got the time?

PR #759, opened 2019-01-28
Migrate to flow selectors of users

jespino:
#### Summary
Migrate to flow selectors and utils for users.

#### Checklist

### DIFF
--- a/src/selectors/entities/users.js
+++ b/src/selectors/entities/users.js
@@ -237,7 +237,7 @@ export const getProfiles: (GlobalState, Filters) => Array<UserProfile> = createS
     }
 );
 
-function filterProfiles(profiles: IDMappedObjects<UserProfile>, filters: ?Filters): IDMappedObjects<UserProfile> {
+function filterProfiles(profiles: IDMappedObjects<UserProfile>, filters?: Filters): IDMappedObjects<UserProfile> {
     if (!filters) {
         return profiles;
     }
@@ -245,7 +245,7 @@ function filterProfiles(profiles: IDMappedObjects<UserProfile>, filters: ?Filter
     let users = Object.keys(profiles).map((key) => profiles[key]);
 
     if (filters.role && filters.role !== '') {
-        users = users.filter((user) => user.roles && user.roles.indexOf((filters && filters.role) || '') !== -1);
+        users = users.filter((user) => user.roles && user.roles.includes((filters && filters.role) || ''));
     }
 
     if (filters.inactive) {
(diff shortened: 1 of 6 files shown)


jespino:
I updated the flow version to 0.91.0. I'm not 100% sure if was part of the problem but I was experiencing certain problems with imports using the 0.89.0 version.

src/selectors/entities/users.js
```diff
@@ -215,23 +229,23 @@ function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
...
     getUsers,
     (state, filters) => filters,
     (profiles, filters) => {
         return sortAndInjectProfiles(filterProfiles(profiles, filters), PROFILE_SET_ALL);
     }
 );
 
-function filterProfiles(profiles, filters) {
-    if (!filters || Object.keys(filters).length === 0) {
+function filterProfiles(profiles: IDMappedObjects<UserProfile>, filters: ?Filters): IDMappedObjects<UserProfile> {
```
cometkim: `filters` seems like an optional parameter, so `filters?: Filters` instead of `filters: ?Filters`?

src/selectors/entities/users.js
```diff
@@ -215,23 +229,23 @@ function sortAndInjectProfiles(profiles, profileSet, skipInactive = false) {
...
+    if (!filters) {
         return profiles;
     }
 
-    let users = Object.values(profiles);
+    let users = Object.keys(profiles).map((key) => profiles[key]);
 
     if (filters.role && filters.role !== '') {
-        users = users.filter((user) => user.roles && user.roles.indexOf(filters.role) !== -1);
+        users = users.filter((user) => user.roles && user.roles.indexOf((filters && filters.role) || '') !== -1);
```
cometkim:
```suggestion
        users = users.filter((user) => user.roles?.includes(filters.role));
```

src/selectors/entities/users.js
```diff
@@ -295,25 +309,27 @@ export const getProfilesWithoutTeam = createSelector(
...
 
-export function getTotalUsersStats(state) {
+export function getTotalUsersStats(state: GlobalState): Object {
     return state.entities.users.stats;
 }
 
-export function searchProfiles(state, term, skipCurrent = false, filters = {}) {
-    const profiles = filterProfilesMatchingTerm(Object.values(getUsers(state)), term, filters);
-    const filteredProfiles = Object.values(filterProfiles(profiles, filters));
+export function searchProfiles(state: GlobalState, term: string, skipCurrent: boolean = false, filters: ?Filters = null): Array<UserProfile> {
```
cometkim: `filters` here also could be an optional parameter

jespino:
@cometkim PTAL. I added a new commit with the proposed changes, which aren't exactly equal, because there are some restrictions in flow (like chainging operator + function calls).

cometkim:
@jespino Sorry for the delay, new changes look good to me

PR #761, opened 2019-02-02
[MM-12737] Migrate preference-related actions from user_actions.jsx to redux

MirlanMaksv:
#### Summary
Moved preference related actions [savePreferences, savePreference, deletePreferences] from `mattermost-webapp/actions/user_actions.js` into `mattermost-redux/src/actions/preferences.js`.
Additionally updated `channels.js, preferences.js` to pass flow check.
[Webapp PR](https://github.com/mattermost/mattermost-webapp/pull/2320)

#### Ticket Link
[MM-12737](https://github.com/mattermost/mattermost-server/issues/10147)

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -118,7 +118,7 @@ export function createDirectChannel(userId: string, otherUserId: string): Action
             {user_id: userId, category: Preferences.CATEGORY_CHANNEL_OPEN_TIME, name: created.id, value: new Date().getTime().toString()},
         ];
 
-        savePreferences(userId, preferences)(dispatch);
+        savePreferences(userId, preferences)(dispatch, getState);
 
         dispatch(batchActions([
             {
@@ -686,7 +686,7 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
             const preferences = [
                 {user_id: currentUserId, category: Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME, name: channelId, value: new Date().getTime().toString()},
             ];
-            savePreferences(currentUserId, preferences)(dispatch);
+            savePreferences(currentUserId, preferences)(dispatch, getState);
         }
 
         try {
@@ -1230,7 +1230,7 @@ export function favoriteChannel(channelId: string): ActionFunc {
 
         Client4.trackEvent('action', 'action_channels_favorite');
 
-        return savePreferences(currentUserId, [preference])(dispatch);
+        return savePreferences(currentUserId, [preference])(dispatch, getState);
     };
 }
 
--- a/src/actions/preferences.js
+++ b/src/actions/preferences.js
@@ -68,7 +68,7 @@ export function makeDirectChannelVisibleIfNecessary(otherUserId: string): Action
                 value: 'true',
             };
             getProfilesByIds([otherUserId])(dispatch, getState);
-            savePreferences(currentUserId, [preference])(dispatch);
+            savePreferences(currentUserId, [preference])(dispatch, getState);
         }
 
         return {data: true};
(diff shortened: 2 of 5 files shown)


src/actions/channels.js
```diff
@@ -118,7 +118,7 @@ export function createDirectChannel(userId: string, otherUserId: string): Action
             {user_id: userId, category: Preferences.CATEGORY_CHANNEL_OPEN_TIME, name: created.id, value: new Date().getTime().toString()},
         ];
 
-        savePreferences(userId, preferences)(dispatch);
+        savePreferences(userId, preferences)(dispatch, getState);
```
enahum: This should be `dispatch(savePreferences(userId, preferences));` the same applies to the other places as well.

src/actions/preferences.js
```diff
@@ -99,32 +99,55 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
...
-                    },
-                },
-            },
-        });
+        dispatchPrefSave(dispatch, userId, preferences);
+        return {data: true};
+    };
+}
+
+export function saveCurrentUserPreferences(preferences: Array<PreferenceType>): ActionFunc {
```
enahum: Is this action needed here? This can be done directly in the webapp, and the call savePreferences

src/actions/preferences.js
```diff
@@ -99,32 +99,55 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
...
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const userId = getCurrentUserId(getState());
+
+        dispatchPrefSave(dispatch, userId, preferences);
+
+        return {data: true};
+    };
+}
+
+export function savePreference(category: string, name: string, value: string): ActionFunc {
```
enahum: What’s the reason for this action?

src/actions/preferences.js
```diff
@@ -99,32 +99,55 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
...
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const userId = getCurrentUserId(getState());
+
+        dispatchPrefSave(dispatch, userId, [{user_id: userId, category, name, value}]);
 
         return {data: true};
     };
 }
+
+function dispatchPrefSave(dispatch: DispatchFunc, userId: string, preferences: Array<PreferenceType>) {
```
enahum: This was already in savePreferences action, can’t follow why change it

src/actions/channels.js
MirlanMaksv: All changes in this file I did only to pass flow check, ok I'll change them to dispatch actions

src/actions/preferences.js
MirlanMaksv: Yes totally possible, but in that case components should know `userId`

src/actions/preferences.js
MirlanMaksv: I just moved it from `user_actions` file, it is used by some components

src/actions/preferences.js
enahum: Compinent do not need to know the userId, you can add an action in the webapp that gets the userId from the store and then calls the savePreferences action, no need to have it here

src/actions/preferences.js
MirlanMaksv: all actions that use this function dispatch the same action (with diff in some places) and that action structure was very huge, so I decided to put it into function instead of duplicating

src/actions/preferences.js
MirlanMaksv: For me it seems to be a bit declarative way, `saveCurrentUserPreferences()` you know that it updates current users prefs comparing to `savePreferences()`

enahum: Yes agree, but this action can live only in the webapp, no need to have it here

enahum: I’m just looking to have a clean code base and avoid client specific actions as much as possible.

enahum:
Monile already uses redux actions, desktop is the webapp running on a webview so we are good there.

Webapp was using flux before redux that is why we need the migrations

MirlanMaksv:
> Monile already uses redux actions, desktop is the webapp running on a webview so we are good there.
> 
> Webapp was using flux before redux that is why we need the migrations

got it, so we close this PR and I should update webapp PR?

enahum:
Yes we can close this PR

PR #763, opened 2019-02-04
Match webapp websocket actions

enahum:
#### Summary
The WebSocket actions between the webapp and the ones in this library are out of sync and I found that the webapp behaves a bit different, this PR will match the actions for:
* handleNew Post
* reconnecting the websocket 
* marking the channel as read

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13896

#### Checklist

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1158,7 +1158,7 @@ export function handleNewPost(msg) {
         const state = getState();
         const currentUserId = getCurrentUserId(state);
         const post = JSON.parse(msg.data.post);
-        const myChannelMember = getMyChannelMemberSelector(getState(), post.channel_id);
+        const myChannelMember = getMyChannelMemberSelector(state, post.channel_id);
         let websocketMessageProps = {};
         if (msg) {
             websocketMessageProps = msg.data;
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1159,10 +1159,7 @@ export function handleNewPost(msg) {
         const currentUserId = getCurrentUserId(state);
         const post = JSON.parse(msg.data.post);
         const myChannelMember = getMyChannelMemberSelector(state, post.channel_id);
-        let websocketMessageProps = {};
-        if (msg) {
-            websocketMessageProps = msg.data;
-        }
+        const websocketMessageProps = msg.data;
 
         if (myChannelMember && Object.keys(myChannelMember).length === 0 && myChannelMember.constructor === 'Object') {
             await dispatch(getMyChannelMember(post.channel_id));
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1187,12 +1187,7 @@ function completePostReceive(post, websocketMessageProps) {
         const rootPost = Selectors.getPost(state, post.root_id);
 
         if (post.root_id && !rootPost) {
-            const {data: posts} = await dispatch(getPostThread(post.root_id));
-            if (posts) {
-                dispatch(lastPostActions(post, websocketMessageProps));
-            }
-
-            return;
+            dispatch(getPostThread(post.root_id));
         }
 
         dispatch(lastPostActions(post, websocketMessageProps));
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -4,21 +4,31 @@
 import {batchActions} from 'redux-batched-actions';
 
 import {Client4} from 'client';
-import {General, Preferences, Posts} from 'constants';
+import {General, Preferences, Posts, WebsocketEvents} from 'constants';
 import {PostTypes, FileTypes, IntegrationTypes} from 'action_types';
-import {getUsersByUsername} from 'selectors/entities/users';
+
+import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} from 'selectors/entities/channels';
 import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
-
-import * as Selectors from 'selectors/entities/posts';
 import {getConfig} from 'selectors/entities/general';
+import * as Selectors from 'selectors/entities/posts';
+import {getCurrentUserId, getUsersByUsername} from 'selectors/entities/users';
 
+import {getUserIdFromChannelName} from 'utils/channel_utils';
 import {parseNeededCustomEmojisFromText} from 'utils/emoji_utils';
+import {isFromWebhook, isSystemMessage, shouldIgnorePost} from 'utils/post_utils';
 
-import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
-import {logError} from './errors';
-import {deletePreferences, savePreferences} from './preferences';
-import {getProfilesByIds, getProfilesByUsernames, getStatusesByIds} from './users';
+import {getMyChannelMember, markChannelAsUnread, markChannelAsRead, markChannelAsViewed} from './channels';
 import {systemEmojis, getCustomEmojiByName, getCustomEmojisByName} from './emojis';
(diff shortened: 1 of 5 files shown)


lieut-data:
@enahum, this seems to require a bit more domain knowledge on my part to review effectively. Can you ping me when you have ~10 minutes and we can sync up?

enahum:
@lieut-data what about right after the dev meeting?

lieut-data:
Sounds good, @enahum!

src/actions/websocket.js
```diff
@@ -132,37 +115,32 @@ function doReconnect() {
...
-                    if (!stillMemberOfCurrentChannel) {
-                        const defaultChannel = channels.find((c) => c.team_id === currentTeamId && c.name === General.DEFAULT_CHANNEL);
-                        if (defaultChannel) {
-                            EventEmitter.emit(General.DEFAULT_CHANNEL, defaultChannel.display_name);
-                            selectChannel(defaultChannel.id)(dispatch, getState);
-                        }
-                    }
+            dispatch(getPosts(currentChannelId)).then(({error}) => {
+                // Getting the posts return the user is not authorized, redirecting to the default channel
+                if (error && error.status_code === 403) {
```
hmhealey: I don't have any specific suggestions to improve this, but something about "try getting the posts and if we get a 403, we've been removed from the channel" feels hacky. I'd prefer something more straightforward like "get the latest channels and if we're not in the current channel, we've been removed from the channel".

src/actions/websocket.js
```diff
@@ -132,37 +115,32 @@ function doReconnect() {
...
+                    EventEmitter.emit(General.DEFAULT_CHANNEL, currentTeamId);
```
hmhealey: If it intentional that you're changing the second argument here from the Town Square's display name to the current team's ID?

Also, I know this is how it worked before, but I feel like we should be emitting an event-specific constant instead of just "town-square" here

src/actions/websocket.js
```diff
@@ -172,7 +150,7 @@ function doReconnect() {
                         team_id: currentTeamId,
                     },
                 };
-                handleLeaveTeamEvent(newMsg, dispatch, getState);
+                dispatch(handleLeaveTeamEvent(newMsg));
```
hmhealey: Funny how the old version of this code didn't actually do anything before

src/actions/websocket.js
```diff
@@ -132,37 +115,32 @@ function doReconnect() {
         const currentChannelId = getCurrentChannelId(state);
         const currentUserId = getCurrentUserId(state);
 
-        // We try and get the posts for the current channel as soon as possible
-        if (currentChannelId) {
-            dispatch(markChannelAsRead(currentChannelId));
-            dispatch(loadPostsHelper(currentChannelId));
-        }
+        if (currentTeamId) {
+            const dmPrefs = getDirectShowPreferences(state);
```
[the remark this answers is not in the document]
enahum: I'll group this

src/actions/websocket.js
enahum: agree, I'll change it.

src/actions/websocket.js
enahum: Yes `teamId` is what we want. I've changed the event to be `SWITCH_TO_DEFAULT_CHANNEL`

src/actions/websocket.js
```diff
@@ -311,113 +289,17 @@ function handleEvent(msg) {
...
-                    order: [],
-                    posts: {
-                        [post.id]: post,
-                    },
-                },
-                channelId: post.channel_id,
+        if (post.user_id !== getCurrentUserId(getState()) && !fromAutoResponder(post)) {
+            dispatch({
+                type: UserTypes.RECEIVED_STATUSES,
+                data: [{user_id: post.user_id, status: General.ONLINE}],
```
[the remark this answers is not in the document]
enahum: This is following what we have on the webapp, should we remove it?

src/actions/websocket.js
```diff
@@ -311,113 +289,17 @@ function handleEvent(msg) {
...
 
-        // getProfilesAndStatusesForPosts only gets the status if it doesn't exist, but we
-        // also want it if the user does not appear to be online
-        if (userId !== currentUserId && status && status !== General.ONLINE) {
-            dispatch(getStatusesByIds([userId]));
-        }
-
-        switch (post.type) {
-        case Posts.POST_TYPES.HEADER_CHANGE:
-            dispatch(updateChannelHeader(post.channel_id, post.props.new_header));
```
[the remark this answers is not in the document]
enahum: don't know if this is being used somewhere else

src/actions/websocket.js
```diff
@@ -311,113 +289,17 @@ function handleEvent(msg) {
...
-            break;
-        case Posts.POST_TYPES.PURPOSE_CHANGE:
-            dispatch(updateChannelPurpose(post.channel_id, post.props.new_purpose));
```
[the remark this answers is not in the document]
enahum: same here

src/actions/websocket.js
```diff
@@ -537,11 +419,8 @@ function handleUserRemovedEvent(msg) {
             }
 
             if (msg.data.channel_id === currentChannelId) {
-                const defaultChannel = Object.values(channels).find((c) => c.team_id === currentTeamId && c.name === General.DEFAULT_CHANNEL);
```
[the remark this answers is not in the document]
enahum: the default channel is selected by the client

src/actions/posts.js
```diff
@@ -1143,6 +1153,97 @@ export function moveHistoryIndexForward(index) {
     };
 }
 
+export function handleNewPost(msg) {
+    return async (dispatch, getState) => {
+        const state = getState();
+        const currentUserId = getCurrentUserId(state);
+        const post = JSON.parse(msg.data.post);
+        const myChannelMember = getMyChannelMemberSelector(getState(), post.channel_id);
```
jespino: `state` instead of `getState()`

src/actions/posts.js
```diff
@@ -1143,6 +1153,97 @@ export function moveHistoryIndexForward(index) {
...
 }
 
+export function handleNewPost(msg) {
+    return async (dispatch, getState) => {
+        const state = getState();
+        const currentUserId = getCurrentUserId(state);
+        const post = JSON.parse(msg.data.post);
+        const myChannelMember = getMyChannelMemberSelector(state, post.channel_id);
+        let websocketMessageProps = {};
+        if (msg) {
```
lieut-data: This check seems redundant: if `msg` is falsy, then we'll have excepted on line 1160 trying to access `msg.data.post`. Or should we be checking this earlier?

enahum: god catch, that's what you get when doing a blind copy/paste

enahum: updated

src/actions/posts.js
```diff
@@ -1143,6 +1153,94 @@ export function moveHistoryIndexForward(index) {
...
+        const myChannelMember = getMyChannelMemberSelector(state, post.channel_id);
+        const websocketMessageProps = msg.data;
+
+        if (myChannelMember && Object.keys(myChannelMember).length === 0 && myChannelMember.constructor === 'Object') {
+            await dispatch(getMyChannelMember(post.channel_id));
+        }
+
+        dispatch(completePostReceive(post, websocketMessageProps));
+
+        if (msg && msg.data) {
```
lieut-data: Worth doing the same here?

enahum: done

hmhealey:
Assigning this to Elias for merge since cherry-picking redux changes is difficult 😛

PR #735, opened 2018-12-10
MM-13188 Fix for channels with unreads having loading indicator placed above unread messages line

sudheerDev:
#### Summary

  * Add empty postIds array if posts are empty
    This is to have a difference on the store wrt channels with empty posts and
    channels user haven't visited.

  * Do not add postId to postsInChannel for channel is not initiated
  * add posts to posts object as we can be using them for pinned posts search etc

Changes include a ticket from RN as well as it depends on mm-redux socket actions

#### Ticket Link
[MM-13188](https://mattermost.atlassian.net/browse/MM-13188)
[MM-13428](https://mattermost.atlassian.net/browse/MM-13428)

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Ran `make check-style` to check for style errors (required for all pull requests)
- [x] Ran `make test` to ensure unit tests passed
- [x] Ran `make flow` to ensure type checking passed
- [x] Added or updated unit tests (required for all new features)

#### Test Information
This PR was tested on: [Device name(s), OS version(s)]

PR #767, opened 2019-02-06
MM-13957 Reorganize post actions 

hmhealey:
This PR is based on top of #762, so until that is merged, you can see changes specific to this PR here: https://github.com/mattermost/mattermost-redux/compare/mm11887-1-split-reducers...mm11887-2-replace-actions

This is the second preparation step for redoing the postsInChannel store. I went through and audited all of the post actions to ensure they behave in a way that will work with the changes to postsInChannel. I've made the following changes to the post actions:
1. RECEIVED_POST and RECEIVED_POSTS are only used to update existing posts now. The only effect those actions have on postsInChannel is to replace pending posts once they've been created.
2. RECEIVED_POSTS_IN_CHANNEL, RECEIVED_POSTS_IN_THREAD, and RECEIVED_POSTS_SINCE are new actions specific to their respective APIs because each of those actions need to be handled differently by postsInChannel due to how they return their data.
3. Those actions along with RECEIVED_SEARCH_POSTS must now be dispatched along with a RECEIVED_POSTS to update the other reducers. I wasn't planning on that originally, but this is easier than having each reducer that previously listened for a RECEIVED_POSTS action now have to listen for 5+ different actions.
4. RECEIVED_NEW_POST is now used whenever a new post is received from any user. Previously, it was only used when the current user made a pending post.
5. REMOVE_POST has been renamed to POST_REMOVED to be consistent with POST_DELETED.
6. REMOVE_PENDING_POST has been removed in favour of POST_REMOVED

Some other changes I made at the same time are:
1. I added simpler action creators like `postReceived`, `postDeleted`, etc so that we don't need to remember the exact structure of each of those actions ([link](https://github.com/mattermost/mattermost-redux/compare/mm11887-1-split-reducers...mm11887-2-replace-actions#diff-e32de3c64c2ee6b0ec20a2b9711d903fR23))
2. I removed some actions like `getPostThreadWithRetry` that weren't actually used anywhere
3. I removed the `sendingPostIds` reducer since it was almost identical to the `pendingPostIds` one, and it was only used in one place ([link](https://github.com/mattermost/mattermost-redux/compare/mm11887-1-split-reducers...mm11887-2-replace-actions#diff-a7ab98643e1a996aa39d1eec2f1a296bL485))

There will also be web app and mobile PRs coming to accompany this

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13957

#### Checklist

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -20,6 +20,66 @@ import {deletePreferences, savePreferences} from './preferences';
 import {getProfilesByIds, getProfilesByUsernames, getStatusesByIds} from './users';
 import {systemEmojis, getCustomEmojiByName, getCustomEmojisByName} from './emojis';
 
+export function receivedPost(post) {
+    return {
+        type: PostTypes.RECEIVED_POST,
+        data: post,
+    };
+}
+
+export function receivedNewPost(post) {
+    return {
+        type: PostTypes.RECEIVED_NEW_POST,
+        data: post,
+    };
+}
+
+export function receivedPosts(posts) {
+    return {
+        type: PostTypes.RECEIVED_POSTS,
+        data: posts,
+    };
+}
+
+export function receivedPostsSince(posts, channelId, since) {
+    return {
+        type: PostTypes.RECEIVED_POSTS_SINCE,
+        channelId,
+        data: posts,
+        since,
+    };
+}
+
+export function receivedPostsInChannel(posts, channelId) {
+    return {
+        type: PostTypes.RECEIVED_POSTS_IN_CHANNEL,
+        channelId,
+        data: posts,
+    };
+}
+
+export function receivedPostsInThread(posts, rootId) {
+    return {
+        type: PostTypes.RECEIVED_POSTS_IN_CHANNEL,
+        data: posts,
+        rootId,
+    };
+}
+
+export function postDeleted(post) {
+    return {
+        type: PostTypes.POST_DELETED,
+        data: post,
+    };
+}
+
+export function postRemoved(post) {
+    return {
+        type: PostTypes.POST_REMOVED,
+        data: post,
+    };
+}
+
 export function getPost(postId) {
     return async (dispatch, getState) => {
         let post;
@@ -37,10 +97,7 @@ export function getPost(postId) {
         }
 
         dispatch(batchActions([
-            {
-                type: PostTypes.RECEIVED_POST,
-                data: post,
-            },
+            receivedPost(post),
             {
                 type: PostTypes.GET_POSTS_SUCCESS,
             },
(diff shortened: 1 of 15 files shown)


src/actions/posts.js
```diff
@@ -519,75 +537,18 @@ export function getPostThread(postId, skipAddToChannel = true) {
...
-                channelId: post.channel_id,
-                skipAddToChannel,
-            },
+            receivedPosts(posts),
+            receivedPostsInThread(posts, rootId),
             {
                 type: PostTypes.GET_POST_THREAD_SUCCESS,
             },
-        ], 'BATCH_GET_POST_THREAD'), getState);
+        ]), getState);
```
enahum: `getState` is not needed here

src/actions/websocket.js
```diff
@@ -381,16 +381,7 @@ function handleNewPostEvent(msg) {
...
-                type: PostTypes.RECEIVED_POSTS,
-                data: {
-                    order: [],
-                    posts: {
-                        [post.id]: post,
-                    },
-                },
-                channelId: post.channel_id,
-            });
+            actions.push(receivedNewPost(post));
```
enahum: nothing to do here, but this will conflict with #763

hmhealey: Yeah, this PR also conflicts with https://github.com/mattermost/mattermost-redux/pull/735 as well. This is on the branch, so we can deal with those once this branch is updated

src/actions/posts.js
hmhealey: Once we're all unblocked with this redux stuff, I'm going to look at doing a mass refactor to get rid of this once and for all

src/actions/posts.js
```diff
@@ -30,6 +30,66 @@ import {
...
+    };
+}
+
+export function postRemoved(post) {
+    return {
+        type: PostTypes.POST_REMOVED,
+        data: post,
+    };
+}
+
```
sudheerDev: Good call making the above dispatches into separate functions 👍

src/reducers/entities/search.js
```diff
@@ -13,7 +13,7 @@ function results(state = [], action) {
         }
         return action.data.order;
     }
-    case PostTypes.REMOVE_POST: {
+    case PostTypes.POST_REMOVED: {
```
sudheerDev: This would require us to make changes to webapp as well as we are using them `PostTypes.REMOVE_POST` action type

src/actions/posts.js
sudheerDev: This would require us to make changes to webapp as well as we are using them `PostTypes.REMOVE_POST` action type

sudheerDev:
@hmhealey PR looks good to me. Other than the one action type change which needs update in webapp. I am ok merging this as long as we track and make a PR for that. Your call

src/actions/posts.js
hmhealey: Yeah, I'm liking this idea. It also makes it easier to see what the arguments should be without having to find the reducers

src/reducers/entities/search.js
hmhealey: Already done :)

https://github.com/mattermost/mattermost-webapp/pull/2343/files#diff-352a5ff733bbce75800e45b15f77ee14R305

hmhealey:
Yeah, there's already accompanying mobile and webapp PRs for this. They use the new actions and clean up some surrounding logic

PR #769, opened 2019-02-07
Added doPostActionWithCookie for MM-10516, actions in ephemeral messages

levb:
#### Summary
Added new functions (action, and client method) augmenting doPostAction to
include the "cookie" value. The old functions should no longer be used,
but I wasn't sure if it's safe to deprecate them, thus new additions.

An upcoming mattermost-server PR will have a better write-up on the use-case and the overall flow.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10516

#### Checklist
- [X] Ran `make check-style` to check for style errors (required for all pull requests)
- [X] Ran `make test` to ensure unit tests passed
- [X] Ran `make flow` to ensure type checking passed
- [X] Added or updated unit tests (required for all new features)

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1079,25 +1079,7 @@ export function getOpenGraphMetadata(url) {
 }
 
 export function doPostAction(postId, actionId, selectedOption = '') {
-    return async (dispatch, getState) => {
-        let data;
-        try {
-            data = await Client4.doPostAction(postId, actionId, selectedOption);
-        } catch (error) {
-            forceLogoutIfNecessary(error, dispatch, getState);
-            dispatch(logError(error));
-            return {error};
-        }
-
-        if (data && data.trigger_id) {
-            dispatch({
-                type: IntegrationTypes.RECEIVED_DIALOG_TRIGGER_ID,
-                data: data.trigger_id,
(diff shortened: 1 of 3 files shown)


src/actions/posts.js
```diff
@@ -1095,6 +1095,29 @@ export function doPostAction(postId, actionId, selectedOption = '') {
                 data: data.trigger_id,
             });
         }
+
+        return {data};
+    };
+}
+
+export function doPostActionWithCookie(postId, actionId, actionCookie, selectedOption = '') {
```
jespino: Maybe we can rewrite `doPostAction` to use `doPostActionWithCookie(postId, actionId, null, selectedOption)` or something like that.

src/client/client4.js
```diff
@@ -1589,6 +1589,23 @@ export default class Client4 {
         );
     };
 
+    doPostActionWithCookie = async (postId, actionId, actionCookie, selectedOption = '') => {
```
jespino: Probably we can do the same here rewrite `doPostAction` to simply call `doPostActionWithCookie` with the cookie value set to null, or something like that.

levb:
@crspeller done

PR #771, opened 2019-02-11
Replacing allow_open_invites with is_public

jespino:
#### Summary
Replacing `allow_open_invites` and `type` from `Team` struct with `is_public` and define or not `invite_id`.

#### Ticket Link
[MM-13931](https://mattermost.atlassian.net/browse/MM-13931)

#### Checklist

### DIFF
--- a/src/types/teams.js
+++ b/src/types/teams.js
@@ -13,8 +13,6 @@ export type TeamMembership = {|
     scheme_admin: boolean
 |};
 
-export type TeamType = 'O' | 'I';
-
 export type Team = {|
     id: string,
     create_at: number,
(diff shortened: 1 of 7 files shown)


src/types/teams.js
```diff
@@ -24,11 +22,11 @@ export type Team = {|
     name: string,
     description: string,
     email: string,
-    type: TeamType,
     company_name: string,
     allowed_domains: string,
     invite_id: string,
     allow_open_invite: boolean,
```
jespino: I keep allow_open_invite in the type definition because backward compatibility with older server versions. This is used in one place which is not typed for now, but will be in the future and will need the field.

jespino:
Added to the list of Tech Moonshots for the next major version.

PR #774, opened 2019-02-14
MM-14115 Do not clobber reactions when dispatching RECEIVED_POST that has no reactions metadata

jwilander:
#### Summary
When a post had reactions and was pinned, we would dispatch RECEIVED_POST with a post that had an empty metadata object which would be interpreted as the post not having reactions. The reactions would get clobbered and the post's reactions would stop rendering.

By checking to see if the reactions array is defined within the post's metadata, we only remove the reactions when they are actually removed from the post.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14115

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -606,7 +606,7 @@ export function reactions(state = {}, action) {
 }
 
 function storeReactionsForPost(state, post) {
-    if (!post.metadata) {
+    if (!post.metadata || !post.metadata.reactions) {
         return state;
     }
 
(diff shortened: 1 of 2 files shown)


src/reducers/entities/posts.js
```diff
@@ -606,7 +606,7 @@ export function reactions(state = {}, action) {
 }
 
 function storeReactionsForPost(state, post) {
-    if (!post.metadata) {
+    if (!post.metadata || !post.metadata.reactions) {
```
hmhealey: Does this work if you remove the last reaction from the post? I think we use `omitempty` for this field, so it might not update to remove the last reaction

jwilander: I'm pretty sure I tried that but I'll try again really quick

jwilander: Seems to work fine 👍

PR #776, opened 2019-02-17
[MM-12738] Migrate OAuth app actions from user_actions.jsx and admin_actions.jsx to redux

MirlanMaksv:
#### Summary
Part of [this issue](https://github.com/mattermost/mattermost-server/issues/10153). Moved 
- `allowOAuth2`
- `getOAuthAppInfo`
- `getAuthorizedApps` 
- `deauthorizeOAuthApp`

actions from webapp repo. 

#### Ticket Link
[MM-12738](https://github.com/mattermost/mattermost-server/issues/10153)

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/admin.js
+++ b/src/actions/admin.js
@@ -547,3 +547,23 @@ export function disablePlugin(pluginId: string): ActionFunc {
         return {data: true};
     };
 }
+
+export function getOAuthAppInfo(clientId: String): ActionFunc {
+    return bindClientFunc({
+        clientFunc: Client4.getOAuthAppInfo,
+        params: [clientId],
+    });
+}
+
+export function allowOAuth2(params: URLSearchParams): ActionFunc {
+    const responseType = params.get('response_type');
+    const clientId = params.get('client_id');
+    const redirectUri = params.get('redirect_uri');
+    const state = params.get('state');
+    const scope = params.get('scope');
+
+    return bindClientFunc({
+        clientFunc: Client4.authorizeOAuthApp,
+        params: [responseType, clientId, redirectUri, state, scope],
+    });
+}
--- a/src/actions/admin.test.js
+++ b/src/actions/admin.test.js
@@ -1037,4 +1037,30 @@ describe('Actions.Admin', () => {
         assert.ok(groups[key].mattermost_group_id === null);
         assert.ok(groups[key].has_syncables === null);
     });
+
+    it('getOAuthAppInfo', async () => {
+        const clientId = 'appId';
+        nock(Client4.getOAuthAppRoute(clientId)).
+            get('/info').
+            reply(200, OK_RESPONSE);
+
+        const {data, error} = await store.dispatch(Actions.getOAuthAppInfo(clientId));
+
+        assert.ok(data);
+        assert.deepStrictEqual(data, OK_RESPONSE);
+        assert.ok(!error);
+    });
+
+    it('allowOAuth2', async () => {
+        nock(Client4.getUrl()).
+            post('/oauth/authorize').
+            reply(200, OK_RESPONSE);
+
+        const params = new URLSearchParams('');
+        const {data, error} = await store.dispatch(Actions.allowOAuth2(params));
+
+        assert.ok(data);
+        assert.deepStrictEqual(data, OK_RESPONSE);
+        assert.ok(!error);
+    });
 });
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -1353,6 +1353,25 @@ export function clearUserAccessTokens(): ActionFunc {
     };
 }
 
+export function getAuthorizedApps(): ActionFunc {
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const currentUserId = getCurrentUserId(getState());
+        const getAuthAppsAction = bindClientFunc({
+            clientFunc: Client4.getAuthorizedOAuthApps,
+            params: [currentUserId],
+        });
+
+        return dispatch(getAuthAppsAction);
+    };
+}
+
+export function deauthorizeOAuthApp(appId: String): ActionFunc {
+    return bindClientFunc({
+        clientFunc: Client4.deauthorizeOAuthApp,
+        params: [appId],
+    });
+}
+
 export default {
     checkMfa,
     generateMfaSecret,
--- a/src/actions/users.test.js
+++ b/src/actions/users.test.js
@@ -1358,4 +1358,33 @@ describe('Actions.Users', () => {
 
         assert.ok(Object.values(myUserAccessTokens).length === 0);
     });
+
+    it('getAuthorizedApps', async () => {
+        const dispatch = store.dispatch;
+        const user = TestHelper.basicUser;
+        TestHelper.mockLogin();
+        await dispatch(Actions.login(user.email, user.password));
+
+        nock(Client4.getUserRoute(user.id)).
+            get('/oauth/apps/authorized').
+            reply(200, OK_RESPONSE);
+
+        const {data, error} = await dispatch(Actions.getAuthorizedApps());
+
+        assert.ok(data);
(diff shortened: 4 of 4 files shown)


src/actions/admin.js
```diff
@@ -547,3 +547,23 @@ export function disablePlugin(pluginId: string): ActionFunc {
         return {data: true};
     };
 }
+
+export function getOAuthAppInfo(clientId: String): ActionFunc {
+    return bindClientFunc({
+        clientFunc: Client4.getOAuthAppInfo,
+        params: [clientId],
```
enahum: is there a reducer that goes along side this? cause I'm not seeing the data sent to any reducer here

src/actions/admin.js
```diff
@@ -547,3 +547,23 @@ export function disablePlugin(pluginId: string): ActionFunc {
...
+export function allowOAuth2(params: URLSearchParams): ActionFunc {
+    const responseType = params.get('response_type');
+    const clientId = params.get('client_id');
+    const redirectUri = params.get('redirect_uri');
+    const state = params.get('state');
+    const scope = params.get('scope');
+
+    return bindClientFunc({
+        clientFunc: Client4.authorizeOAuthApp,
+        params: [responseType, clientId, redirectUri, state, scope],
```
enahum: same as above

src/actions/admin.test.js
```diff
@@ -1037,4 +1037,30 @@ describe('Actions.Admin', () => {
...
+    it('getOAuthAppInfo', async () => {
+        const clientId = 'appId';
+        nock(Client4.getOAuthAppRoute(clientId)).
+            get('/info').
+            reply(200, OK_RESPONSE);
+
+        const {data, error} = await store.dispatch(Actions.getOAuthAppInfo(clientId));
+
+        assert.ok(data);
+        assert.deepStrictEqual(data, OK_RESPONSE);
```
enahum: this test expects the status of the response to be 200 but nothing about the data?

src/actions/admin.test.js
```diff
@@ -1037,4 +1037,30 @@ describe('Actions.Admin', () => {
...
+    it('allowOAuth2', async () => {
+        nock(Client4.getUrl()).
+            post('/oauth/authorize').
+            reply(200, OK_RESPONSE);
+
+        const params = new URLSearchParams('');
+        const {data, error} = await store.dispatch(Actions.allowOAuth2(params));
+
+        assert.ok(data);
+        assert.deepStrictEqual(data, OK_RESPONSE);
```
enahum: same here

MirlanMaksv:
Hi @enahum, thanks for having a look at PR. Yes it could be implemented in a webapp repo, but it was suggested to move them into this repo. Have a look please at the [discussion](https://github.com/mattermost/mattermost-server/issues/10153)

src/actions/admin.js
MirlanMaksv: It only calls the `Client4.getOAuthAppInfo`, it works in this way in a webapp, I did not make any changes in the logic

src/actions/admin.test.js
MirlanMaksv: Yes, the idea is to check that anything server returns passes through the action without being changed. First I did replicate data structure in the response, but at the end we were checking only if they match, so this is the reasong why I changed it to a simpler object `OK_RESPONSE`

MirlanMaksv:
@enahum It is not a problem if it does not meet migration campaign goals. I can close it and update the webapp PR

enahum:
@MirlanMaksv I see you point.. I don't really mind having it here. My only problem is that we are wrapping the client call that we had on the webapp with this `bindClientFun` that I get it does some other checks and that's fine but then we keep on using the data that was returned by the server without storing it in the redux store... no I don't really see the point. Maybe I'm wrong and that's fine but having a redux action that does not changes the redux state seems weird to me.

sudheerDev:
> we keep on using the data that was returned by the server without storing it in the redux store
Why should we be using redux actions only when using data returned by the server into store?

That is perfectly fine. We don't need to store all the data in redux returned by API.
Redux actions should be used whenever unless there is a good reason not to and i don't see a reason 
here not to use it. 

> When it doubt pick a solution which keeps the entropy of codebase low. For me having redux action sounds right

Using client4 calls directly does not seem like a good idea to me in broader sense of as we would be having few API calls going through client4 directly and few through the redux actions. 
Why not just stick to one pattern at the risk of making a decision or having this discussion everytime we add an API call of should we call client4 directly? Are there any downside of having it as redux action?

I don't know the initial intention of  `bindClientFun` but nevertheless it looks like a good abstraction for making API calls.

sudheerDev:
> When it doubt pick a solution which keeps the entropy of codebase low. For me having redux action sounds right

If you ask why it keeps the entropy low it is because we will be having one pattern

The reason why i would stick with redux actions is because we will have access to redux state. 
It will let us do things uniformly across the app if we end up making new decisions on network handling or batching/queuing  of API's calls or even used by plugins to track an event of API call

src/actions/admin.test.js
hmhealey: I'd prefer to fake a real response and use that here. While it doesn't make much any difference to the actual behaviour of the test, I think mocking real responses is a good habit to have. It also makes it so that the tests don't immediately break if someone wants to have the action actually dispatch something based on the response.

MirlanMaksv: It makes sense to me, I will update it

src/actions/admin.js
hmhealey: What would be your thoughts on storing the data returned returned by these endpoints in the redux state instead of in the component state as you're currently doing it [here](https://github.com/mattermost/mattermost-webapp/pull/2387/files#diff-9192ed86bdc9e05ec21d324bbf848198R100)? That'd require adding a few actions and a reducer, but they should be fairly simple ones.

Generally, we try to keep as much state as possible in the redux store, and I think it might be a good idea to do that here as well, at least for the list of authorized OAuth apps

src/actions/admin.js
```diff
@@ -547,3 +547,23 @@ export function disablePlugin(pluginId: string): ActionFunc {
...
 }
+
+export function getOAuthAppInfo(clientId: String): ActionFunc {
+    return bindClientFunc({
+        clientFunc: Client4.getOAuthAppInfo,
+        params: [clientId],
+    });
+}
+
+export function allowOAuth2(params: URLSearchParams): ActionFunc {
```
hmhealey: Minor nit, but it would be nice if we could just use a plain object for the params instead of a `URLSearchParams` since I could see this possibly being used where those values aren't coming directly from the URL. That's just a thought and not something we have to change though.

src/actions/admin.js
MirlanMaksv: I think single source of truth is a good idea, components would be realy pure and following one pattern through out the code is good. But there is another small thing, when data is used very rarely do we need to keep it in a store. Both of them are good for this use case and I think I will go with a redux store as you
> try to keep as much state as possible in the redux store

src/actions/admin.js
MirlanMaksv: Yes nice point, making it plain object would broaden it's usage

MirlanMaksv:
After having [discussed](https://github.com/mattermost/mattermost-server/issues/10153) decision was made to leave actions in a web app repository as they are client specific. Closing the PR

src/actions/admin.js
hmhealey: Thanks! If it ends up being too much work to do that, feel free to leave it in the view state. While I like the idea of moving everything into redux, we've definitely been burned a few times by the "everything" part of decisions like that before. You're definitely right about the rarely used data not really being fit for this store since it can be kept for a long time when we only really need it while that one modal is open.

MirlanMaksv: Yes, I think keeping it simple in this case is a good fit

PR #780, opened 2019-02-25
Handle pending posts when dispatching RECEIVED_NEW_POST

enahum:
#### Summary
When a user submits a post we dispatch a `RECEIVED_NEW_POST` action and then two things happen:
1. When we get the response we dispatch a `RECEIVED_NEW_POST` that will handle the pending post and remove it from the store while adding the one we got from the response.
2. The post is also received through the WebSocket and here we also dispatch a `RECEIVED_NEW_POST` action [example](https://github.com/mattermost/mattermost-webapp/blob/master/actions/post_utils.js#L60)

Sometimes depending on the network we can receive the WebSocket response before the actual Request, meaning that a second `RECEIVED_NEW_POST` is dispatched thus duplicating the same post in the store for a small period of time.

Here we make sure that no matter what response gets to the client first if we have a pending post in the store it gets cleared avoiding the duplication.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14160

#### Checklist

#### Test Information

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -43,6 +43,23 @@ function handleReceivedPost(posts = {}, postsInChannel = {}, postsInThread = {},
         ];
     }
 
+    // Remove any temporary posts
+    if (nextPosts[post.pending_post_id] && post.id !== post.pending_post_id) {
+        Reflect.deleteProperty(nextPosts, post.pending_post_id);
+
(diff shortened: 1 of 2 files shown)


enahum:
@hmhealey don't forget to review this one

enahum:
Thanks @hmhealey I'm going to submit the PR's to update the mm-redux ref for the web and mobile apps

PR #781, opened 2019-02-25
MM-13958/MM-13959 Make postsInChannel into a sparse array

hmhealey:
So big things to note here are:
1. Instead of `postsInChannel` holding an array of all loaded post IDs in the channel, it will now be broken up into "blocks" containing an ordered, contiguous array of post IDs. When loading more posts, we can either add those posts to an existing block (as with getPostsBefore and getPostsAfter) or create a separate block (as with getPosts and getPostsUnread). getPostsSince is a special case because it doesn't ensure that each post is adjacent to one another, so we have some handling to make it sometimes add those posts to the most recent block (more on that below)
2. One of the blocks for each channel will be designated as the most recent block. It will act similarly to the old version of `postsInChannel` where newly received posts will be added to it. We have to take some special care with it since any mixups with it will cause the app to miss messages.
3. Without API changes, we have to manually determine when posts blocks are truly adjacent and when they are not. This can mostly be done by only using the first page of each API, so I've made those changes to the apps. The biggest change here is that loading more posts in the app now uses getPostsBefore and getPostsAfter instead of getPosts.

Also some smaller things to note:
1. Only one block will generally be used at a time. This will normally be the most recent block unless the user is viewing a permalink or an unread post further up in the channel, once that is implemented.
2. Overlapping blocks should automatically be merged by the reducer once we're sure they overlap. It's possible to have two adjacent blocks of posts (say one containing posts 1 to 10 and another with 11 to 20), but we can't be sure there aren't missing posts between them, so we can't change that without API changes.
3. Some of the reducers have had their default values removed and can now return null.

Mobile PR: https://github.com/mattermost/mattermost-mobile/pull/2600
Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/2411

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13958
https://mattermost.atlassian.net/browse/MM-13959

#### Checklist
- Ran `make check-style` to check for style errors (required for all pull requests)
- Ran `make test` to ensure unit tests passed
- Ran `make flow` to ensure type checking passed

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -51,20 +51,38 @@ export function receivedPosts(posts) {
     };
 }
 
-export function receivedPostsSince(posts, channelId, since) {
+export function receivedPostsAfter(posts, channelId, afterPostId) {
+    return {
+        type: PostTypes.RECEIVED_POSTS_AFTER,
+        channelId,
+        data: posts,
+        afterPostId,
+    };
+}
+
+export function receivedPostsBefore(posts, channelId, beforePostId) {
+    return {
+        type: PostTypes.RECEIVED_POSTS_BEFORE,
+        channelId,
+        data: posts,
+        beforePostId,
+    };
+}
+
+export function receivedPostsSince(posts, channelId) {
     return {
         type: PostTypes.RECEIVED_POSTS_SINCE,
         channelId,
         data: posts,
-        since,
     };
 }
 
-export function receivedPostsInChannel(posts, channelId) {
+export function receivedPostsInChannel(posts, channelId, recent = false) {
     return {
         type: PostTypes.RECEIVED_POSTS_IN_CHANNEL,
         channelId,
         data: posts,
+        recent,
     };
 }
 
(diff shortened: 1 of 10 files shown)


src/actions/posts.js
```diff
@@ -663,10 +681,53 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
             getProfilesAndStatusesForPosts(posts.posts, dispatch, getState);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch, getState);
-            dispatch(getState);
+            dispatch(logError(error));
```
enahum: incredible that this was there for who knows how long

src/types/posts.js
```diff
@@ -43,7 +48,7 @@ export type PostMetadata = {|
     reactions: Array<Reaction>
 |};
 
-export type Post = {|
+export type Post = {
```
hmhealey: @jespino Are you okay with this change? This is what I mentioned yesterday where Post being an exact type prevented PostWithFormatData from being passed into functions like `isSystemMessage`

jespino: Yes, that works for me, is not "super-strict" about typing, but I think is a good solution. We can revisit it later when the whole thing is in Flow.

PR #782, opened 2019-02-26
MM-13622 MM-13620 Adding client4 methods, reducers, actions, and selectors for bot accounts.

crspeller:
#### Summary
Added for bot accounts:
- Client4 methods
- Reducers
- Actions
- Selectors

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13620
https://mattermost.atlassian.net/browse/MM-13622

#### Checklist

#### Test Information
This PR was tested on: ArchLinux

### DIFF
--- a/src/actions/bots.test.js
+++ b/src/actions/bots.test.js
@@ -5,6 +5,7 @@ import assert from 'assert';
 import nock from 'nock';
 
 import * as BotActions from 'actions/bots';
+import * as UserActions from 'actions/users';
 import {Client4} from 'client';
 
 import TestHelper from 'test/test_helper';
@@ -31,7 +32,7 @@ describe('Actions.Bots', () => {
             query(true).
             reply(201, bots);
 
-        await BotActions.loadBots()(store.dispatch, store.getState);
+        await store.dispatch(BotActions.loadBots());
 
         const state = store.getState();
         const botsResult = state.entities.bots.accounts;
@@ -45,7 +46,7 @@ describe('Actions.Bots', () => {
             query(true).
             reply(201, bot);
 
-        await BotActions.loadBot(bot.user_id)(store.dispatch, store.getState);
+        await store.dispatch(BotActions.loadBot(bot.user_id));
 
         const state = store.getState();
         const botsResult = state.entities.bots.accounts[bot.user_id];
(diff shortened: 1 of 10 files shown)


src/actions/bots.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+    beforeEach(async () => {
+        store = await configureStore();
+    });
+
+    afterAll(async () => {
+        await TestHelper.tearDown();
+    });
+
+    it('loadBots', async () => {
```
levb: nit: should this test load multiple (2?) fake bots?

src/actions/bots.test.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+
+    it('disableBot', async () => {
+        const bot = TestHelper.fakeBot();
+        nock(Client4.getBotRoute()).
+            post('').
+            reply(200, bot);
+        await BotActions.createBot(bot)(store.dispatch, store.getState);
+
+        bot.delete_at = 1507840900065;
```
levb: nit: maybe a quick comment explaining the relationship between delete_at and `disableBot`?

src/actions/bots.test.js
crspeller: fixed.

src/actions/bots.test.js
crspeller: added.

src/actions/bots.test.js
```diff
@@ -0,0 +1,132 @@
...
+    });
+
+    it('loadBots', async () => {
+        const bots = [TestHelper.fakeBot(), TestHelper.fakeBot()];
+        nock(Client4.getBotsRoute()).
+            get('').
+            query(true).
+            reply(201, bots);
+
+        await BotActions.loadBots()(store.dispatch, store.getState);
```
hmhealey: We should actually be calling `dispatch` on the result of these action creators.

```javascript
await store.dispatch(BotActions.loadBots());
```

Both ways of calling this work the same with async actions, but the way you're doing it doesn't work with plain redux actions. Also, doing it this way saves having to pass `getState` explicitly

src/reducers/entities/bots.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+        return nextState;
+    }
+    case BotTypes.RECEIVED_BOT_ACCOUNT: {
+        const bot = action.data;
+        const nextState = {...state};
+        nextState[bot.user_id] = bot;
+        return nextState;
+    }
+    default:
```
hmhealey: We should probably have a case here to clear the state on logout

src/actions/bots.test.js
crspeller: Fixed.

src/reducers/entities/bots.js
crspeller: Added.

crspeller:
@hmhealey Ready for re-review

PR #789, opened 2019-03-07
MM-13960 Move preparePostIdsForPostList from mobile and re-add combined system posts

hmhealey:
There's a lot of logic moving around here (notably, `preparePostIdsForPostList` is being brought over from the mobile repo and the logic to combine system posts has moved to `utils/post_list` along with it), so this PR might look a bit bigger than it actually is.

The biggest change here is the new `makeCombineUserActivityPosts` which returns a selector that finds adjacent user activity posts and returns an ID like `user-activity-post1_post2_post3` that the `PostList` components in each app will have to understand and turn into a combined post.

I know we went back and forth on how to combine these posts back when it was originally implemented, so please let me know if you have any concerns with doing it this way. Frankly, it made the `postsInChannel` changes easier to do it this way, but there's definitely ways that this is worse than the previous method of doing it in the post reducer.

Mobile PR: https://github.com/mattermost/mattermost-mobile/pull/2637
Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/2465

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13960

#### Checklist


src/utils/post_list.js
```diff
@@ -0,0 +1,404 @@
...
+    return createIdsSelector(
+        (state, {postIds}) => getPostsForIds(state, postIds),
+        (state, {lastViewedAt}) => lastViewedAt,
+        (state, {indicateNewMessages}) => indicateNewMessages,
+        (state) => state.entities.posts.selectedPostId,
+        getCurrentUser,
+        shouldShowJoinLeaveMessages,
+        (posts, lastViewedAt, indicateNewMessages, selectedPostId, currentUser, showJoinLeave) => {
+            if (posts.length === 0 || !currentUser) {
+                return [];
```
saturninoabril: How about returning `null`?

src/utils/post_list.js
```diff
@@ -0,0 +1,404 @@
...
+            let combinedCount = 0;
+
+            const out = [];
+            let changed = false;
+
+            for (let i = 0; i < postIds.length; i++) {
+                const postId = postIds[i];
+
+                if (postId === START_OF_NEW_MESSAGES || postId.startsWith(DATE_LINE)) {
+                    // Not a post, so it won't be combined
```
saturninoabril: Does it mean that we are combining system posts per day?  If yes, then maybe worth mentioning to PM/UX team.

src/actions/posts.js
```diff
@@ -909,7 +907,21 @@ export function getNeededCustomEmojis(state, posts) {
...
+
+            for (const systemPostId of post.system_post_ids) {
+                const systemPost = Selectors.getPost(state, systemPostId);
+
+                if (systemPost) {
+                    dispatch(removePost(systemPost));
+                }
+            }
+        } else {
+            dispatch(postRemoved(post));
```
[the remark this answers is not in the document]
hmhealey: If it was `removePost`, then it would be an infinite loop 😛 

Basically, this just works like `deletePost` now where it recurses on combined posts

src/utils/post_list.js
```diff
@@ -0,0 +1,404 @@
...
+import {Posts, Preferences} from 'constants';
+
+import {makeGetPostsForIds} from 'selectors/entities/posts';
+import {getBool} from 'selectors/entities/preferences';
+import {getCurrentUser} from 'selectors/entities/users';
+
+import {createIdsSelector, memoizeResult} from 'utils/helpers';
+import {isUserActivityPost, shouldFilterJoinLeavePost} from 'utils/post_utils';
+
+export const COMBINED_USER_ACTIVITY = 'user-activity-';
```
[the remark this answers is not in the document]
hmhealey: I don't think we really need to do that. The post type is almost always prefixed with `PostTypes.`, so I don't think they'd get confused. I can still change it if you have a better suggestion, but I can't think of anything right now

src/utils/post_list.js
hmhealey: This matches the current version of the code [here](https://github.com/mattermost/mattermost-mobile/blob/master/app/selectors/post_list.js#L35), but I'm 0/5 on changing it or leaving it the same

src/utils/post_list.js
hmhealey: I actually would've expected the previous version to not combine posts across days, but I can understand why it would happen. That being said, it would be very difficult to combine them across days when doing it this way.

@esethna Any concerns with making this change? Currently, if you invite someone to a channel at 11:59pm and then invite someone else at 12:01am, it will combine those messages, but with these changes it'll no longer do that since there's a date separator in the way.

src/utils/post_list.js
```diff
@@ -0,0 +1,404 @@
...
+                ) {
+                    out.push(START_OF_NEW_MESSAGES);
+                    addedNewMessagesIndicator = true;
+                }
+
+                out.push(post.id);
+            }
+
+            // Flip it back to newest to oldest
+            return out.reverse();
```
sudheerDev: Not related to the PR but it would have been nice to have posts appended to the start of array so we dont have to do reverse here.

PR #791, opened 2019-03-09
MM-1447: check post?.metadata?.files, not just metadata

lieut-data:
#### Summary
On receiving a post from the server, the redux actions will strip out the contents of metadata.files, but leave the metadata object intact (and empty). Some actions take that data and re-emit it back to the store, leading to a race condition where code that doesn't properly check metadata.files will conclude there are no files at all.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14447

#### Checklist

#### Test Information
This PR was tested on: Chrome OSX

### DIFF
--- a/src/reducers/entities/files.js
+++ b/src/reducers/entities/files.js
@@ -109,7 +109,7 @@ export function fileIdsByPostId(state = {}, action) {
 }
 
 function storeFilesIdsForPost(state, post) {
-    if (!post.metadata) {
+    if (!post.metadata || !post.metadata.files) {
         return state;
     }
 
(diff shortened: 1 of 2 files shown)


enahum:
@lieut-data will this also be an issue with reactions and opengraph or is it only with files?

lieut-data:
@enahum, looks like reactions and open graph are already covered. In fact, this access pattern was lifted from those examples when @hmhealey and I narrowed down the mismatch.

amyblais:
Cutting RC-2 soon. Please cherry-pick this soon if this is ready.

PR #793, opened 2019-03-14
For terms of service, don't equate unknown with rejected.

lieut-data:
#### Summary
Explicitly model a `loaded` parameter to distinguish the unknown state from the loaded but rejected state, avoiding a race condition in deciding that the terms of service should be shown when in fact we had just loaded the user prior to fetching the accepted terms of service.

Ideally, the server would return the terms of service inline with the user request (JOINing to the table in question, like we do for fetching bot metadata).

#### Ticket Link
TBD.

#### Checklist

#### Test Information


lieut-data:
This approach turned out just to flip the problem on its head, preventing the TOS from showing when it should.

PR #794, opened 2019-03-15
MM-14386: Get the membership if the group channel is not new

jespino:
#### Summary
We have an optionmization to create the membership in the frontend directly
without calling the server, but this works only if the group channel is brand
new, for a re-created group channel, I have to get the membership from the
server.

#### Ticket Link
[MM-14386](https://mattermost.atlassian.net/browse/MM-14386)

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -9,6 +9,7 @@ import {General, Preferences} from 'constants';
 import {ChannelTypes, PreferenceTypes, UserTypes} from 'action_types';
 import {savePreferences, deletePreferences} from 'actions/preferences';
 import {getChannelsIdForTeam} from 'utils/channel_utils';
+import {getMyChannelMember as getMyChannelMemberSelector} from 'selectors/entities/channels';
 
 import {logError} from './errors';
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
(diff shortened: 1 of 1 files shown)


src/actions/channels.js
```diff
@@ -181,6 +181,10 @@ export function createGroupChannel(userIds: Array<string>): ActionFunc {
             last_update_at: created.create_at,
         };
 
+        if (created.create_at !== created.update_at) {
+            member = await Client4.getMyChannelMember(created.id);
```
jwilander: I think you'll need to wrap this in a try catch, if the request fails this will throw

jespino: Yes, good point

src/actions/channels.js
```diff
@@ -181,6 +181,15 @@ export function createGroupChannel(userIds: Array<string>): ActionFunc {
             last_update_at: created.create_at,
         };
 
+        if (created.create_at !== created.update_at) {
+            try {
+                member = await Client4.getMyChannelMember(created.id);
```
hmhealey: Why not use the `getMyChannelMember` action instead of directly calling the client?

jespino: I going to add a first try to get my member from the store, but it is very possible that you don't have it in the store, because you are opening a new channel, so, probably is because your previous group chat was hidden, so, probably you don't have it in the store.

src/actions/channels.js
```diff
@@ -181,6 +182,20 @@ export function createGroupChannel(userIds: Array<string>): ActionFunc {
             last_update_at: created.create_at,
         };
 
+        if (created.create_at !== created.update_at) {
```
jwilander: Is this checking that the GM is not new? Is this always going to work? Feels like a bit of a hacky assumption. Might be a good idea to add a comment stating the assumption

jespino: Yes, that was the idea, but I think is better to use the `total_msg_count` field, because it detects exactly what i expect. I going to change it and leave a comment anyway.

jespino: Fixed :)

PR #797, opened 2019-03-20
Used user TOS data embedded in user object itself

harshilsharma:
#### Summary
Embedded user TOS data into the user object itself.

#### Checklist

#### Test Information
This PR was tested on: Mozilla Firefox on MacOS Mojave. Pending testing on mattermost desktop app and mobile apps.

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -555,7 +555,7 @@ export function updateMyTermsOfServiceStatus(termsOfServiceId: string, accepted:
                 dispatch({
                     type: UserTypes.RECEIVED_TERMS_OF_SERVICE_STATUS,
                     data: {
-                        create_at: new Date().getTime(),
+                        terms_of_service_create_at: new Date().getTime(),
                         terms_of_service_id: accepted ? termsOfServiceId : null,
                         user_id: getCurrentUserId(getState()),
                     },
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -35,7 +35,7 @@ import {
 } from './preferences';
 
 import {getConfig} from 'selectors/entities/general';
-import {getCurrentUserId, getCurrentUser} from 'selectors/entities/users';
+import {getCurrentUserId} from 'selectors/entities/users';
 
 export function checkMfa(loginId: string): ActionFunc {
     return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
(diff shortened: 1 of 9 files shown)


src/actions/users.js
```diff
@@ -559,14 +547,16 @@ export function updateMyTermsOfServiceStatus(termsOfServiceId: string, accepted:
...
         if (data) {
-            dispatch({
-                type: UserTypes.RECEIVED_TERMS_OF_SERVICE_STATUS,
-                data: {
-                    create_at: new Date().getTime(),
-                    terms_of_service_id: accepted ? termsOfServiceId : null,
-                    user_id: getCurrentUserId(getState()),
-                },
-            });
+            if (accepted) {
```
lieut-data: This pattern worries me, because it's a fake `RECEIVED_ME` event. I get that we're just replacing the data we have plus the now known terms of service status, but perhaps it would be better to continue to emit `RECEIVED_TERMS_OF_SERVICE_STATUS` and instead update the user reducer to respond to this event and write the terms of service into the user?

harshilsharma63: Hmm, makes sense. Will be making the changes.

harshilsharma: Makes sense. Will be making those changes.

harshilsharma: Done.

harshilsharma:
@lieut-data I've made the requested changes.

harshilsharma:
@lieut-data I actually found a bug here. I'm currently investigating it. Will update as soon as I find something.

harshilsharma:
The bug has been fixed. The bug basically was that TOS would re-appear if you updated your profile.

harshilsharma:
@lieut-data also to note, once this PR is merged, the commit ID needs to be reflected in webapp and mobile app.

amyblais:
A quick note that tomorrow (v5.10 Feature Complete) is the last day to merge feature PRs if we want this for v5.10.

src/actions/users.js
```diff
@@ -559,14 +547,20 @@ export function updateMyTermsOfServiceStatus(termsOfServiceId: string, accepted:
...
+                const user = {...getCurrentUser(getState())};
+                user.terms_of_service_id = termsOfServiceId;
+                user.terms_of_service_create_at = new Date().getTime();
```
enahum: I'm not following why we need this here, is not being used anywhere

harshilsharma: You're right, it's not being used. Removed.

harshilsharma:
> other than my comment it looks good. Question does the server return the TOS info on the user object?

Yes, the user's TOS data being returned along with the user object itself to prevent a race condition. In the server it's stored separately, but just returned along with user object in two of the APIs.

PR #799, opened 2019-03-26
Creating the concept of RedirectChannel and use it before the DefaultChannel when we are redirecting the user

jespino:
#### Summary
Change the concept of DEFAULT_CHANNEL to RedirectChannel

#### Ticket Link
[MM-14639](https://mattermost.atlassian.net/browse/MM-14639)

#### Checklist

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -655,7 +655,7 @@ export function joinChannel(userId: string, teamId: string, channelId: string, c
 
 export function deleteChannel(channelId: string): ActionFunc {
     return async (dispatch, getState) => {
-        const state = getState();
+        let state = getState();
         const viewArchivedChannels = state.entities.general.config.ExperimentalViewArchivedChannels === 'true';
 
         try {
@@ -666,16 +666,13 @@ export function deleteChannel(channelId: string): ActionFunc {
             return {error};
         }
 
-        const entities = getState().entities;
-        const {currentChannelId} = entities.channels;
+        state = getState();
+        const {currentChannelId} = state.entities.channels;
         if (channelId === currentChannelId && !viewArchivedChannels) {
             const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, getCurrentTeamId(state)));
-            let redirectChannelId = '';
             if (channel && channel.id) {
-                redirectChannelId = channel.id;
+                dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channel.id}, getState);
             }
-
-            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: redirectChannelId}, getState);
         }
 
         dispatch({type: ChannelTypes.DELETE_CHANNEL_SUCCESS, data: {id: channelId, viewArchivedChannels}}, getState);
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -670,8 +670,12 @@ export function deleteChannel(channelId: string): ActionFunc {
         const {currentChannelId} = entities.channels;
         if (channelId === currentChannelId && !viewArchivedChannels) {
             const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, getCurrentTeamId(state)));
+            let redirectChannelId = '';
+            if (channel && channel.id) {
+                redirectChannelId = channel.id;
+            }
 
-            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channel && channel.id}, getState);
+            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: redirectChannelId}, getState);
         }
 
         dispatch({type: ChannelTypes.DELETE_CHANNEL_SUCCESS, data: {id: channelId, viewArchivedChannels}}, getState);
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -9,7 +9,8 @@ import {General, Preferences} from 'constants';
 import {ChannelTypes, PreferenceTypes, UserTypes} from 'action_types';
 import {savePreferences, deletePreferences} from 'actions/preferences';
 import {getChannelsIdForTeam} from 'utils/channel_utils';
-import {getMyChannelMember as getMyChannelMemberSelector} from 'selectors/entities/channels';
+import {getMyChannelMember as getMyChannelMemberSelector, getChannelByName, getRedirectChannelNameForTeam} from 'selectors/entities/channels';
+import {getCurrentTeamId} from 'selectors/entities/teams';
 
 import {logError} from './errors';
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
@@ -666,15 +667,11 @@ export function deleteChannel(channelId: string): ActionFunc {
         }
 
         const entities = getState().entities;
-        const {channels, currentChannelId} = entities.channels;
+        const {currentChannelId} = entities.channels;
         if (channelId === currentChannelId && !viewArchivedChannels) {
-            const channel = Object.keys(channels).filter((key) => channels[key].name === General.DEFAULT_CHANNEL);
-            let defaultChannelId = '';
-            if (channel.length) {
-                defaultChannelId = channel[0];
-            }
+            const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, getCurrentTeamId(state)));
 
-            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: defaultChannelId}, getState);
+            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channel && channel.id}, getState);
         }
 
         dispatch({type: ChannelTypes.DELETE_CHANNEL_SUCCESS, data: {id: channelId, viewArchivedChannels}}, getState);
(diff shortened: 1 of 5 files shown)


lieut-data:
@jespino, I'm not familiar with the guest accounts feature and all it entails. Can you describe for me what this pull request is accomplishing and how it fits into the larger story?

lieut-data:
Ah, never mind: I found the details I think I need on your description of https://github.com/mattermost/mattermost-webapp/pull/2542. Thanks, @jespino!

src/actions/channels.js
```diff
@@ -666,15 +667,11 @@ export function deleteChannel(channelId: string): ActionFunc {
...
         const entities = getState().entities;
-        const {channels, currentChannelId} = entities.channels;
+        const {currentChannelId} = entities.channels;
         if (channelId === currentChannelId && !viewArchivedChannels) {
-            const channel = Object.keys(channels).filter((key) => channels[key].name === General.DEFAULT_CHANNEL);
-            let defaultChannelId = '';
-            if (channel.length) {
-                defaultChannelId = channel[0];
-            }
+            const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, getCurrentTeamId(state)));
```
hmhealey: Not specific to this PR, but I wonder if there's a utility to compose selectors like this. It seems like two common ways to use selectors are either to combine the results of many selectors which can be done using `createSelector` or to chain them together like this. It would be nice if you didn't have to pass `state` into them repeatedly.

src/selectors/entities/channels.js
```diff
@@ -419,6 +419,14 @@ export const getDefaultChannel: (GlobalState) => ?Channel = createSelector(
     }
 );
 
+export const getDefaultChannelForTeam: (GlobalState, string) => ?Channel = createSelector(
```
hmhealey: Same comment here about not using `createSelector` since this the selector has a parameter.

src/selectors/entities/channels.js
```diff
@@ -1156,3 +1164,21 @@ export const getSortedDirectChannelWithUnreadsIds: (GlobalState, Channel, boolea
         return filterChannels(unreadChannelIds, favoritePreferences, directChannelIds, false, favoritesAtTop);
     },
 );
+
+export const getRedirectChannelNameForTeam: (GlobalState, string) => string = createSelector(
```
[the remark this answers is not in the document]
hmhealey: One other option would be to just return a map of team IDs to channel names. That would also address my comment on `admin_navbar_dropdown/index.js` as well

src/actions/channels.js
jespino: we easily create a function that behave like createSelector but only for returning the result. something like:
```js
function runSelector(...selectors) {
  return (state, options) => {
    let result = options;
    for (const selector of selectors) {
      result = selector(state, result)
    }
    return result
  }
}
let channel = runSelector(
  getCurrentTeamId,
  getRedirectChannelNameForTeam,
  getChannelByName,
)(state)
```

jespino: But I'm 0/5 about adding it, I don't know if its going to be more clear or not.

src/actions/channels.js
```diff
@@ -666,15 +667,11 @@ export function deleteChannel(channelId: string): ActionFunc {
...
 
-            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: defaultChannelId}, getState);
+            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channel && channel.id}, getState);
```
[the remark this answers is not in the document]
jespino: good point, I'll do it

src/selectors/entities/channels.js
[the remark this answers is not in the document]
jespino: The option 1 is not possible, sometimes I need to get the default channel for other teams, not only the current one.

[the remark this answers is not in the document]
jespino: I think I prefer to generate the selectors builders.

src/actions/channels.js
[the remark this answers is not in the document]
jespino: Fixed

src/actions/websocket.js
```diff
@@ -494,24 +494,16 @@ function handleChannelCreatedEvent(msg) {
...
-                const teamChannels = Array.from(channelsInTeam[currentTeamId]);
-                const channel = teamChannels.filter((key) => channels[key].name === General.DEFAULT_CHANNEL);
-
-                if (channel.length) {
-                    channelId = channel[0];
-                }
+                const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, currentTeamId));
 
-                dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channelId});
+                dispatch({type: ChannelTypes.SELECT_CHANNEL, data: channel && channel.id});
```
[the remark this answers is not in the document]
jespino: Fixed

src/selectors/entities/channels.js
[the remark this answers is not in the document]
jespino: Finally I did something in between, let me know what do you think.

src/selectors/entities/channels.js
jespino: Done

jespino:
@hmhealey PTAL

src/selectors/entities/channels.js
```diff
@@ -1156,3 +1156,44 @@ export const getSortedDirectChannelWithUnreadsIds: (GlobalState, Channel, boolea
         return filterChannels(unreadChannelIds, favoritePreferences, directChannelIds, false, favoritesAtTop);
     },
 );
+
+export const getDefaultChannelForTeams: (GlobalState) => RelationOneToOne<Team, Channel> = createSelector(
```
lieut-data: Worth adding some function-level documentation describing the "concept" of default channels here?

src/selectors/entities/channels.js
```diff
@@ -1156,3 +1156,44 @@ export const getSortedDirectChannelWithUnreadsIds: (GlobalState, Channel, boolea
...
+        }
+        return result;
+    }
+);
+
+export const getMyFirstChannelForTeams: (GlobalState) => RelationOneToOne<Team, Channel> = createSelector(
+    getAllChannels,
+    getMyChannelMemberships,
+    getCurrentUser,
+    (channels: IDMappedObjects<Channel>, members: RelationOneToOne<Channel, ChannelMembership>, currentUser: UserProfile): RelationOneToOne<Team, Channel> => {
```
lieut-data: Would it be clearer to rename the parameters here to `allChannels` and `myChannelMemberships`? Matches `currentUser`, and makes it much more obvious what `members[channel.id]` means.

src/selectors/entities/channels.js
```diff
@@ -1156,3 +1156,44 @@ export const getSortedDirectChannelWithUnreadsIds: (GlobalState, Channel, boolea
...
+            if (!result[channel.team_id] && members[channel.id]) {
+                result[channel.team_id] = channel;
+            }
+        }
+        return result
+    }
+);
+
+export const getRedirectChannelNameForTeam = (state: GlobalState, teamId: string): string => {
+    const defaultChannels = getDefaultChannelForTeams(state);
```
lieut-data: nit: naming consistency would help:
* `defaultChannelForteams`
* `myFirstChannelForTeam`
* `canIJoinPublicTeams`
* `myChannelMemberships`

src/selectors/entities/channels.js
```diff
@@ -1156,3 +1156,44 @@ export const getSortedDirectChannelWithUnreadsIds: (GlobalState, Channel, boolea
...
+    const firstChannelForTeams = getMyFirstChannelForTeams(state);
+    const canIJoinPublicTeams = !hasNewPermissions(state) || haveITeamPermission(state, {team: teamId, permission: Permissions.JOIN_PUBLIC_CHANNELS});
+    const members = getMyChannelMemberships(state);
+
+    if ((!defaultChannels[teamId] || !members[defaultChannels[teamId].id]) && !canIJoinPublicTeams) {
```
lieut-data: It's not immediately clear how to interpret this business logic. Could we simplify with early returns and comments?

jespino: Sure

src/selectors/entities/channels.js
jespino: Yes, maybe.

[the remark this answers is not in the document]
jespino: Yes, exactly, I'm not replacing the concept entirely, because the default channel will exist anyway, I'm adding the RedirectChannel in a lot of places as the concept to use, instead the DEFAULT_CHANNEL, but the default channel is still there, in most of the cases under the redirect channel, and in some cases is used as is (for example for read only default channel, and things like that).

src/actions/channels.js
```diff
@@ -666,15 +667,15 @@ export function deleteChannel(channelId: string): ActionFunc {
...
-            let defaultChannelId = '';
-            if (channel.length) {
-                defaultChannelId = channel[0];
+            const channel = getChannelByName(state, getRedirectChannelNameForTeam(state, getCurrentTeamId(state)));
+            let redirectChannelId = '';
+            if (channel && channel.id) {
+                redirectChannelId = channel.id;
             }
 
-            dispatch({type: ChannelTypes.SELECT_CHANNEL, data: defaultChannelId}, getState);
```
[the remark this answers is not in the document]
jespino: I think this can lead to a failure, probably makes sense to not dispatch it at all, and maybe throw an exception.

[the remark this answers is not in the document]

lieut-data:
@jespino, besides the (new) merge conflicts, is this ready to merge to `master`?

jespino:
@lieut-data I'm waiting to get ready the mobile version, probably I'll be able today to review my mobile implementation and then everything will be ready to merge.

PR #805, opened 2019-04-02
MM-14729: discard posts on leaving channel

lieut-data:
#### Summary
Discard posts on leaving a channel (in addition to the existing behaviour of discarding posts when the channel is deleted). This, in turn, suppresses those posts from the flagged and pinned RHS. Note that the postIds will still be present in those data structures (until they are refreshed), but the UI seamlessly handles "missing" posts as such.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14729

#### Checklist

#### Test Information

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -396,7 +396,7 @@ function handlePostDeleted(posts = {}, postsInChannel = {}, postsInThread = {}, 
     return {posts: nextPosts, postsInChannel: nextPostsForChannel, postsInThread: nextPostsForThread};
 }
 
-function handleChannelDeleted(posts = {}, postsInChannel = {}, postsInThread = {}, channelId) {
+function handleChannelRemoved(posts = {}, postsInChannel = {}, postsInThread = {}, channelId) {
     const nextPosts = {...posts};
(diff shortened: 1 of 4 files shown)


src/reducers/entities/posts.js
```diff
@@ -502,7 +502,8 @@ function handlePosts(posts = {}, postsInChannel = {}, postsInThread = {}, action
         return {posts, postsInChannel, postsInThread};
     case ChannelTypes.RECEIVED_CHANNEL_DELETED:
     case ChannelTypes.DELETE_CHANNEL_SUCCESS:
-        if (!action.data.viewArchivedChannels) {
+    case ChannelTypes.LEAVE_CHANNEL:
+        if (action.data && !action.data.viewArchivedChannels) {
             return handleChannelDeleted(posts, postsInChannel, postsInThread, action.data.id);
```
jespino: nit: I think would be a good option to change the name of the function to `handleChannelDeletedOrLeft`

lieut-data: Renamed to `handleChannelRemoved` to cover both deleted, left, and technically kicked.

PR #816, opened 2019-04-20
Add actions to save/delete theme preference

cometkim:
#### Summary

This is a migration PR for https://github.com/mattermost/mattermost-webapp/pull/2679

#### Checklist

### DIFF
--- a/src/actions/preferences.js
+++ b/src/actions/preferences.js
@@ -5,7 +5,7 @@
 import {Client4} from 'client';
 import {Preferences} from 'constants';
 import {PreferenceTypes} from 'action_types';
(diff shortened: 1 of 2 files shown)


enahum:
Wondering if these functions should be part of the webapp instead of having them in this repo.

Thoughts?

jwilander:
Themes work for mobile right? Why would these just be in the web app?

enahum:
They do work on mobile but if I'm not mistaken it makes use of `savePreferences` directly.

I'm ok having this here and then changing the code in the mobile app to make use of it.

cometkim:
I think saveTheme here is fine. It's a simple wrapper of savePreferences with a side effect and having this in webapp would make a circular dependency.

cometkim:
Yep, sure thing.

PR #820, opened 2019-04-24
[MM-13500] Add function to match a search term against a group channel

mgdelacroix:
#### Summary
Adds a function to filter a list of group channels checking if they match against a search term using their `display_name` property.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13500

#### Checklist

### DIFF
--- a/src/utils/channel_utils.js
+++ b/src/utils/channel_utils.js
@@ -656,3 +656,15 @@ export function filterChannelsMatchingTerm(channels: Array<Channel>, term: strin
             displayName.startsWith(lowercasedTerm);
     });
 }
+
+export function filterGroupChannelsMatchingTerm(groupChannels: Array<Channel>, term: string): Array<Channel> {
+    const lowercasedTerm = term.toLowerCase();
+
+    return groupChannels.filter((channel: Channel): boolean => {
+        const displayName = (channel.display_name || '').toLowerCase();
+
+        return displayName.
+            split(', ').
+            some((username) => username.startsWith(lowercasedTerm));
+    });
+}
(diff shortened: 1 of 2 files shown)


src/utils/channel_utils.js
```diff
@@ -656,3 +656,10 @@ export function filterChannelsMatchingTerm(channels: Array<Channel>, term: strin
             displayName.startsWith(lowercasedTerm);
     });
 }
+
+export function groupChannelMatchesTerm(groupChannel: Channel, term: string): boolean {
```
jespino: Why not follow the previous naming convention? something like `filterGroupChannelMatchingTerm`

mgdelacroix: Makes sense not only with the name, but the implementation too :)

src/utils/channel_utils.js
```diff
@@ -656,3 +656,15 @@ export function filterChannelsMatchingTerm(channels: Array<Channel>, term: strin
             displayName.startsWith(lowercasedTerm);
     });
 }
+
+export function filterGroupChannelsMatchingTerm(groupChannels: Array<Channel>, term: string): Array<Channel> {
+    const lowercasedTerm = term.toLowerCase();
+
+    return groupChannels.filter((channel: Channel): boolean => {
+        const displayName = (channel.display_name || '').toLowerCase();
```
jwilander: Unfortunately I'm not sure if this will work in all cases. Mainly if the group channel includes a lot of users or users with longer usernames since we truncate the display_name here if it exceeds the character limit https://github.com/mattermost/mattermost-server/blob/master/model/channel.go#L228

I can't think of a good way around that without adding the other server endpoint for searching groups or by including some more metadata in the GMs when we create them (which causes problems with migration)

Happy to hop on a call if you want to discuss

PR #822, opened 2019-04-25
Reverting serialize-error to version 2.1.0

jespino:
#### Summary
Reverting serialize-error to version 2.1.0

serialize-error is though for node, and the version 3.0.0 is using ES6 directly, without transpiling anything (added the Node 6 requirement) and it breaks IE11, so... we must stick to 2.1.0 or remove entirely the dependency (is used in one single place, the `logError` action.

#### Ticket Link
[MM-15285](https://mattermost.atlassian.net/browse/MM-15285)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10136,12 +10136,9 @@
       }
     },
     "serialize-error": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/serialize-error/-/serialize-error-4.0.0.tgz",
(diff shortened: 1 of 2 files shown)


jespino:
Other option would be embed the serialize-error code entirely in our code (is not too much code, only 2 functions) and we can remove the dependency entirely, avoiding future upgrades to screw us.

jespino:
Now I'm starting to think this is more related to babel upgrade, or something like that. I'm still investigating

jespino:
Finally I can test it, and is a combination of 4 things:

- 1. the development compilation generate a security error (which is not important, because the production build doesn't execute that code).
- 2. The serialize-error dependency here.
- 3. the exif2css dependency in mattermost-webapp.
- 4. a error accessing plugin.initialize when a plugin has fail to initialize.

So I going to make another PR in mattermost-webapp to fix everything else.

PR #826, opened 2019-05-03
Fixes RN master branch build

enahum:
#### Summary
Mattermost-mobile is failing to build in release mode cause of the way mm-redux is built, by updating dependencies and setting `@babel/preset-env` to 7.2.0 the mobile app builds successfully.

We should definitely need to find a way to not have to build the library before it is used by any of the clients.


package.json
```diff
@@ -27,38 +27,39 @@
...
-    "@babel/core": "7.4.0",
-    "@babel/plugin-proposal-class-properties": "7.4.0",
-    "@babel/plugin-transform-flow-comments": "7.4.0",
-    "@babel/preset-env": "7.4.2",
-    "@babel/register": "7.4.0",
+    "@babel/cli": "7.4.4",
+    "@babel/core": "7.4.4",
+    "@babel/plugin-proposal-class-properties": "7.4.4",
+    "@babel/plugin-transform-flow-comments": "7.4.4",
+    "@babel/preset-env": "7.2.0",
```
migbot: What changed between 7.2.0 and 7.4.2 that is the problem? Also, does the latest 7.2.* not work?

enahum: I didn't look into the detail of what changed, nor I'm very interested in it at the moment as this helps to fix the current issue of not being able to build the app.

Also the changelog seems outdated https://github.com/babel/babel-preset-env/blob/1.x/CHANGELOG.md

crspeller: My guess is that it's the update to corejs 3.

enahum: @migbot 7.2.3 works as well, I can bump it to that one if we prefer.

First I'm going to try and include corejs 3 in the mobile app and see if that fixes it, I'll report back as soon as I test it

enahum:
@crspeller webapp is working fine, just tested it. lets hold the merge until I test it all with core-js@3

enahum:
core-js@3 does not crash the app but it does not work.. it does not render anything.

Anyway ticket is here https://mattermost.atlassian.net/browse/MM-15424 I really don't know what team should own it and I included a recommendation about having the clients building the ES6 code of this library.

Closing and re-opening to re-trigger the build so I can merge it

PR #821, opened 2019-04-25
MM-14410: Adds ability to retrieve groups that are associated (or not-associated) to teams.

mkraft:
#### Summary

Adds ability to retrieve groups that are associated (or not-associated) to teams.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-14410

#### Checklist

- [x] Ran `make check-style` to check for style errors (required for all pull requests)
- [x] Ran `make test` to ensure unit tests passed
- [x] Ran `make flow` to ensure type checking passed
- [x] Added or updated unit tests (required for all new features)
- [x] All new/modified APIs include changes to the drivers

#### Test Information

Chrome on macOS

#### Awaiting

- https://github.com/mattermost/mattermost-server/pull/10701

PR #827, opened 2019-05-06
MM-14897: Changes to be able to add and remove groups from channels.

mkraft:
#### Summary

Change to be able to add and remove groups from channels.

This is pretty much the same thing as https://github.com/mattermost/mattermost-redux/pull/821, but for channels.

Relies on: https://github.com/mattermost/mattermost-server/pull/10794

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-14897

#### Checklist

#### Test Information

Chrome browser, macOS.


src/selectors/entities/groups.js
```diff
@@ -51,6 +51,11 @@ function getTeamGroupIDSet(state, teamID) {
     return new Set(arr);
 }
 
+function getChannelGroupIDSet(state, channelID) {
```
jespino: this selector will every single time generate a new object, we could create a function for the array, and a selector (based on that function) for the set, that way, if the underneath array hasn't change, the set will be cached.

mkraft: Good point.

PR #831, opened 2019-05-08
MM-14295  - Add "user_actual_role" field to client-side diagnostics to track if a user is a System Admin

reflog:
#### Summary
Add "user_actual_role" field to client-side diagnostics to track if a user is a System Adminuser_actual_role", to connect it with 

Whenever a client-side event is sent to Segment, record the user's role and mark it as "system_admin" if the user is a System Admin, otherwise null.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14295

#### Checklist

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -139,6 +139,7 @@ function completeLogin(data: UserProfile): ActionFunc {
         });
 
         Client4.setUserId(data.id);
+        Client4.setUserRoles(data.roles);
 
         let teamMembers = null;
         try {
@@ -236,10 +237,11 @@ export function loadMe(): ActionFunc {
             dispatch(getAllCustomEmojis());
         }
 
-        await Promise.all(promises);
+        const results: $Subtype<ActionResult>[] = await Promise.all(promises);
 
         const {currentUserId} = getState().entities.users;
         Client4.setUserId(currentUserId);
+        Client4.setUserRoles(results[0].data.roles);
 
         return {data: true};
     };
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -237,11 +237,12 @@ export function loadMe(): ActionFunc {
             dispatch(getAllCustomEmojis());
         }
 
-        const results: $Subtype<ActionResult>[] = await Promise.all(promises);
+        await Promise.all(promises);
 
         const {currentUserId} = getState().entities.users;
+        const user = getState().entities.users.profiles[currentUserId];
         Client4.setUserId(currentUserId);
-        Client4.setUserRoles(results[0].data.roles);
+        Client4.setUserRoles(user.roles);
 
         return {data: true};
     };
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -2745,7 +2745,7 @@ export default class Client4 {
         const properties = Object.assign({
             category,
             type: event,
-            user_actual_role: isSystemAdmin(this.userRoles) ? 'system_admin' : null,
+            user_actual_role: this.userRoles && isSystemAdmin(this.userRoles) ? 'system_admin' : null,
             user_actual_id: this.userId,
         }, props);
         const options = {
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {isSystemAdmin} from '../utils/user_utils';
+
 const FormData = require('form-data');
 
 import fetch from './fetch_etag';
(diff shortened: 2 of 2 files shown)


reflog:
Please not that I am not sure on how to test this, since analytics are disabled whne running locally. Please advise :)

src/actions/users.js
```diff
@@ -236,10 +237,11 @@ export function loadMe(): ActionFunc {
             dispatch(getAllCustomEmojis());
         }
 
-        await Promise.all(promises);
+        const results: $Subtype<ActionResult>[] = await Promise.all(promises);
 
         const {currentUserId} = getState().entities.users;
         Client4.setUserId(currentUserId);
+        Client4.setUserRoles(results[0].data.roles);
```
jespino: Probably here is a better idea to follow the same pattern than used in setUserId, something like:
```suggestion
        const user = getState().entities.users.profiles[currentUserId];
        Client4.setUserRoles(user.roles);
```

src/client/client4.js
```diff
@@ -2736,7 +2742,12 @@ export default class Client4 {
             return;
         }
 
-        const properties = Object.assign({category, type: event, user_actual_id: this.userId}, props);
+        const properties = Object.assign({
+            category,
+            type: event,
+            user_actual_role: this.userRoles && isSystemAdmin(this.userRoles) ? 'system_admin' : null,
```
jespino: Probably here we will need in the future something more clear to include the guest users. So I would set here it to 'system_admin' or 'system_user' for now, and in the future we can return 'system_admin', 'system_user' or 'system_guest'. (The "system" is important here, because we are only talking about the role at system level, not the role in the context of the team or channel).

src/actions/users.js
reflog: :+1:

src/client/client4.js
reflog: I agree, but the ticket specifically asked to only include the system_admin flag...

jespino: Yes, you are right, this was explicitly design that way @jasonblais what do you think about my proposed change?

reflog: Done!

reflog: done!

lieut-data: We're modelling data here, not the UI, so IMO we should be using the internal terms of `sytem_admin` and `system_user`. Grepping for either shows lots of context in `mattermost-server`. Grepping for `system_member` shows nothing. 

Whatever interprets this data downstream is free to render it as "system_member" if it makes more sense to those consuming it, but I think it will be easier to work with if we maintain consistency.

lieut-data: (GitHub has somehow screwed up the comment order here -- my reply above was definitely /after/ @jasonblais's comment at the bottom.)

reflog: @jespino , @jasonblais - do we have a consensus on the value? system_member vs system_user

jespino: @lieut-data I think I agree with you about it, but I get the @jasonblais point which is, the analytics data is not something consumed by the developers, normally is somehow user facing data. I don't have an strong opinion here, for me, makes more sense the `system_user` but I'm very biased by the fact that is what I see all the time in the code.

jasonblais: What's consistent with other diagnostics data? Do we typically follow the internal term, or the external (UI) term?

jespino: I think is not super consistent in that sense, so... I feel right either way, adding this consistent with the system data or use it as user facing data.

reflog:
I vote for system_user. Easier to grep...

src/client/client4.js
jasonblais: Question: Do System Admins inheirt both the `system_user` and `system_admin` role, or is it only the `system_admin` role?

reflog:
Both

src/client/client4.js
jasonblais: > Both

Then I'm wondering how you all feel about this proposal:

1. If System Admin, mark the user's role as `system_admin, system_user`
2. Else, mark the user's role as `system_user`

Pros:
1. Follows our data model (obvious for a developer audience)
2. Follows the convention used in the UI, where "everyone in the system" is considered a user (obvious for a data consumer audience)
3. Scaleable, as we might add more roles in future (think of team_admin, channel_admin) where we'd want to support a list of roles the user has inherited.

@reflog @lieut-data @jespino

jespino: I like it too

lieut-data:
@reflog, don't forget to re-request a review once you've made changes, otherwise your reviewers won't be notified. (Some details at https://developers.mattermost.com/blog/submitting-great-prs/, exhaustive details at https://developers.mattermost.com/contribute/getting-started/code-review/).

reflog:
Thanks Jesse, I'll take note for future PRs!

PR #836, opened 2019-05-21
[MM-15474] Dispatch getMyTeams and getMyTeamMembers on websocket reconnect

migbot:
#### Summary
The TeamList component of the main sidebar uses the `getMySortedTeamIds` selector, which itself calls `getTeams` and `getTeamMemberships`, to compute the IDs for the teams the user is a member of, however, if the client's websocket connection is disconnected then any team removals/additions will be missed as `getTeams` and `getTeamMemberships` will not include the updated state. Dispatching `getMyTeams` and `getMyTeamMembers` fixes this and awaiting these actions prior to checking membership prevents crashes due to an unhandled promise rejection in `fetchMyChannelsAndMembers`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15474

#### Checklist

#### Test Information
This PR was tested on:
* Galaxy S7, Android 7.0
* iPhone 8, iOS 12.2

Note: I was having trouble using [test_store.js](https://github.com/mattermost/mattermost-redux/blob/MM-15474/test/test_store.js) with the unit test I added but then saw this [comment](https://github.com/mattermost/mattermost-redux/blob/MM-15474/test/test_store.js#L44) about redux-mock-store so I went ahead and installed it as a dev dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "mock-socket": "8.0.5",
     "nock": "10.0.6",
     "react": "16.8.6",
+    "redux-mock-store": "^1.5.3",
     "redux-persist-node-storage": "2.0.0",
     "remote-redux-devtools": "0.5.16",
     "remotedev-rn-debugger": "0.8.4",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "mock-socket": "8.0.5",
     "nock": "10.0.6",
     "react": "16.8.6",
-    "redux-mock-store": "^1.5.3",
+    "redux-mock-store": "1.5.3",
     "redux-persist-node-storage": "2.0.0",
     "remote-redux-devtools": "0.5.16",
     "remotedev-rn-debugger": "0.8.4",
--- a/src/actions/websocket.js
+++ b/src/actions/websocket.js
@@ -116,7 +116,7 @@ export function close(shouldReconnect = false) {
     };
 }
 
-function doReconnect() {
+export function doReconnect() {
     return (dispatch, getState) => {
         const state = getState();
         const currentTeamId = getCurrentTeamId(state);
--- a/src/actions/websocket.js
+++ b/src/actions/websocket.js
@@ -145,16 +145,15 @@ export function doReconnect() {
             const currentTeamMembership = getCurrentTeamMembership(getState());
             if (currentTeamMembership) {
                 dispatch(getPosts(currentChannelId));
-                dispatch(fetchMyChannelsAndMembers(currentTeamId)).then(({data}) => {
-                    dispatch(loadProfilesForDirect());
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "mock-socket": "8.0.5",
     "nock": "10.0.6",
     "react": "16.8.6",
+    "redux-mock-store": "^1.5.3",
```
jwilander:
```suggestion
    "redux-mock-store": "1.5.3",
```
We don't auto-update any dependencies

enahum: once you do this you'll need to update the package-lock.json as well

src/actions/websocket.js
```diff
@@ -131,22 +136,26 @@ function doReconnect() {
...
-            const myTeamMembers = getTeamMemberships(getState());
-            if (!myTeamMembers[currentTeamId]) {
+            // We need to wait for these actions so that we have an
+            // up-to-date state of the current user's team memberships.
+            await dispatch(getMyTeams());
+            await dispatch(getMyTeamMembers());
+            const currentTeamMembership = getCurrentTeamMembership(getState());
+            if (currentTeamMembership) {
+                dispatch(getPosts(currentChannelId));
+                dispatch(fetchMyChannelsAndMembers(currentTeamId)).then(({data}) => {
```
enahum: if you are using async/await then why not changing this from the promise to also use await?

src/actions/websocket.test.js
```diff
@@ -430,3 +439,106 @@ describe('Actions.Websocket', () => {
...
 });
+
+describe('Actions.Websocket doReconnect', () => {
+    const mockStore = configureMockStore([thunk]);
+
+    const currentTeamId = 'team-id';
+    const currentUserId = 'user-id';
+    const currentChannelId = 'channel-id';
+
+    nock(Client4.getBaseRoute()).
```
enahum: I believe you also need to mock the other endpoints or it will fail to run the tests

src/actions/websocket.js
jwilander: I was going to comment the same thing but using async/await here does have a functional change. `dispatch({type: GeneralTypes.WEBSOCKET_SUCCESS});` will now wait for all those requests to complete before being run. Not sure if this is a problem or not but it is a change we should be aware of

migbot: @jwilander Ah right. I'm going to revert the change unless @enahum objects. I'm wondering if `WEBSOCKET_SUCCESS` should be the first thing dispatched in this function as we know the connection is actually established since this is a reconnect callback on the websocket client?

jwilander: It actually might be better to wait for the requests to finish since the connection bar will go green once all the updates have been fetched (nothing pops in after the connection bar is green). I'm 0/5 either way

migbot: OK merging as is.

PR #839, opened 2019-05-22
MM-15667 if post was removed exclude it from postsInChannel

enahum:
#### Summary
When a post was removed from a channel and then we retrieved the posts since the reducer was trying to sort the previously removed post and trying to access a property on an undefined object.

Note: We'll need to update the reference for the webapp and mobile apps

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15667

#### Checklist

### DIFF
--- a/src/reducers/entities/posts.js
+++ b/src/reducers/entities/posts.js
@@ -511,7 +511,7 @@ export function postsInChannel(state = {}, action, prevPosts, nextPosts) {
         for (let i = order.length - 1; i >= 0; i--) {
             const postId = order[i];
 
-            if (nextPosts[postId].create_at <= mostRecentCreateAt) {
+            if (!nextPosts[postId] || nextPosts[postId].create_at <= mostRecentCreateAt) {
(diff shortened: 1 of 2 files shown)


enahum:
@hmhealey review?

src/reducers/entities/posts.js
```diff
@@ -511,7 +511,7 @@ export function postsInChannel(state = {}, action, prevPosts, nextPosts) {
         for (let i = order.length - 1; i >= 0; i--) {
             const postId = order[i];
 
-            if (nextPosts[postId].create_at <= mostRecentCreateAt) {
+            if (!nextPosts[postId] || nextPosts[postId].create_at <= mostRecentCreateAt) {
```
hmhealey: Can you add a separate `if (...) continue` block for this with a separate comment explaining what's going on? The "This is an old post" comment doesn't apply here.

enahum: @hmhealey done

PR #840, opened 2019-05-27
#10883 Allow plugins to dismiss posts through the MessageWillBePosted hook

carmo-evan:
#### Summary
adding check for dissmissPostError to dismiss posts through the MessageWillBePosted hook , adding [batchActionsWithThunk](https://github.com/tshelburne/redux-batched-actions/issues/3)

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/10883

#### Checklist

#### Test Information
This PR was tested on: Chrome 74

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1,6 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {batchActionsWithThunk} from 'utils/helpers';
 import {batchActions} from 'redux-batched-actions';
 
 import {Client4} from 'client';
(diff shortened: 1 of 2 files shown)


carmo-evan:
@hmhealey would you please review this draft after reading through [the issue](https://github.com/mattermost/mattermost-server/issues/10883)? Just wondering if this is the best approach before I make a full PR.

hmhealey:
I'm having trouble understanding how `batchActionsWithThunk` works since it seems pretty complicated. What if we just remove the call to `batchActions` instead of adding a version that works with thunk? This part of the code isn't critical to performance, so I don't think it would change much to have the redux state change twice here.

carmo-evan:
That works for me.

PR #841, opened 2019-05-27
MM-15751 Fix for system messages not rendering in IE

sudheerDev:
#### Summary
  * Export constants COMBINED_USER_ACTIVITY_PREFIX and MAX_COMBINED_SYSTEM_POSTS
    from posts constants file for usage in webapp

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15751

#### Checklist

### DIFF
--- a/src/constants/posts.js
+++ b/src/constants/posts.js
@@ -36,6 +36,8 @@ export default {
         POST: 'post',
         COMMENT: 'comment',
     },
+    COMBINED_USER_ACTIVITY_PREFIX: 'user-activity-',
+    MAX_COMBINED_SYSTEM_POSTS: 100,
     MAX_PREV_MSGS: 100,
     POST_COLLAPSE_TIMEOUT: 1000 * 60 * 5, // five minutes
     IGNORE_POST_TYPES: [
--- a/src/constants/posts.js
+++ b/src/constants/posts.js
@@ -36,8 +36,6 @@ export default {
         POST: 'post',
         COMMENT: 'comment',
     },
-    COMBINED_USER_ACTIVITY_PREFIX: 'user-activity-',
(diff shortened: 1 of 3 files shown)


hmhealey:
Do we have to move those constants to `constants/posts`? The files in `constants` are full of random, unrelated values, so I left `COMBINED_USER_ACTIVITY` in that file because it'll always be used with the other values in that file.

sudheerDev:
@hmhealey Not really but if i am going to use it in webapp i felt we should move to constants. If in the previous file i doubt if it would be obvious that it is being used in webapp and might cause an issue if removed.

hmhealey:
I think we should keep them all in `utils/post_list` since that's where they're used. It doesn't make sense to have them in separate files when you'd want to use them together anyway

sudheerDev:
@hmhealey I am going to use these in https://github.com/mattermost/mattermost-webapp/pull/2865 so moving them to constants file. Sorry i am not sure if you had noticed it. You still want to keep them in the same file even if i am going to use them in webapp?

hmhealey:
Yeah, I'd prefer to keep them in the same place. The other constants that are exported from that file are also used in the web app

sudheerDev:
@hmhealey make changes but it seems odd that we want to keep constants in the file and export them when using in different repos. I feel we should be keeping them in constants file if we use them in different repos

> Re: The files in constants are full of random, unrelated values

Sounds like a tech debt ticket but if we don't keep them in constants file then cleaning up after these later will be even harder.

So i propose that we keep constants in file if we use only use in the repo and keep them in constants file if we use them across repos.

hmhealey:
Thanks for moving those back. I think using the constants file makes sense for more general values (like action types or file extensions), but for things like this where the code is tightly coupled to the constants, I think they should stay together

PR #849, opened 2019-06-11
Adding flow types

Azero123:
#### Summary
Updating src/actions/posts.js to use Flow. Some changes were made to have conformity amongst the redux functions.

#### Ticket Link
mattermost/mattermost-server#9771

#### Checklist

#### Test Information
This PR was tested on: [Macbook Pro, Mac OS 10.13.6]

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// @flow
 
 import {batchActions} from 'redux-batched-actions';
 
@@ -30,9 +31,14 @@ import {
 } from './preferences';
 import {getProfilesByIds, getProfilesByUsernames, getStatusesByIds} from './users';
 
+import type {DispatchFunc, GetStateFunc, ActionFunc, GenericAction} from '../types/actions';
+import type {Post, PostEmbed} from '../types/posts';
+import type {FileInfo} from '../types/files';
+import type {GlobalState} from '../types/store';
+
 // receivedPost should be dispatched after a single post from the server. This typically happens when an existing post
 // is updated.
-export function receivedPost(post) {
+export function receivedPost(post: Post): GenericAction {
     return {
         type: PostTypes.RECEIVED_POST,
         data: post,
@@ -41,7 +47,7 @@ export function receivedPost(post) {
 
 // receivedNewPost should be dispatched when receiving a newly created post or when sending a request to the server
 // to make a new post.
-export function receivedNewPost(post) {
+export function receivedNewPost(post: Post): GenericAction {
     return {
         type: PostTypes.RECEIVED_NEW_POST,
         data: post,
@@ -50,7 +56,7 @@ export function receivedNewPost(post) {
 
 // receivedPosts should be dispatched when receiving multiple posts from the server that may or may not be ordered.
 // This will typically be used alongside other actions like receivedPostsAfter which require the posts to be ordered.
-export function receivedPosts(posts) {
+export function receivedPosts(posts: Array<Post>): GenericAction {
     return {
         type: PostTypes.RECEIVED_POSTS,
         data: posts,
@@ -58,7 +64,7 @@ export function receivedPosts(posts) {
 }
 
 // receivedPostsAfter should be dispatched when receiving an ordered list of posts that come before a given post.
-export function receivedPostsAfter(posts, channelId, afterPostId) {
+export function receivedPostsAfter(posts: Array<Post>, channelId: string, afterPostId: string): Object {
     return {
         type: PostTypes.RECEIVED_POSTS_AFTER,
         channelId,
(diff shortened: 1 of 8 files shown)


src/actions/posts.js
```diff
@@ -165,7 +171,7 @@ export function createPost(post, files = []) {
             return {data: true};
         }
 
-        let newPost = {
+        let newPost: Object = {
```
hmhealey: Shouldn't this be a Post?

src/actions/posts.js
```diff
@@ -790,7 +796,7 @@ export function getProfilesAndStatusesForPosts(postsArrayOrMap, dispatch, getSta
     return Promise.all(promises);
 }
 
-export function getNeededAtMentionedUsernames(state, posts) {
+export function getNeededAtMentionedUsernames(state: GlobalState, posts: Array<Post>): ?Set<string> {
```
hmhealey: Can this just return a `Set<string>` without the `?`? It doesn't look like we ever return null here

src/actions/posts.js
```diff
@@ -808,27 +814,30 @@ export function getNeededAtMentionedUsernames(state, posts) {
 
         let match;
         while ((match = pattern.exec(post.message)) !== null) {
+            // $FlowFixMe, flow does not recognize the check of !== null in the while loop
```
hmhealey: We might be able to fix this by changing the loop to use a `while(true)` instead like
```javascript
while (true) {
    const match = pattern.exec(post.message);
    if (!match) {
        break;
    }

    // ...
}
```

src/actions/posts.js
```diff
@@ -899,8 +908,8 @@ export function getNeededCustomEmojis(state, posts) {
     return customEmojisToLoad;
 }
 
-export function removePost(post) {
-    return (dispatch, getState) => {
+export function removePost(post: Post): ActionFunc {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
```
hmhealey: Why change this to be an async function?

src/types/actions.js
```diff
@@ -17,12 +17,13 @@ export type GenericAction = {|
...
     currentUserId?: string,
     remove?: Function,
+    url?: string
 |};
 
 type Thunk = (DispatchFunc, GetStateFunc) => Promise<ActionResult>; // eslint-disable-line no-use-before-define
 
 type BatchAction = {
-    type: 'BATCHING_REDUCER.BATCH';
+    type: string;
```
hmhealey: Why change this type? If I understand correctly, the previous version of this just means that the `type` field always has to have that value

src/types/posts.js
```diff
@@ -29,12 +29,18 @@ export type PostType = 'system_add_remove' |
 
 export type PostEmbedType = 'image' | 'message_attachment' | 'opengraph';
 
+/* eslint-disable no-use-before-define */
```
hmhealey: Why is this needed here?

src/types/preferences.js
```diff
@@ -6,5 +6,5 @@ export type PreferenceType = {|
     category: string,
     name: string,
     user_id: string,
-    value: string
+    value?: string
```
hmhealey: Why is this being changed to optional?

src/actions/posts.js
Azero123: I asked earlier in the developer community what some thought as to if we should standardize them all to be the same since they all have the behavior of returning a dispatch and getstate function and it seemed that this was the preference.

hmhealey: Oh, I see. I didn't realize that the `ActionFunc` required it to be async. This is fine to leave as is then, although I'm generally not a fan of adding async everywhere since it makes it harder to test this code.

hanzei:
@Azero123 Do you have questions about the provided feedback? Looking forward to seeing this PR merged.

src/actions/posts.js
sudheerDev: I didn't realise that we are changing them to be promises for `ActionFunc` type when this was being discussed. Rather than changing this to be `async` i think we should change the type to be either a `func` or a `promise`. Same with `return`. If we are returning `{data: true}` just for the sake of returning  then it does not add much value so we probably should change the type for return too. 
i am 0/5 on the change for return type but for changing async i am at 2/5 that we should change the type to be forgiving if it is a a simple func.

@jespino Thoughts?

jespino: Yes, I think is a valid usage to have sync actions, so I agree with @sudheerDev, we should change the `ActionFunc` to accept both, regular functions or async functions. about the return, I would say was the pre-existing pattern, so I wouldn't change it for now, maybe in the future, we have to rethink a lot of things in mattermost-redux.

sudheerDev: @Azero123 Can you change the type of `ActionFunc` to accept a func? And change back the instances of newly added async function to be sync as before?. I am good with changing return type to be `{data: true}`.
Sorry about the confusion and delay on reviewing this.

hanzei:
Hey @Azero123,

Please note that we will close this PR after another 10 days of inactivity. If you don't object, other contributors might complete this PR for you.

hanzei:
Closing this PR due to inactivity. Still, thank you for this contribution @Azero123 :+1:

PR #855, opened 2019-06-15
Set post list date separators based on the user timezone if available

enahum:
#### Summary
With the changes of the post list and by moving the logic to get the list to redux we omitted the fact that a user can have a timezone set causing the date separators in the post list to reflect the wrong date when the timezone is set manually.

A follow up PR will be submitted against the webapp and mobile to include this redux changes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16387

#### Checklist

#### Test Information

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "harmony-reflect": "1.6.1",
     "isomorphic-fetch": "2.2.1",
     "mime-db": "1.40.0",
+    "moment-timezone": "0.5.25",
     "redux": "4.0.1",
     "redux-action-buffer": "1.2.0",
     "redux-batched-actions": "0.4.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "harmony-reflect": "1.6.1",
     "isomorphic-fetch": "2.2.1",
     "mime-db": "1.40.0",
+    "moment-timezone": "0.5.25",
```
enahum: added to simplify the way we get the offset for the use timezone setting

package.json
```diff
@@ -57,7 +58,6 @@
     "redux-mock-store": "1.5.3",
     "redux-persist-node-storage": "2.0.0",
     "remote-redux-devtools": "0.5.16",
-    "remotedev-rn-debugger": "0.8.4",
```
enahum: no really needed and this dependency has a lot of insecure dependencies

PR #861, opened 2019-06-21
MM-16442: expose post hasReactions utils function

lieut-data:
#### Summary
This helper method takes into account the possibility of a server without post metadata, enabling the client to check for the presence of reactions (or the need to load them!).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16442

#### Checklist

#### Test Information
This PR was tested on: Safari OSX, iOS Simulator

### DIFF
--- a/src/types/posts.js
+++ b/src/types/posts.js
@@ -69,6 +69,7 @@ export type Post = {
     failed?: boolean,
     user_activity_posts?: Array<Post>,
     state?: 'DELETED',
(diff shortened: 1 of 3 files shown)


lieut-data:
Sigh, this won't work -- I was mistakenly testing this with a server change that reverted the websocket event. It turns out that we don't /get/ the post metadata without that websocket event, so there's nothing there to help improve the situation.

PR #862, opened 2019-06-21
MM-16480 Don't store incomplete channels 

hmhealey:
This bug was happening because users receive `channel_deleted` events for channels that they're not members of, and that caused the reducer for `entities.channels.channels` to an incomplete channel object that only contains a `delete_at` timestamp and nothing else. When passed into `sortChannelsByDisplayName`, this caused it to throw an error because the channel's `name` field was missing.

This fixes the issue by preventing the channels reducer from storing incomplete channels. I also added some additional protection to `getMyFirstChannelForTeams` since it seems to be the only place where we pass channels into `sortChannelsByDisplayName` without indirectly checking if the channel is complete by seeing things like if the current user is a member of it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16480

#### Checklist

### DIFF
--- a/src/actions/channels.test.js
+++ b/src/actions/channels.test.js
@@ -1550,7 +1550,7 @@ describe('Actions.Channels', () => {
 
     it('updateChannelHeader', async () => {
         nock(Client4.getChannelsRoute()).
-            post(`${TestHelper.basicChannel.id}`).
+            get(`/${TestHelper.basicChannel.id}`).
             reply(200, TestHelper.basicChannel);
 
         await store.dispatch(Actions.getChannel(TestHelper.basicChannel.id));
@@ -1567,7 +1567,7 @@ describe('Actions.Channels', () => {
 
     it('updateChannelPurpose', async () => {
         nock(Client4.getChannelsRoute()).
-            post(`${TestHelper.basicChannel.id}`).
+            get(`/${TestHelper.basicChannel.id}`).
             reply(200, TestHelper.basicChannel);
 
         await store.dispatch(Actions.getChannel(TestHelper.basicChannel.id));
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -1217,8 +1217,9 @@ export const getMyFirstChannelForTeams: (GlobalState) => RelationOneToOne<Team, 
         const result = {};
         for (const team of myTeams) {
             // Get a sorted array of all channels in the team that the current user is a member of
+            // $FlowFixMe
             const teamChannels = Object.values(allChannels).
-                filter((channel) => channel && channel.team_id === team.id && myChannelMemberships[channel.id]).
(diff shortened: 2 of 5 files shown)


src/selectors/entities/channels.js
```diff
@@ -1209,15 +1209,25 @@ export const getDefaultChannelForTeams: (GlobalState) => RelationOneToOne<Team,
...
-    (allChannels: IDMappedObjects<Channel>, myChannelMemberships: RelationOneToOne<Channel, ChannelMembership>, currentUser: UserProfile): RelationOneToOne<Team, Channel> => {
+    (allChannels: IDMappedObjects<Channel>, myChannelMemberships: RelationOneToOne<Channel, ChannelMembership>, myTeams: Array<Team>, currentUser: UserProfile): RelationOneToOne<Team, Channel> => {
         const locale = currentUser.locale || General.DEFAULT_LOCALE;
+
         const result = {};
-        for (const channel of Object.keys(allChannels).map((key) => allChannels[key]).sort(sortChannelsByDisplayName.bind(null, locale))) {
-            if (!result[channel.team_id] && myChannelMemberships[channel.id]) {
-                result[channel.team_id] = channel;
+        for (const team of myTeams) {
+            // Get a sorted array of all channels in the team that the current user is a member of
```
lieut-data: Appreciate the extra guard, but wondering if it's strictly necessary? If we added all checks downstream like this, the code could quickly become cluttered.

hmhealey: It's not strictly necessary, but I don't think it really adds complexity in this case since I'm kind of just rewriting the same function in a different order by checking channel membership before sorting. There is a small functional change since we won't return channels for teams that you're not a member of, but I would've expected the function to work this way anyway.

src/actions/channels.test.js
```diff
@@ -1550,7 +1550,7 @@ describe('Actions.Channels', () => {
 
     it('updateChannelHeader', async () => {
         nock(Client4.getChannelsRoute()).
-            post(`${TestHelper.basicChannel.id}`).
+            get(`/${TestHelper.basicChannel.id}`).
```
jespino: Why this change?

src/actions/channels.test.js
```diff
@@ -1567,7 +1567,7 @@ describe('Actions.Channels', () => {
 
     it('updateChannelPurpose', async () => {
         nock(Client4.getChannelsRoute()).
-            post(`${TestHelper.basicChannel.id}`).
+            get(`/${TestHelper.basicChannel.id}`).
```
jespino: and this one

src/actions/channels.test.js
hmhealey: The getChannel API wasn't mocked properly before since it uses a GET request, not a POST. The test still passed because of the bug that this PR fixes, so we had to fix the mock to be able to fix the bug.

src/actions/channels.test.js
hmhealey: Same here

PR #864, opened 2019-06-24
MM-15783 Adding automatic CSRF from cookie

crspeller:
#### Summary
In order to facilitate an easier experience for plugins and other clients that use Client4, adding automatic handing of the CSRF cookie. It can be overridden by calling the old methods.

Webapp PR forthcoming to use this automatic system.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15783

### DIFF
--- a/src/client/client4.js
+++ b/src/client/client4.js
@@ -269,6 +269,19 @@ export default class Client4 {
         return `${this.getBotsRoute()}/${botUserId}`;
     }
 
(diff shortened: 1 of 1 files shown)


crspeller:
@DSchalla It would be dead code for mobile. But it won't be dead code for any browser user like the webapp or plugins. So I think it's justified in having it here. I left the possibility of overriding using the old API in.

PR #865, opened 2019-06-25
Adding promote/demote actions to redux

jespino:
#### Summary
Adding promote/demote actions to redux

#### Ticket Link
[MM-15060](https://mattermost.atlassian.net/browse/MM-15060)


jespino:
@enahum We can, but I don't see the point of it, I'm using the `bindCliendFunc` right away, and the client function, so I'll have to mock the client function, and then I'll be testing the `bindCliendFunc` which I understand that is tested already. I'm not generating store changes, or anything else, so I think the tests aren't needed.

PR #870, opened 2019-06-27
MM-16477 Check for modified users on reconnect

hmhealey:
The mobile and web apps can miss changes to other users that happen when the websocket is disconnected. This makes it so that the app will download any modified users after reconnecting to the websocket.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16477

#### Related PRs
Server: https://github.com/mattermost/mattermost-server/pull/11406
Mobile: https://github.com/mattermost/mattermost-mobile/pull/2936
Web: https://github.com/mattermost/mattermost-webapp/pull/3028
API Reference: https://github.com/mattermost/mattermost-api-reference/pull/457

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -1398,7 +1398,7 @@ export function clearUserAccessTokens(): ActionFunc {
 }
 
 export function checkForModifiedUsers() {
-    return (dispatch, getState) => {
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
         const state = getState();
         const users = getUsers(state);
         const lastDisconnectAt = state.websocket.lastDisconnectAt;
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -35,7 +35,7 @@ import {
 } from './preferences';
 
 import {getConfig} from 'selectors/entities/general';
-import {getCurrentUserId} from 'selectors/entities/users';
+import {getCurrentUserId, getUsers} from 'selectors/entities/users';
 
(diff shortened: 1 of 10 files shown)


src/actions/users.js
```diff
@@ -1397,6 +1397,16 @@ export function clearUserAccessTokens(): ActionFunc {
     };
 }
 
+export function checkForModifiedUsers() {
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
+        const users = getUsers(state);
+        const lastDisconnectAt = state.websocket.lastDisconnectAt;
+
+        return dispatch(getProfilesByIds(Object.keys(users), {since: lastDisconnectAt}));
```
enahum: So on an older server, this will fetch all the users that you have already loaded in the store, right?

hmhealey: Oh right, I forgot to add the version check. Thanks for the reminder

src/actions/users.test.js
```diff
@@ -1338,4 +1338,61 @@ describe('Actions.Users', () => {
...
+                        profiles: {},
+                    },
+                },
+                websocket: {
+                    lastDisconnectAt,
+                },
+            });
+
+            await store.dispatch(Actions.checkForModifiedUsers());
+        });
```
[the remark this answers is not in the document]
hmhealey: I was kind of testing the implicit behaviour that the mock wouldn't be called, but it wouldn't hurt to test that as well, so I'll add it

PR #873, opened 2019-07-02
MM-11504: Add unread API actions and selectors for bi-directional scrolling

sudheerDev:
#### Summary
* Add getUnreadPostsAction
* Add getUnreadPosts API to client4
* Add selector makeGetPostsChunkAroundPost
* Add util func getPostsChunkInChannelAroundTime
* Mark recent true for chunks when dispatching based on post.next_post_id
  for unread API, posts after API

#### Ticket Link
[MM-11504](https://mattermost.atlassian.net/browse/MM-11504)

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -58,12 +58,13 @@ export function receivedPosts(posts) {
 }
 
 // receivedPostsAfter should be dispatched when receiving an ordered list of posts that come before a given post.
-export function receivedPostsAfter(posts, channelId, afterPostId) {
+export function receivedPostsAfter(posts, channelId, afterPostId, recent = false) {
     return {
         type: PostTypes.RECEIVED_POSTS_AFTER,
         channelId,
         data: posts,
         afterPostId,
+        recent,
     };
 }
 
@@ -620,6 +621,33 @@ export function getPosts(channelId, page = 0, perPage = Posts.POST_CHUNK_SIZE) {
     };
 }
 
+export function getPostsUnread(channelId) {
+    return async (dispatch, getState) => {
+        const userId = getCurrentUserId(getState());
+        let posts;
+        try {
+            posts = await Client4.getPostsUnread(channelId, userId);
+            getProfilesAndStatusesForPosts(posts.posts, dispatch, getState);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+
+        dispatch(batchActions([
+            receivedPosts(posts),
+            receivedPostsInChannel(posts, channelId, posts.next_post_id === ''),
+        ]));
+        dispatch({
+            type: PostTypes.RECEIVED_POSTS,
+            data: posts,
+            channelId,
+        });
+
+        return {data: posts};
+    };
+}
+
 export function getPostsSince(channelId, since) {
     return async (dispatch, getState) => {
         let posts;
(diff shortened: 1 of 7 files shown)


src/reducers/entities/posts.js
```diff
@@ -666,6 +661,10 @@ export function mergePostBlocks(blocks, posts) {
     // Remove any blocks that may have become empty by removing posts
     nextBlocks = removeEmptyPostBlocks(blocks);
 
+    if (!nextBlocks.length) {
```
hmhealey: When would merging post blocks ever remove all of the post blocks?

sudheerDev: If a channel does not have any posts(Experimental feature where join and leave messages don't exist) its good have the state with empty order and posts as it will have a recent flag which is beneficial in various ways. This check here removes that only block.

Sorry i should have commented few things before turning on the label. My bad. i am gonna through again to see if i can clarify any other tricky parts of the changes here

src/actions/posts.js
```diff
@@ -679,7 +707,7 @@ export function getPostsAfter(channelId, postId, page = 0, perPage = Posts.POST_
 
         dispatch(batchActions([
             receivedPosts(posts),
-            receivedPostsAfter(posts, channelId, postId),
+            receivedPostsAfter(posts, channelId, postId, posts.next_post_id === ''),
```
sudheerDev: Using `next_post_id` from API to turn on recent flag on chunk which is used to determine if we have a loader for loading newer posts

src/selectors/entities/posts.js
```diff
@@ -493,22 +505,48 @@ export const getCurrentUsersLatestPost: (GlobalState, $ID<Post>) => ?PostWithFor
...
+    if (!postsForChannel) {
+        return null;
+    }
+
+    const blockAroundTimestamp = postsForChannel.find((block) => {
+        const {order} = block;
+        const recentPostInBlock = posts[order[0]];
+        const oldestPostInBlock = posts[order[order.length - 1]];
+        if (recentPostInBlock && oldestPostInBlock) {
+            return (recentPostInBlock.create_at > timeStamp && oldestPostInBlock.create_at < timeStamp) || recentPostInBlock.create_at === timeStamp || oldestPostInBlock.create_at === timeStamp;
```
migbot: Can this just be `>= && <=`?

src/reducers/entities/posts.js
hmhealey: When in doubt, adding comments like that to the code is always good :)

And that makes sense

src/selectors/entities/posts.js
```diff
@@ -493,22 +505,48 @@ export const getCurrentUsersLatestPost: (GlobalState, $ID<Post>) => ?PostWithFor
...
+
     if (!recentBlock) {
         return null;
     }
 
     return recentBlock.order;
 }
 
+export function getPostsChunkInChannelAroundTime(state: GlobalState, channelId: $ID<Channel>, timeStamp: number): ?Object {
+    const postsEnitiy = state.entities.posts;
```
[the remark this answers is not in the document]
sudheerDev: Aah ya, Made the change.

src/selectors/entities/posts.js
sudheerDev: Yup i this works, I can never wrap my head around multiple conditions ever. Changed it

src/reducers/entities/posts.js
sudheerDev: Added the comment in the code

PR #875, opened 2019-07-08
MM-11039

mzaks:
#### Summary
Migrate WebSocket actions to Flow

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/9762
https://mattermost.atlassian.net/browse/MM-11039

#### Checklist

#### Test Information
Did not test. Suppose to be a non destructive refactoring.

#### Notes from contributor
I introduce comments in place, which I find need discussion.


src/actions/websocket.js
```diff
@@ -212,9 +222,10 @@ function handleEvent(msg) {
     case WebsocketEvents.UPDATE_TEAM:
         doDispatch(handleUpdateTeamEvent(msg));
         break;
-    case WebsocketEvents.PATCH_TEAM:
-        doDispatch(handlePatchTeamEvent(msg));
-        break;
+
+    // case WebsocketEvents.PATCH_TEAM:
+    //     doDispatch(handlePatchTeamEvent(msg));
+    //     break;
```
mzaks: There is no `WebsocketEvents.PATH_TEAM` constant present. I commented out the case. But it is rather needs to be removed. Comments welcome

src/actions/websocket.js
```diff
@@ -355,19 +366,20 @@ function handleUpdateTeamEvent(msg) {
...
-        type: TeamTypes.PATCHED_TEAM,
-        data: JSON.parse(msg.data.team),
-    };
-}
+// function handlePatchTeamEvent(msg) {
+//     return {
+//         type: TeamTypes.PATCHED_TEAM,
+//         data: JSON.parse(msg.data.team),
+//     };
+// }
```
mzaks: There is no WebsocketEvents.PATH_TEAM constant present. I commented out the handle function. But it is rather needs to be removed. Comments welcome

src/actions/websocket.js
```diff
@@ -388,7 +400,7 @@ function handleUserAddedEvent(msg) {
         }, getState);
 
         if (msg.broadcast.channel_id === currentChannelId) {
-            dispatch(getChannelStats(teamId, currentChannelId));
+            dispatch(getChannelStats(currentChannelId));
```
mzaks: based on the function signature, the code before `getChannelStats(teamId, currentChannelId)` was a bug. However I would like to get input from more experienced developers.

src/actions/websocket.js
```diff
@@ -435,7 +447,7 @@ function handleUserRemovedEvent(msg) {
                 EventEmitter.emit(General.SWITCH_TO_DEFAULT_CHANNEL, currentTeamId);
             }
         } else if (msg.data.channel_id === currentChannelId) {
-            dispatch(getChannelStats(currentTeamId, currentChannelId));
+            dispatch(getChannelStats(currentChannelId));
```
mzaks: See comment on line 403

src/actions/websocket.js
jespino: Yes, that was a bug, clearly, I think is the correct behavior to change it here :)

cpanato:
please ignore the messages from matterbuild. I will check what is wrong when comeback from PTO

src/actions/websocket.js
```diff
@@ -239,9 +253,6 @@ function handleEvent(msg) {
     case WebsocketEvents.UPDATE_TEAM:
         doDispatch(handleUpdateTeamEvent(msg));
         break;
-    case WebsocketEvents.PATCH_TEAM:
```
reflog: Why was this removed?

mzaks: I discussed it with @jespino before in a private conversation, sorry I did not write a comment in PR. Long story short, there is no such event 🙂. By introducing Flow, the flow checker showed that `WebsocketEvents.PATCH_TEAM` constant is not existing. I then checked with @jespino and he confirmed that it is not there and there is also no `PATCHED_TEAM` on the BackEnd. So we decided that it would be the best to delete the case and the handle function.

mzaks: @reflog sorry I am not familiar with the way it is handled correctly, do I get to click on the `Resolve conversation` button? 🙂

reflog: Thanks for the explanation @mzaks !
About resolve conversation - the originator of the comment presses Resolve if the question was answered (so in this case I will press Resolve) or by the committer if they handled the comment (for example if you fixed the bug mentioned).

src/actions/websocket.js
reflog: Please remove it.

src/actions/websocket.js
reflog: Understood. Please remove it.

mzaks: All done. The snipped is `Outdated`. All commented out code is already deleted.

src/actions/websocket.js
mzaks: The code is removed. Now resolving the conversation.

mzaks:
Any news on this PR?

reflog:
> Any news on this PR?

@mzaks you forgot to press 'request review' after pushing your changes, so I had no idea there was any work on the PR :)

PR #887, opened 2019-07-26
[MM-15379] Check for undefined channel

migbot:
#### Summary
I'm not sure how we're getting undefined values [here](https://github.com/mattermost/mattermost-redux/blob/master/src/selectors/entities/channels.js#L874), but the stack trace in the [Sentry report](https://sentry.io/organizations/mattermost-mr/issues/1075141568/?project=205624) points to that being the issue. Added a check prior to calling `.name`. Once this PR is merged I'll open a PR on the mobile repo.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15379

#### Checklist

### DIFF
--- a/src/selectors/entities/channels.js
+++ b/src/selectors/entities/channels.js
@@ -875,7 +875,7 @@ export const getDirectChannels: (GlobalState) => Array<Channel> = createSelector
         const directChannelsIds = [];
         teammates.reduce((result, teammateId) => {
             const name = getDirectChannelName(currentUser.id, teammateId);
-            const channel = channelValues.find((c: Channel) => c.name === name); //eslint-disable-line max-nested-callbacks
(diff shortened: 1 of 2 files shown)


jespino:
@migbot yes, that is related to flow and type assertions, is not possible to assert types with Object.values(), so the simplest workaround is that. Probably with the "typescript" proposal from @reflog we can back to previous approach.

enahum:
Flow in this case is just being stupid, the impact is not that negative, but cmon...

jespino:
@enahum probably not good enough because you using a map so... I guest by default any map object can contains empty "not-found" keys, which is undefined, so... probably it always is something like `?Type`, instead of `Type`. Clearly flow have some corner cases that aren't well covered.

PR #896, opened 2019-08-08
Mm 17064

Willyfrog:
#### Summary

Add marking a post as unread feature to redux.
this will reset the counters and mentions on the store.

#### Ticket Link

Jira [MM-17064](https://mattermost.atlassian.net/browse/MM-17064)

#### Checklist

#### Test Information

### DIFF
--- a/src/action_types/channels.js
+++ b/src/action_types/channels.js
@@ -76,5 +76,5 @@ export default keyMirror({
 
     RECEIVED_TOTAL_CHANNEL_COUNT: null,
 
-    CHANNEL_UNREAD_SUCCESS: null,
+    POST_UNREAD_SUCCESS: null,
 });
--- a/src/action_types/channels.js
+++ b/src/action_types/channels.js
@@ -75,4 +75,6 @@ export default keyMirror({
     RECEIVED_CHANNEL_MEMBERS_MINUS_GROUP_MEMBERS: null,
 
     RECEIVED_TOTAL_CHANNEL_COUNT: null,
+
+    CHANNEL_UNREAD_SUCCESS: null,
 });
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -366,8 +366,9 @@ export function editPost(post) {
 
 export function setUnreadChannel(postId) {
     return async (dispatch, getState) => {
+        let unreadChan;
         try {
-            await Client4.markChannelAsUnread(postId);
+            unreadChan = await Client4.markChannelAsUnread(postId);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch, getState);
             dispatch(logError(error));
@@ -375,9 +376,8 @@ export function setUnreadChannel(postId) {
         }
         dispatch({
             type: ChannelTypes.CHANNEL_UNREAD_SUCCESS,
-            data: null,
+            data: unreadChan,
         });
-        // NOTE: should we update the data from the channel?
         return {data: true};
     };
 }
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -5,7 +5,7 @@ import {batchActions} from 'redux-batched-actions';
 
 import {Client4} from 'client';
 import {General, Preferences, Posts, WebsocketEvents} from 'constants';
-import {PostTypes, FileTypes, IntegrationTypes} from 'action_types';
+import {PostTypes, FileTypes, IntegrationTypes, ChannelTypes} from 'action_types';
 
 import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} from 'selectors/entities/channels';
 import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
@@ -362,6 +362,24 @@ export function editPost(post) {
             post,
         ],
     });
+}
+
+export function setUnreadChannel(postId) {
+    return async (dispatch, getState) => {
+        try {
+            await Client4.markChannelAsUnread(postId);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+        dispatch({
+            type: ChannelTypes.CHANNEL_UNREAD_SUCCESS,
+            data: null,
+        });
+        // NOTE: should we update the data from the channel?
+        return {data: true};
+    };
 }
 
 export function pinPost(postId) {
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -364,19 +364,29 @@ export function editPost(post) {
     });
 }
 
-export function setUnreadChannel(postId) {
+export function setUnreadPost(userId, postId) {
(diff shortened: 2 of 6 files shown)


src/actions/posts.test.js
```diff
@@ -1117,6 +1117,51 @@ describe('Actions.Posts', () => {
...
+            team_id: teamId,
+            channel_id: channelId,
+            msg_count: 3,
+            mention_count: 1,
+        });
+
+        await store.dispatch(Actions.setUnreadPost(userId, postId));
+        const state = store.getState();
+
+        // assert.equal(state.entities.channels.myMembers[channelId], null);
```
jespino: unnecesary comment

jespino:
@Willyfrog I miss the changes here `src/reducers/requests/posts.js` and here `store/initial_state.js` to really track the requests in the store

Willyfrog:
> @Willyfrog I miss the changes here `src/reducers/requests/posts.js` and here `store/initial_state.js` to really track the requests in the store

ah, now it makes more sense. I didn't find those files, guess I was looking at one that wasn't tracked. I'll submit the required changes right away :)

hmhealey:
We actually started removing the request state at some point since we don't use it very often, and we thought we could improve performance slightly by removing some unused redux actions. I don't think we came to any real conclusions regarding performance though, so it doesn't hurt to add new ones.

src/actions/posts.js
```diff
@@ -364,6 +364,40 @@ export function editPost(post) {
     });
 }
 
+export function setUnreadPost(userId, postId) {
+    return async (dispatch, getState) => {
+        dispatch({type: ChannelTypes.POST_UNREAD_REQUEST}, getState);
```
hmhealey: `dispatch` actually only takes one argument, so you don't need to pass `getState` into it
```suggestion
        dispatch({type: ChannelTypes.POST_UNREAD_REQUEST});
```

src/actions/posts.js
```diff
@@ -364,6 +364,40 @@ export function editPost(post) {
...
+        let unreadChan;
+        const state = getState();
+        try {
+            unreadChan = await Client4.markPostAsUnread(userId, postId);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: PostTypes.POST_UNREAD_FAILURE, error},
+                logError(error),
+            ]), getState);
```
hmhealey: Same here
```suggestion
            ]));
```

src/actions/posts.js
```diff
@@ -364,6 +364,40 @@ export function editPost(post) {
     });
 }
 
+export function setUnreadPost(userId, postId) {
+    return async (dispatch, getState) => {
+        dispatch({type: ChannelTypes.POST_UNREAD_REQUEST}, getState);
+
+        let unreadChan;
+        const state = getState();
```
hmhealey: This call should be moved to happen after the network request. Depending on how long the network request takes, the state might have changed, so we could be looking at old data by doing it this way.

src/actions/posts.js
```diff
@@ -364,6 +364,40 @@ export function editPost(post) {
...
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: PostTypes.POST_UNREAD_FAILURE, error},
+                logError(error),
+            ]), getState);
+            return {error};
+        }
+
+        const oldMsgCount = state.entities.channels.myMembers[unreadChan.channel_id].msg_count;
```
hmhealey: Is there any chance that `state.entities.channels.myMembers[unreadChan.channel_id]` might end up being undefined? I'm wondering if this could use a null check at all

src/actions/posts.js
Willyfrog: I shaw dispatch receiving the getState function as parameter in several other places, I thought I was missing something as having it didn't break something (also js doesn't complain about too many arguments...) so I decided to err on the safe side.

I'll remove it

src/actions/posts.js
Willyfrog: It's intentional, by having it there I'll get the amount it had when requested which feels as close to the expected outcome as I can think of. having it after the network request might mean that an action taken after this one, might affect it. Same thing can happen with a previous action being slower with this one.

So it seems that we might have some race conditions wherever I put this, what I'm not sure is which one is the one that will feel more ackward or is it going to happen more often.

src/action_types/channels.js
```diff
@@ -75,4 +75,8 @@ export default keyMirror({
     RECEIVED_CHANNEL_MEMBERS_MINUS_GROUP_MEMBERS: null,
 
     RECEIVED_TOTAL_CHANNEL_COUNT: null,
+
+    POST_UNREAD_FAILURE: null,
```
enahum: Do we need to track the request?

src/actions/posts.js
Willyfrog: It shouldn't happen as we are considering an action which happens after viewing the channel, but that check won't hurt. So again I'll err on the safe side and add the check :)

src/action_types/channels.js
Willyfrog: Not for functionality, it was suggested by @jespino and it might be useful in the future, so I added it as it didn't hurt. Let me know if you prefer to remove it

enahum: @jespino any reason you wanted to track the request? If there is no clear reason we should definitely remove. “It does not hurt” is clearly an understatement cause by tracking the request you’ll cause the state to change at least 3 times which is not ideal

Willyfrog: you are right about the state. I'll remove it unless @jespino has a reason for that

jespino: No, I have no reason, only consistency, but if we going to remove that, I would add a Jira ticket for removing unused request tracks there, because looks like everything is tracked, if we only have a small bunch of them is more obvious that is only used when is needed.

enahum: I believe the majority were already removed but @sudheerDev can confirm

sudheerDev: @enahum Nope not all are removed. Only 2/5 were removed which effect center channel perf. 

@jespino @Willyfrog Lets not add them if we are not going to use them as we want to remove any of these flags which are not needed.

We cannot move away from this pattern completely as we might need to have the state if two components are looking for the status of these requests

Willyfrog: already removed!

sudheerDev: @enahum Correction more than half of the important request flags are removed. https://mattermost.atlassian.net/browse/MM-13544. 

But other two tickets which should have been on backlog got closed https://mattermost.atlassian.net/browse/MM-13545, https://mattermost.atlassian.net/browse/MM-13545. I just reopened them again and triaged. We can consider them for HW

Willyfrog:
switching to a feature branch as I need to add more stuff on a different PR but everything should be merged together in the end

PR #897, opened 2019-08-10
Add keyboard appearance field in default themes

cardoso:
#### Summary
This adds the `keyboardAppearance` parameter to the default themes which can be used in the mobile app to specify keyboard appearance. It makes the keyboard `'light'` or `'dark'` on iOS. React Native hasn't implemented this effect on Android yet as of 0.60. ([ref](https://facebook.github.io/react-native/docs/textinput#keyboardappearance))

#### Checklist

#### Test Information
This PR was tested on: [
    iPhone SE iOS 12.2,
    iPhone 5s iOS 12.2 Simulator,
    Google Nexus 5X Genymotion Simulator API Version 24
]

### DIFF
--- a/src/constants/preferences.js
+++ b/src/constants/preferences.js
@@ -70,6 +70,7 @@ export default {
             mentionHighlightBg: '#ffe577',
             mentionHighlightLink: '#166de0',
             codeTheme: 'github',
+            keyboardAppearance: 'light',
         },
         organization: {
             type: 'Organization',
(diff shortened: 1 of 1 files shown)


enahum:
@cardoso thanks for this, but I think you are missing how does this is set for custom themes and what would be the default value if it's not present in the current selected theme.

enahum:
in fact I think this is not needed, we can just use the mechanism we have to detect the color to set the status bar to `light` or `dark`

cardoso:
@enahum I thought this way it would be more flexible for theme makers. But since it's not flexible for status bar already, I can roll with the same approach. I'll update the PR in mobile and close this one. Thanks!

PR #898, opened 2019-08-14
[MM-17390] Rate limited posts fail to load threads

cpoile:
#### Summary
- Rate-limited posts are batched, and those posts skip the `completePostReceived` path, which includes getting the post's root. If the user then clicks on the thread, they would receive the message:
image
- This fix uses the same strategy as the `getProfilesAndStatusesForPosts` function, which is also run for the list of rate-limited posts.

#### Ticket Link
- [MM-17390](https://mattermost.atlassian.net/browse/MM-17390)

#### Checklist
- [x]  Ran `make check-style` to check for style errors (required for all pull requests)

#### Test Information
This PR was tested on: Mac, Chrome

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -11,6 +11,7 @@ import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} f
 import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
 import {getConfig} from 'selectors/entities/general';
 import * as Selectors from 'selectors/entities/posts';
+import * as PostActions from 'actions/posts';
 import {getCurrentUserId, getUsersByUsername} from 'selectors/entities/users';
 
 import {getUserIdFromChannelName} from 'utils/channel_utils';
@@ -757,6 +758,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
 
         return {data: posts};
     };
+}
+
+// getThreadsForPosts is intended for an array of posts that have been batched
+// (see the actions/websocket_actions/handleNewPostEvents function in the webapp)
+export function getThreadsForPosts(posts, dispatch, getState) {
+    if (!Array.isArray(posts) || posts.length === 0) {
+        return Promise.resolve();
+    }
+
+    const promises = [];
+
+    posts.forEach((post) => {
+        if (!post.root_id) {
+            return;
+        }
+        const rootPost = Selectors.getPost(getState(), post.root_id);
+        if (post.root_id && !rootPost) {
+            promises.push(PostActions.getPostThread(post.root_id)(dispatch, getState));
+        }
+    });
+
+    return Promise.all(promises);
 }
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -762,24 +762,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
 
 // getThreadsForPosts is intended for an array of posts that have been batched
 // (see the actions/websocket_actions/handleNewPostEvents function in the webapp)
-export function getThreadsForPosts(posts, dispatch, getState) {
-    if (!Array.isArray(posts) || posts.length === 0) {
-        return Promise.resolve();
-    }
-
-    const promises = [];
-
-    posts.forEach((post) => {
-        if (!post.root_id) {
-            return;
+export function getThreadsForPosts(posts) {
+    return (dispatch, getState) => {
+        if (!Array.isArray(posts) || !posts.length) {
+            return {data: true};
         }
-        const rootPost = Selectors.getPost(getState(), post.root_id);
-        if (post.root_id && !rootPost) {
-            promises.push(PostActions.getPostThread(post.root_id)(dispatch, getState));
-        }
-    });
 
-    return Promise.all(promises);
+        const state = getState();
+        const promises = [];
+
+        posts.forEach((post) => {
+            if (!post.root_id) {
+                return;
+            }
+
+            const rootPost = Selectors.getPost(state, post.root_id);
+            if (!rootPost) {
+                promises.push(dispatch(PostActions.getPostThread(post.root_id)));
+            }
+        });
+
+        return Promise.all(promises);
+    };
 }
 
 // Note that getProfilesAndStatusesForPosts can take either an array of posts or a map of ids to posts
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -11,7 +11,6 @@ import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} f
 import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
 import {getConfig} from 'selectors/entities/general';
 import * as Selectors from 'selectors/entities/posts';
-import * as PostActions from 'actions/posts';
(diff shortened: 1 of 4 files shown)


src/actions/posts.js
```diff
@@ -759,6 +760,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
...
+// getThreadsForPosts is intended for an array of posts that have been batched
+// (see the actions/websocket_actions/handleNewPostEvents function in the webapp)
+export function getThreadsForPosts(posts, dispatch, getState) {
+    if (!Array.isArray(posts) || posts.length === 0) {
+        return Promise.resolve();
+    }
+
+    const promises = [];
+
+    posts.forEach((post) => {
```
lieut-data: Aha, here's the complexity. Is it true that for each post we would have "normally" received, we call `getPostThread`, and so doing them all now isn't any "worse" as such? Slightly worried about slamming the server in this scenario, and wondering if we need a `getPostsThreads` instead.

cpoile: Yep, for every normal post we would do this (and more). Here, we only get the thread if we don't have it in the store already. I think a `getPostsThreads` wouldn't be any different (we've got to get each of those threads somehow, right?).

When Christopher was making the rate-limiting changes, he said the performance hit from `getProfilesAndStatusesForPosts` isn't significant, and it's got a lot more going on than `getThreadsForPosts`.  (https://github.com/mattermost/mattermost-webapp/pull/2041)

src/actions/posts.js
```diff
@@ -759,6 +760,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
...
+        if (!post.root_id) {
+            return;
+        }
+        const rootPost = Selectors.getPost(getState(), post.root_id);
+        if (post.root_id && !rootPost) {
```
lieut-data: nit: we've already checked `post.root_id` above:
```suggestion
        if (!rootPost) {
```

cpoile: d'oh! Thanks. :)

src/actions/posts.js
```diff
@@ -759,6 +760,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
...
+
+    const promises = [];
+
+    posts.forEach((post) => {
+        if (!post.root_id) {
+            return;
+        }
+        const rootPost = Selectors.getPost(getState(), post.root_id);
+        if (!rootPost) {
+            promises.push(PostActions.getPostThread(post.root_id)(dispatch, getState));
```
enahum: Nit:
```suggestion
            promises.push(dispatch(PostActions.getPostThread(post.root_id));
```

cpoile: @enahum Cool, thanks!

src/actions/posts.js
```diff
@@ -759,6 +760,28 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
     };
 }
 
+// getThreadsForPosts is intended for an array of posts that have been batched
+// (see the actions/websocket_actions/handleNewPostEvents function in the webapp)
+export function getThreadsForPosts(posts, dispatch, getState) {
```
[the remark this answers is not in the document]
cpoile: @enahum I was going off of the function just below it (`getProfilesAndStatusesForPosts`) since they're both called in the same context (see: https://github.com/mattermost/mattermost-webapp/pull/3449). Does that change your suggestion?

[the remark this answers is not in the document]
enahum: I would change it from `getThreadsForPosts(posts, myDispatch, myGetState);` to `myDispatch(getThreadsForPosts(posts));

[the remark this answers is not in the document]
cpoile: @enahum Would you replace `return Promise.all(promises)` with...?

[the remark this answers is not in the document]
cpoile: Changes made, thanks for the improvements @enahum !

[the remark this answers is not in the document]
enahum: is ok to leave the promise.all for now as we don't really use the result of it anywhere, but for the future if you need the results you can go with an `allSettle` kind of Promise resolution

src/actions/posts.test.js
```diff
@@ -1556,4 +1556,40 @@ describe('Actions.Posts', () => {
...
+                        currentUserId,
+                        statuses: {
+                            [currentUserId]: 'status',
+                        },
+                    },
+                },
+            });
+
+            it('null', async () => {
+                await Actions.getThreadsForPosts(null, dispatch, getState);
```
enahum:
```suggestion
                await dispatch(Actions.getThreadsForPosts(null));
```

src/actions/posts.test.js
```diff
@@ -1556,4 +1556,40 @@ describe('Actions.Posts', () => {
...
+            });
+
+            it('array of posts', async () => {
+                const posts = [post1, post2];
+
+                await Actions.getThreadsForPosts(posts, dispatch, getState);
```
enahum:
```suggestion
                await dispatch(Actions.getThreadsForPosts(posts));
```

cpoile:
I updated the tests and then realized they weren't really testing anything. So I rewrote them. I would appreciate your eyes on the new code, thanks!

@enahum I have one other question... The test for the webapp portion is failing: https://github.com/mattermost/mattermost-webapp/pull/3449 . I think it's because we changed `getThreadsForPosts` to be a HOF, and now the tests in the webapp fail with `Error: Actions must be plain objects. Use custom middleware for async actions.` at line 224 of `websocket_actions.test.jsx`.

I've tried adding redux-thunk to the configureStore: `const testStore = configureStore([thunk]);` but that doesn't work. I'm sorry to ask, but I've spent an hour on it and I just don't understand redux well enough yet to figure it out. How would you fix this?

enahum:
@cpoile this looks good, the issue on the webapp is probably cause you need to update the mm-redux reference in the package(-lock).json files

cpoile:
@enahum I mean the tests are failing on my local build, which is updated (using npm run dev:watch in the redux repo).

enahum:
@cpoile let me try it locally and see

src/actions/posts.js
```diff
@@ -11,6 +11,7 @@ import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} f
 import {getCustomEmojisByName as selectCustomEmojisByName} from 'selectors/entities/emojis';
 import {getConfig} from 'selectors/entities/general';
 import * as Selectors from 'selectors/entities/posts';
+import * as PostActions from 'actions/posts';
```
saturninoabril: Is this necessary?  It's importing the same file, right?

src/actions/posts.js
```diff
@@ -759,6 +760,32 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
...
+        const promises = [];
+
+        posts.forEach((post) => {
+            if (!post.root_id) {
+                return;
+            }
+
+            const rootPost = Selectors.getPost(state, post.root_id);
+            if (!rootPost) {
+                promises.push(dispatch(PostActions.getPostThread(post.root_id)));
```
saturninoabril: Can this be
```suggestion
                promises.push(dispatch(getPostThread(post.root_id)));
```

src/actions/posts.js
cpoile: good point, thanks!

cpoile:
Removing cherry-pick because this is for 5.15, not 5.14.

amyblais:
@cpoile Please cherry-pick to v5.15.

lieut-data:
(Tried to do this automatically, then realized it needs special handling re: redux.)

PR #900, opened 2019-08-19
[MM-17198] Send serverVersion to redux store on login and loadMe

mgdelacroix:
#### Summary
This PR sets the server version as part of the login complete and load me calls. Server version is needed to correctly decide which is the appropiate channel to redirect a guest user either right after login or when the user is logged in and loads the root of the server, therefore the `setServerVersion` action is dispatched as part of those two processes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17198

#### Checklist

#### Related Pull Requests
- [Has webapp changes](https://github.com/mattermost/mattermost-webapp/pull/3476)

### DIFF
--- a/src/actions/users.js
+++ b/src/actions/users.js
@@ -11,7 +11,7 @@ import {Client4} from 'client';
 import {General} from 'constants';
 import {UserTypes, TeamTypes, AdminTypes} from 'action_types';
 import {getAllCustomEmojis} from 'actions/emojis';
-import {getClientConfig} from 'actions/general';
+import {getClientConfig, setServerVersion} from 'actions/general';
 import {getMyTeams, getMyTeamMembers, getMyTeamUnreads} from 'actions/teams';
 import {loadRolesIfNeeded} from 'actions/roles';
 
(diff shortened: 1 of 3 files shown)


srkgupta:
Changes looks good. This will be tested as part of the 5.16 release Guest Accounts feature. Removing the QA Review label.

mgdelacroix:
@enahum my rationale here is that this is used in a method that belongs to `mattermost-redux` as well, `hasNewPermissions` inside `getRedirectChannelNameForTeam` here https://github.com/mattermost/mattermost-redux/blob/master/src/selectors/entities/channels.js#L1236, so chances are that the same problem would happen in any application that used `mattermost-redux`

I'm 0/5 on this approach, happy to move it to the webapp if that's preferred.

enahum:
@mgdelacroix I can speak for the mobile app that is using this https://github.com/mattermost/mattermost-redux/blob/ab9f1ff1b5c077a6a898326fcc989d4637d05e96/src/actions/websocket.js#L687 and then listening to that event to trigger the `setServerVersion` action

mgdelacroix:
@enahum same happens in the webapp, server version is set after the websocket connection. In this case, that was not enough, as the webapp is making decisions based on the server version (like channel redirection) without establishing the websocket connection first. That's why an alternative method (right after the first request, either login or load me) made sense

enahum:
@mgdelacroix ok understood, thanks for the explanation, if there isn't a better way to handle it I think is ok.

mgdelacroix:
@enahum :+1: 
Thanks for taking the time to review the approach!! :smile:

PR #903, opened 2019-08-22
[MM-17754] track if channel has been marked as unread

Willyfrog:
#### Summary
- track if a user has marked a post as unread. For that we add a new state.
- this also fixes the user marking its own posts as unread which was conflicting with preventing marking as unread when the user has the same channel open from different devices and post from one of those.

#### Ticket Link
[MM-17754](https://mattermost.atlassian.net/browse/MM-17754)

#### Checklist

#### Test Information

### DIFF
--- a/src/actions/channels.js
+++ b/src/actions/channels.js
@@ -684,7 +684,6 @@ export function deleteChannel(channelId: string): ActionFunc {
     };
 }
 
-// TODO: check this one too
 export function viewChannel(channelId: string, prevChannelId: string = ''): ActionFunc {
     return async (dispatch, getState) => {
         const {currentUserId} = getState().entities.users;
@@ -692,6 +691,9 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         const {myPreferences} = getState().entities.preferences;
         const viewTimePref = myPreferences[`${Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME}--${channelId}`];
         const viewTime = viewTimePref ? parseInt(viewTimePref.value, 10) : 0;
+        const pchanMannuallyUnread = isManuallyUnread(prevChannelId, getState);
+        const preChanId = pchanMannuallyUnread ? '' : prevChannelId;
+
         if (viewTime < new Date().getTime() - (3 * 60 * 60 * 1000)) {
             const preferences = [
                 {user_id: currentUserId, category: Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME, name: channelId, value: new Date().getTime().toString()},
@@ -700,7 +702,7 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         }
 
         try {
-            await Client4.viewMyChannel(channelId, prevChannelId);
+            await Client4.viewMyChannel(channelId, preChanId);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch, getState);
             dispatch(logError(error));
@@ -713,6 +715,12 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         const {myMembers} = getState().entities.channels;
         const member = myMembers[channelId];
         if (member) {
+            if (isManuallyUnread(channelId, getState)) {
+                actions.push({
+                    type: ChannelTypes.REMOVE_MANUALLY_UNREAD,
+                    data: {channelId},
+                });
+            }
             actions.push({
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: {...member, last_viewed_at: new Date().getTime()},
@@ -720,7 +728,7 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
             dispatch(loadRolesIfNeeded(member.roles.split(' ')));
         }
 
-        const prevMember = myMembers[prevChannelId];
+        const prevMember = myMembers[preChanId];
         if (prevMember) {
             actions.push({
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
@@ -746,11 +754,18 @@ export function markChannelAsViewed(channelId: string, prevChannelId: string = '
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: {...member, last_viewed_at: Date.now()},
             });
+            if (isManuallyUnread(channelId, getState)) {
+                actions.push({
+                    type: ChannelTypes.REMOVE_MANUALLY_UNREAD,
+                    data: {channelId},
+                });
+            }
+
             dispatch(loadRolesIfNeeded(member.roles.split(' ')));
         }
 
         const prevMember = myMembers[prevChannelId];
-        if (prevMember) {
+        if (prevMember && isManuallyUnread(prevChannelId, getState)) {
             actions.push({
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: {...prevMember, last_viewed_at: Date.now()},
@@ -1128,20 +1143,19 @@ function isManuallyUnread(channelId: string, getState: () => GlobaState): boolea
     if (channelId) {
         const initialState = getState();
         const isUnread = initialState.entities.channels.manuallyUnread[channelId];
-        return isUnread !== undefined && isUnread;
+        return typeof (isUnread) !== 'undefined' && isUnread;
     }
     return false;
-
 }
 
 export function markChannelAsRead(channelId: string, prevChannelId: ?string, updateLastViewedAt: boolean = true): ActionFunc {
     return async (dispatch, getState) => {
         const pchanMannuallyUnread = isManuallyUnread(prevChannelId, getState);
-        const preChanId = !pchanMannuallyUnread ? prevChannelId : "";
+        const preChanId = pchanMannuallyUnread ? '' : prevChannelId;
 
         // Send channel last viewed at to the server
         if (updateLastViewedAt) {
-            Client4.viewMyChannel(channelId, prechan).then().catch((error) => {
+            Client4.viewMyChannel(channelId, preChanId).then().catch((error) => {
                 forceLogoutIfNecessary(error, dispatch, getState);
                 dispatch(logError(error));
                 return {error};
@@ -1178,6 +1192,13 @@ export function markChannelAsRead(channelId: string, prevChannelId: ?string, upd
                     channelId,
                     amount: channelMember.mention_count,
                 },
+            });
+        }
+
+        if (channel && isManuallyUnread(channelId, getState)) {
+            actions.push({
+                type: ChannelTypes.REMOVE_MANUALLY_UNREAD,
+                data: {channelId},
             });
         }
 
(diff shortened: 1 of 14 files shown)


src/actions/channels.js
```diff
@@ -1122,11 +1139,23 @@ export function updateChannelPurpose(channelId: string, purpose: string): Action
     };
 }
 
+function isManuallyUnread(channelId: ?string, getState: () => GlobalState): boolean {
```
enahum: this does not look like an action, maybe consider moving it to a utils file.

src/client/client4.js
```diff
@@ -1631,6 +1631,15 @@ export default class Client4 {
         );
     };
 
+    markPostAsUnread = async (userId, postId) => {
+        this.trackEvent('api', 'api_post_set_unread_post');
+
+        return this.doFetch(
+            `${this.getUserRoute(userId)}/posts/${postId}/set_unread`,
```
enahum: can I mark a channel as unread for another user? if not maybe we should consider using `me` here ?

src/reducers/entities/channels.js
```diff
@@ -492,6 +501,21 @@ function totalCount(state = 0, action) {
     }
 }
 
+export function manuallyUnread(state = {}, action) {
+    switch (action.type) {
+    case ChannelTypes.REMOVE_MANUALLY_UNREAD: {
+        if (state[action.data.channelId]) {
+            return {...state, [action.data.channelId]: false};
```
enahum: even though I think this is fine, maybe consider removing the channel from the state

Willyfrog:
there is a separate ticket for the websocket part: https://mattermost.atlassian.net/browse/MM-17072 to be done soon.

src/actions/channels.js
```diff
@@ -690,6 +691,9 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         const {myPreferences} = getState().entities.preferences;
         const viewTimePref = myPreferences[`${Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME}--${channelId}`];
         const viewTime = viewTimePref ? parseInt(viewTimePref.value, 10) : 0;
+        const pchanMannuallyUnread = isManuallyUnread(prevChannelId, getState);
```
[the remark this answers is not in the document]
Willyfrog: that's right, I guess I got carried away by some passing getState that I saw. No real reason, so changing to receiving the state.

src/actions/channels.js
```diff
@@ -744,11 +754,18 @@ export function markChannelAsViewed(channelId: string, prevChannelId: string = '
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: {...member, last_viewed_at: Date.now()},
             });
+            if (isManuallyUnread(channelId, getState)) {
```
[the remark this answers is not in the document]
Willyfrog: no, this is actually trying to figure out if it needs to remove the unread flag.

src/actions/channels.js
Willyfrog: sounds ok! moved to helpers.

src/client/client4.js
Willyfrog: there was a discussion about it on the [api-doc PR](https://github.com/mattermost/mattermost-api-reference/pull/464#discussion_r311269544) that we might consider this at some point for admins.

src/reducers/entities/channels.js
Willyfrog: Sure!

src/actions/channels.js
```diff
@@ -13,7 +13,7 @@ import {getChannelsNameMapInTeam, getMyChannelMember as getMyChannelMemberSelect
 import {getCurrentTeamId} from 'selectors/entities/teams';
 
 import {logError} from './errors';
-import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
+import {bindClientFunc, forceLogoutIfNecessary, isManuallyUnread} from './helpers';
```
enahum: I think a better place for this function is in `utils/channel_utils`

src/actions/channels.js
```diff
@@ -711,14 +714,20 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         const {myMembers} = getState().entities.channels;
         const member = myMembers[channelId];
         if (member) {
+            if (isManuallyUnread(channelId, getState())) {
```
enahum: what if you do `const state = getState()` and then just pass the state object instead of calling the function every time?

src/actions/helpers.js
```diff
@@ -149,3 +150,11 @@ export class FormattedError extends Error {
         };
     }
 }
+
+export function isManuallyUnread(channelId: ?string, state: GlobalState): boolean {
```
enahum: can we add unit tests for this function?

src/reducers/entities/channels.js
```diff
@@ -492,6 +501,22 @@ function totalCount(state = 0, action) {
     }
 }
 
+export function manuallyUnread(state = {}, action) {
+    switch (action.type) {
+    case ChannelTypes.REMOVE_MANUALLY_UNREAD: {
+        const {[action.data.channelId]: removeThis, ...newState} = state;
```
enahum: neat, took me a couple of seconds to understand but I do like it

Willyfrog: I'll add a comment then, I don't want people struggling with 'smart code'

src/actions/channels.js
Willyfrog: I can only change it in one place ,all the others has an `await Client4.<some api call>` in the middle so there might be some changed in the state by that time (most likely not, but there is the option)

src/actions/channels.js
```diff
@@ -690,6 +690,9 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
         const {myPreferences} = getState().entities.preferences;
         const viewTimePref = myPreferences[`${Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME}--${channelId}`];
         const viewTime = viewTimePref ? parseInt(viewTimePref.value, 10) : 0;
+        const pchanMannuallyUnread = isManuallyUnread(prevChannelId, getState());
```
hmhealey: Minor nit, I'd avoid mixing naming styles for the previous channel ID to make this a bit easier to follow
```suggestion
        const prevChanManuallyUnread = isManuallyUnread(prevChannelId, getState());
```

src/actions/channels.js
```diff
@@ -690,6 +690,9 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
...
+        const preChanId = pchanMannuallyUnread ? '' : prevChannelId;
```
hmhealey: Since we can't modify `prevChannelId`, maybe we can just inline this when we call `Client4.viewMyChannel` to avoid two similarly-named variables?
```javascript
await Client4.viewMyChannel(channelId, prevChannelManuallyUnread ? '' : prevChannelId));
```

src/actions/channels.js
```diff
@@ -711,14 +714,20 @@ export function viewChannel(channelId: string, prevChannelId: string = ''): Acti
...
             actions.push({
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
                 data: {...member, last_viewed_at: new Date().getTime()},
             });
             dispatch(loadRolesIfNeeded(member.roles.split(' ')));
         }
 
-        const prevMember = myMembers[prevChannelId];
+        const prevMember = myMembers[preChanId];
         if (prevMember) {
```
hmhealey: Right now, `myMembers[preChanId]` will just be undefined if the previous channel was manually marked as unread, but that's difficult to see without following the code closely. I'd suggest that we can just be explicit about this like you're doing in `markChannelAsViewed`
```javascript
const prevMember = myMembers[prevChannelId];
if (prevMember && !prevChannelManuallyUnread) {
```

src/actions/channels.js
```diff
@@ -1124,9 +1141,12 @@ export function updateChannelPurpose(channelId: string, purpose: string): Action
 
 export function markChannelAsRead(channelId: string, prevChannelId: ?string, updateLastViewedAt: boolean = true): ActionFunc {
     return async (dispatch, getState) => {
+        const pchanMannuallyUnread = isManuallyUnread(prevChannelId, getState());
```
hmhealey: Same comment here

src/reducers/entities/channels.js
hmhealey: I'd really prefer if this was written out in a way that didn't require a comment to explain it. Even if we can understand something like this, we have to remember that we're going to have others reading this code, including community members who may be less experienced with JavaScript's quirks. It might save a few lines to do it this way compared to the more repetitive way that other reducers do it, but I don't think that's worth it for the confusion that it might cause in the future.

src/reducers/entities/channels.js
```diff
@@ -492,6 +501,23 @@ function totalCount(state = 0, action) {
...
+        if (removeThis) {
+            return newState;
+        }
+        break;
+    }
+    case ChannelTypes.POST_UNREAD_SUCCESS: {
+        return {...state, [action.data.channelId]: true};
+    }
+    }
+    return state;
```
hmhealey: Similarly, I'd recommend that we stick to using a default case in this switch statement like we do elsewhere.

src/reducers/entities/channels.js
```diff
@@ -492,6 +501,23 @@ function totalCount(state = 0, action) {
...
+        // use destructuring to both remove the element and create the new state if it exist.
+        const {[action.data.channelId]: removeThis, ...newState} = state;
+        if (removeThis) {
+            return newState;
+        }
+        break;
+    }
+    case ChannelTypes.POST_UNREAD_SUCCESS: {
+        return {...state, [action.data.channelId]: true};
+    }
```
hmhealey: This is missing a case to clear this state when the user logs out

src/actions/posts.js
```diff
@@ -364,6 +364,40 @@ export function editPost(post) {
...
+        const state = getState();
+        try {
+            unreadChan = await Client4.markPostAsUnread(userId, postId);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+        let delta;
+        if (state.entities.channels.myMembers[unreadChan.channel_id]) {
```
hmhealey: Since we're waiting on an async client call, the redux state may have changed since you called `getState`. You should move the call to happen after the `await`

src/utils/channel_utils.js
```diff
@@ -647,3 +647,11 @@ export function filterChannelsMatchingTerm(channels: Array<Channel>, term: strin
             displayName.startsWith(lowercasedTerm);
     });
 }
+
+export function isManuallyUnread(channelId: ?string, state: GlobalState): boolean {
+    if (channelId) {
+        const isUnread = state.entities.channels.manuallyUnread[channelId];
+        return typeof (isUnread) !== 'undefined' && isUnread;
```
hmhealey: I think you can replace this with
```suggestion
        return Boolean(isUnread);
```
to simplify it

src/reducers/entities/channels.js
Willyfrog: no problem, I'll change it to a delete statement, I like it less not for brevity, but for mutating stuff.

src/actions/posts.js
Willyfrog: while I did that intentionally, I now think that you're right, if something changed the number of messages we probably want the new difference, I can't recall why I thought it was better to do it before the await. If I remember I'll let you know :)

PR #904, opened 2019-08-22
MM-17643 Add the ability of keeping an oldest flag for posts chunks

sudheerDev:
#### Summary
  * Add oldest key for chunks to keep track if oldest post in channel is loaded

This solves the below issue as it was happening because of a race between react setState and redux change of posts props. We were storing the state of oldest in react state and change of posts from props for correcting scroll

We could solve the ticket without this addition but this is a good state to store. This also removes the extra calls we tend to make whenever we open a channel with less posts because we don't store the header state.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17641

#### Checklist

#### Test Information
Mac os chrome

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -747,14 +747,14 @@ export function getPostsAround(channelId, postId, perPage = Posts.POST_CHUNK_SIZ
                 ...before.order,
             ],
             next_post_id: after.next_post_id,
-            before_post_id: before.before_post_id,
+            prev_post_id: before.prev_post_id,
         };
 
         getProfilesAndStatusesForPosts(posts.posts, dispatch, getState);
 
         dispatch(batchActions([
             receivedPosts(posts),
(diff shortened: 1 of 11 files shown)


src/actions/posts.js
```diff
@@ -614,7 +616,7 @@ export function getPosts(channelId, page = 0, perPage = Posts.POST_CHUNK_SIZE) {
 
         dispatch(batchActions([
             receivedPosts(posts),
-            receivedPostsInChannel(posts, channelId, page === 0),
+            receivedPostsInChannel(posts, channelId, page === 0, posts.prev_post_id === ''),
```
migbot: Is it always the case for the first channel post to have a `prev_post_id` key with a value of `''`?

sudheerDev: @migbot Yup that is the case. If there are no posts above or below then the respective flag from API will be empty string

PR #917, opened 2019-09-05
MM-18139 - Add signature_url param when invoking install_from_url api

ali-farooq0:
#### Summary
Add `signature_url` param to `installPluginFromUrl` API.

#### Ticket Link
Webapp: https://github.com/mattermost/mattermost-webapp/pull/3600

#### Checklist

### DIFF
--- a/src/actions/admin.js
+++ b/src/actions/admin.js
@@ -465,13 +465,13 @@ export function uploadPlugin(fileData: File, force: boolean = false): ActionFunc
     };
 }
 
-export function installPluginFromUrl(url: string, signatureUrl: ?string = null, force: boolean = false): ActionFunc {
+export function installPluginFromUrl(url: string, force: boolean = false, signatureUrl: ?string = null): ActionFunc {
     return async (dispatch, getState) => {
         dispatch({type: AdminTypes.INSTALL_PLUGIN_FROM_URL_REQUEST, data: null});
 
         let data;
         try {
-            data = await Client4.installPluginFromUrl(url, signatureUrl, force);
+            data = await Client4.installPluginFromUrl(url, force, signatureUrl);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch, getState);
             dispatch(batchActions([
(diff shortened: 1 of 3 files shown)


src/actions/admin.js
```diff
@@ -465,13 +465,13 @@ export function uploadPlugin(fileData: File, force: boolean = false): ActionFunc
     };
 }
 
-export function installPluginFromUrl(url: string, force: boolean = false): ActionFunc {
+export function installPluginFromUrl(url: string, signatureUrl: ?string = null, force: boolean = false): ActionFunc {
```
marianunez: Just curious, what is the purpose of `?` in `?string`? Given that is a middle param, is it possible to completely omit?

src/actions/admin.test.js
```diff
@@ -805,23 +805,23 @@ describe('Actions.Admin', () => {
         const downloadUrl = 'testplugin.tar.gz';
         const testPlugin = {id: 'testplugin', webapp: {bundle_path: '/static/somebundle.js'}};
 
-        let urlMatch = `/plugins/install_from_url?plugin_download_url=${downloadUrl}&force=false`;
+        let urlMatch = `/plugins/install_from_url?download_url=${downloadUrl}&signature_url=&force=false`;
```
marianunez: @ali-farooq0 same comment as I mention in @iomodo server changes:

`plugin_download_url` has been released in 5.14 though. It's already documented [here](https://api.mattermost.com/#tag/plugins%2Fpaths%2F~1plugins~1install_from_url%2Fpost). This would break backward compatibility.

ali-farooq0: yep reverting it back to preserve backwards compatibility.

src/actions/admin.js
ali-farooq0: `?string` means that the parameter can be `null`. Actually with the new changes it might be possible yes.

ali-farooq0: actually i think it makes sense to allow null because it's an optional parameter.

marianunez: Thanks! I thought that it could also mean to omit the parameter.

ali-farooq0: 👌 any parameter that has a default value can be omitted.

PR #918, opened 2019-09-06
[MM-17068] add websocket support for mark as unread

Willyfrog:
#### Summary

Websocket actions and events to ensure consistency of mark as unread feature across devices

#### Ticket Link
[MM-17068](https://mattermost.atlassian.net/browse/MM-17068)

#### Checklist

#### Test Information
This PR was tested on: [Mac OS webapp with chrome] 

[MM-17066]: https://mattermost.atlassian.net/browse/MM-17066

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -368,12 +368,8 @@ export function editPost(post) {
 }
 
 export function getUnreadPostData(unreadChan, state) {
-    let delta;
-    if (state.entities.channels.myMembers[unreadChan.channel_id]) {
-        delta = state.entities.channels.myMembers[unreadChan.channel_id].msg_count - unreadChan.msg_count;
-    } else {
-        delta = unreadChan.msg_count;
-    }
+    const member = getMyChannelMemberSelector(state, unreadChan.channel_id);
+    const delta = member ? member.msg_count - unreadChan.msg_count : unreadChan.msg_count;
 
     const data = {
         teamId: unreadChan.team_id,
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -364,6 +364,26 @@ export function editPost(post) {
     });
 }
 
+export function getUnreadPostData(unreadChan, state) {
+    let delta;
+    if (state.entities.channels.myMembers[unreadChan.channel_id]) {
+        delta = state.entities.channels.myMembers[unreadChan.channel_id].msg_count - unreadChan.msg_count;
+    } else {
+        delta = unreadChan.msg_count;
+    }
+
+    const data = {
+        teamId: unreadChan.team_id,
+        channelId: unreadChan.channel_id,
+        msgCount: unreadChan.msg_count,
+        mentionCount: unreadChan.mention_count,
+        lastViewedAt: unreadChan.last_viewed_at,
+        deltaMsgs: delta,
+    };
+
+    return data;
+}
+
 export function setUnreadPost(userId, postId) {
     return async (dispatch, getState) => {
         let unreadChan;
@@ -374,23 +394,9 @@ export function setUnreadPost(userId, postId) {
             dispatch(logError(error));
             return {error};
         }
-        let delta;
 
         const state = getState();
-        if (state.entities.channels.myMembers[unreadChan.channel_id]) {
-            delta = state.entities.channels.myMembers[unreadChan.channel_id].msg_count - unreadChan.msg_count;
-        } else {
-            delta = unreadChan.msg_count;
-        }
-
-        const data = {
-            teamId: unreadChan.team_id,
-            channelId: unreadChan.channel_id,
-            msgCount: unreadChan.msg_count,
-            mentionCount: unreadChan.mention_count,
-            lastViewedAt: unreadChan.last_viewed_at,
-            deltaMsgs: delta,
-        };
+        const data = getUnreadPostData(unreadChan, state);
         dispatch({
             type: ChannelTypes.POST_UNREAD_SUCCESS,
             data,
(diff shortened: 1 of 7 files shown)


src/actions/posts.js
```diff
@@ -364,6 +364,26 @@ export function editPost(post) {
     });
 }
 
+export function getUnreadPostData(unreadChan, state) {
```
hmhealey: Perhaps we can turn this into an action creator instead of just having a function that constructs the data portion of the action? It's dispatched with the same type in both places, so it might be cleaner to do it that way. That would also avoid passing around `state` since we'd just get it from thunk

src/reducers/entities/channels.js
```diff
@@ -306,11 +306,8 @@ function myMembers(state = {}, action) {
     }
     case ChannelTypes.POST_UNREAD_SUCCESS: {
         const data = action.data;
-        const channelState = state[data.channelId];
+        const channelState = state[data.channelId] || {};
```
hmhealey: This isn't going to cause any change in behaviour is it? I'm a bit worried this might mean that we're going to end up with a partial channel member object that's missing things like roles or channel-specific settings.

src/reducers/entities/teams.js
```diff
@@ -210,16 +210,13 @@ function myMembers(state = {}, action) {
     }
 
     case ChannelTypes.POST_UNREAD_SUCCESS: {
-        const {teamId, deltaMsgs, mentionCount} = action.data;
-        const teamState = state[teamId];
-        if (!teamState) {
-            return state;
-        }
+        const {teamId, deltaMsgs, mentionCount, msgCount} = action.data;
+        const teamState = (typeof state[teamId] === 'undefined' ? {} : state[teamId]);
```
hmhealey: Same comment here as with the channel member

src/reducers/entities/teams.js
```diff
@@ -210,16 +210,13 @@ function myMembers(state = {}, action) {
...
 
         const newTeamState = {
             ...teamState,
-            msg_count: teamState.msg_count - deltaMsgs,
-            mention_count: teamState.mention_count + mentionCount,
+            msg_count: (typeof teamState.msg_count === 'undefined' ? msgCount : teamState.msg_count - deltaMsgs),
```
hmhealey: Perhaps it would actually be cleaner to make two separate actions so that you can identify what fields should exist on each one? It now seems like we have two types of `POST_UNREAD_SUCCESS` actions with different meanings that contain different fields

src/actions/posts.js
```diff
@@ -364,6 +364,26 @@ export function editPost(post) {
...
+    let delta;
+    if (state.entities.channels.myMembers[unreadChan.channel_id]) {
```
enahum: I think we have a selector for this, not sure and 0/5 in using it

src/reducers/entities/channels.js
enahum: Yes, I would prefer to check for null instead of having it return an empty object

src/actions/posts.js
Willyfrog: having a function that only generates the data portion makes it pure, as there is no side effects, easier to maintain and test if felt necessary.
having it dispatch the action, wouldn't prevent it to get the state, as the only thing this function does is to extract info from the state, so the only way to remove it is to have the caller function extract it, but then, all the redundancy saved was back in the function, so I'm not really sure what are we gaining with that change?
can we discuss this? since both of you see a benefit I don't I'd like to dig deeper to find what I'm probably missing (or maybe its just that I'm too used to functional programing 😆 )

src/reducers/entities/channels.js
Willyfrog: 0/5 on this, so I'm changing it

src/reducers/entities/teams.js
Willyfrog: I see it the other way around, I have 1 action with different effects on different sections on the state, not sure how it would be more clear having multiple actions, as you then would have to track all the info.

Another option would be to switch to something closer to `sagas` where each action would only have a few side effects, but we would then have to rethink the whole repo in a way that matches that architecture.

hmhealey: Does the action always contain both fields? Since you're null checking it, I was thinking that it contains either one field or the other. If the action always contains both, then I think we're fine to keep using one action, but then we could get rid of the null checking to make it clearer

Willyfrog: it should contain both, but running the tests gave me a bit of a headache and sometimes there were missing elements, so I checked for them and choose a default value, which might not make sense (I'm not happy with default for msg_count as I'm not sure it won't casuse trouble, I'm fine with mentions).

src/actions/posts.js
hmhealey: The main reason I was thinking it might be better to turn it into an action creator is that the caller wouldn't have to care as much about the arguments. For example, these two action creators are both called the same way, but one requires the state and the other doesn't:

```javascript
const myAction1 = (value) => ({type: 'MY_ACTION', value});
const myAction2 = (value) => (dispatch, getState) => {
    const otherValue = getState().otherValue;
    dispatch({type: 'MY_ACTION', value, otherValue);
};

dispatch(myAction1(123));
dispatch(myAction2(123));
```

I agree that it's kind of unnecessary to do that if you're wanting to keep it really pure though. In that case, we could make it so that instead of just returning the data portion of the action, it returns the entire object including the `type` field since it's the same in both places.

I'm just 0/5 on those either way though

src/reducers/entities/teams.js
hmhealey: Ah, I see. I'd lean towards making the tests a bit worse to make the rest of the code better, but it's your call then.

PR #920, opened 2019-09-10
MM-17550 Added support for search term in get marketplace plugins

marianunez:
#### Summary
- Added search param for `getMarketplacePlugins` api endpoint. 
- Added missing test for failure case of `getMarketplacePlugins`

#### Ticket Link
[MM-17550](https://mattermost.atlassian.net/browse/MM-17550)

#### Related PR
- [Webapp PR](https://github.com/mattermost/mattermost-webapp/pull/3635)

#### Checklist

### DIFF
--- a/src/actions/plugins.js
+++ b/src/actions/plugins.js
@@ -6,7 +6,7 @@ import {bindClientFunc} from './helpers';
 
 import {PluginTypes} from 'action_types';
 
-export function getMarketplacePlugins(filter = '') {
+export function getMarketplacePlugins(filter) {
     return bindClientFunc({
         clientFunc: Client4.getMarketplacePlugins,
         onSuccess: PluginTypes.RECEIVED_MARKETPLACE_PLUGINS,
--- a/src/actions/plugins.js
+++ b/src/actions/plugins.js
@@ -6,11 +6,14 @@ import {bindClientFunc} from './helpers';
 
 import {PluginTypes} from 'action_types';
 
-export function getMarketplacePlugins() {
+export function getMarketplacePlugins(searchTerm) {
     return bindClientFunc({
         clientFunc: Client4.getMarketplacePlugins,
         onSuccess: PluginTypes.RECEIVED_MARKETPLACE_PLUGINS,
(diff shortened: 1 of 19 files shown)


src/actions/plugins.js
```diff
@@ -6,11 +6,14 @@ import {bindClientFunc} from './helpers';
 
 import {PluginTypes} from 'action_types';
 
-export function getMarketplacePlugins() {
+export function getMarketplacePlugins(searchTerm) {
```
lieut-data: Not sure what pattern we've converged on for actions, but would it make sense to destructure with a default here in anticipation of future parameters? 
```suggestion
export function getMarketplacePlugins({searchTerm: ''}) {
```

src/client/client4.js
```diff
@@ -2641,9 +2641,9 @@ export default class Client4 {
         );
     };
 
-    getMarketplacePlugins = async () => {
+    getMarketplacePlugins = async (searchTerm) => {
```
lieut-data: Should we use the term `filter` vs. `searchTerm` for consistency with the API?

marianunez: Done

src/actions/plugins.js
marianunez: Most actions in this repo pass the parameters separately so I kept that convention but I did add the default there to be cleaner.

src/actions/plugins.js
```diff
@@ -6,11 +6,14 @@ import {bindClientFunc} from './helpers';
 
 import {PluginTypes} from 'action_types';
 
-export function getMarketplacePlugins() {
+export function getMarketplacePlugins(filter = '') {
```
ali-farooq0: better to leave this as is and fix the `null` issue where we build the query params

marianunez: Done

hanzei:
@prapti Do you want to do QA for this PR pre or post merge?

marianunez:
Merging - this would be tested in the feature branch as discussed with @prapti

PR #924, opened 2019-09-12
MM-17023: Plugin Marketplace

lieut-data:
#### Summary
Merging the feature branch for the plugins marketplace MVP.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17023

#### Test Information
This PR was tested on: Chrome

### DIFF
--- a/src/reducers/entities/plugins.js
+++ b/src/reducers/entities/plugins.js
@@ -16,7 +16,7 @@ function marketplacePlugins(state = getInitialState(), action) {
     case PluginTypes.RECEIVED_MARKETPLACE_PLUGINS: {
         return {
             ...state,
-            data: action.data,
+            data: action.data ? action.data : [],
             serverError: null,
         };
     }
--- a/src/reducers/entities/plugins.js
+++ b/src/reducers/entities/plugins.js
@@ -15,14 +15,12 @@ function marketplacePlugins(state = getInitialState(), action) {
     switch (action.type) {
     case PluginTypes.RECEIVED_MARKETPLACE_PLUGINS: {
         return {
-            ...state,
             data: action.data ? action.data : [],
             serverError: null,
         };
     }
     case PluginTypes.GET_MARKETPLACE_PLUGINS_FAILURE: {
         return {
-            ...state,
             data: [],
             serverError: action.error,
         };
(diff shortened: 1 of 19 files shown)


lieut-data:
Thanks for offering to help out, @jwilander!

src/reducers/entities/plugins.js
```diff
@@ -0,0 +1,37 @@
...
+    };
+}
+
+function marketplacePlugins(state = getInitialState(), action) {
+    switch (action.type) {
+    case PluginTypes.RECEIVED_MARKETPLACE_PLUGINS: {
+        return {
+            ...state,
+            data: action.data ? action.data : [],
+            serverError: null,
```
jwilander: Not sure if we've stored the server error like this in the redux store before. Generally we just have the actions return the error and the component that called the action stores it in state. Not a big deal, and I'm 0/5, but thought I'd comment on the pattern difference.

src/reducers/entities/plugins.js
```diff
@@ -0,0 +1,37 @@
...
+        data: [],
+        serverError: null,
+    };
+}
+
+function marketplacePlugins(state = getInitialState(), action) {
+    switch (action.type) {
+    case PluginTypes.RECEIVED_MARKETPLACE_PLUGINS: {
+        return {
+            ...state,
```
jwilander: nit: Seems unnecessary as all the state's fields are overridden anyways

marianunez: Fixed

src/reducers/entities/plugins.js
marianunez: It is true that there are not a lot of cases that actually handle the [`onFailure`](https://github.com/mattermost/mattermost-redux/blob/ae8adedd4808d4d5f37f3be4f8d57767f28f0de6/src/actions/plugins.js#L13) case of the client function. However, there are a couple like this one: https://github.com/mattermost/mattermost-redux/blob/f80b0932d958a554e8eda8904573717b3b0ee48f/src/reducers/requests/users.js#L23 that do return the error in the state.

It gets kinda tricky to handle the return error directly from the call when using this pattern of the [`bindClientFunc`](https://github.com/mattermost/mattermost-redux/blob/ae8adedd4808d4d5f37f3be4f8d57767f28f0de6/src/actions/plugins.js#L10) which is what I was trying to follow as most of the actions in the repo do.

jwilander: Thread resolved on the webapp PR

PR #930, opened 2019-09-19
[MM-17624] Update message count and unread count in redux on createPost

devinbinnie:
#### Summary
When a new post is created by the current user, the `total_msg_count` field of the channel and `msg_count` field of the channel memberships weren't being updated. Most of the app was not affected by this, since both were updating on channel navigation, but autocomplete was only updating the channels in the redux state, causing MM-17624. This PR ensures the message counts and unreads are in sync when new posts are created by the current user.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17624

#### Checklist

Wasn't able to create tests due to the complexity of mocking the redux store/the complexity of `redux-offline` and the `createPost()` function.

#### Test Information
This PR was tested on: Chrome 76, macOS 10.14.6

### DIFF
--- a/src/actions/posts.js
+++ b/src/actions/posts.js
@@ -5,7 +5,7 @@ import {batchActions} from 'redux-batched-actions';
 
 import {Client4} from 'client';
 import {General, Preferences, Posts, WebsocketEvents} from 'constants';
-import {PostTypes, FileTypes, IntegrationTypes} from 'action_types';
+import {PostTypes, ChannelTypes, FileTypes, IntegrationTypes} from 'action_types';
 
 import {getCurrentChannelId, getMyChannelMember as getMyChannelMemberSelector} from 'selectors/entities/channels';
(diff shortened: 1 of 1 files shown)


devinbinnie:
@enahum `markChannelAsUnread()` is called by the websocket event, which increments `total_msg_count` for the channel. Once the user opens the channel, the `msg_count` of their membership is incremented by `markChannelAsRead()`. When it's your own message, neither of these events trigger and thus if either of them are updated without the other, the values are out of sync and the channel looks unread. So we have to update the values manually without marking anything as read/unread.

enahum:
Thanks for explaining!

devinbinnie:
@saturninoabril Thanks! As mentioned in the description, there's a lot of complexity to mocking `redux-offline` and I don't believe it's worth doing. Unless I'm missing something.

PR #938, opened 2019-10-04
MM-18796 - Add POST `/plugins/marketplace` API to redux

ali-farooq0:
#### Summary
Add POST `/plugins/marketplace` API to redux

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-18796


src/client/client4.js
```diff
@@ -2661,6 +2661,15 @@ export default class Client4 {
         );
     }
 
+    installMarketplacePlugin = async (id, version) => {
+        this.trackEvent('api', 'api_install_marketplace_plugin');
```
marianunez: I believe we are already tracking this on the webapp [here](https://github.com/mattermost/mattermost-webapp/blob/23648a1d3cb8c3f31adcc70a37056111afcc9f24/components/plugin_marketplace/marketplace_item/marketplace_item.jsx#L88)

src/reducers/requests/plugins.js
```diff
@@ -0,0 +1,26 @@
...
+
+import {combineReducers} from 'redux';
+import {PluginTypes} from 'action_types';
+
+import {handleRequest, initialRequestState} from './helpers';
+
+import type {GenericAction} from '../../types/actions';
+import type {RequestStatusType, PluginsRequestsStatuses} from '../../types/requests';
+
+function installMarketplacePlugin(state: RequestStatusType = initialRequestState(), action: GenericAction): RequestStatusType {
```
marianunez: Do we need a reducer and store this status in the store at all? The component is just making the call [directly](https://github.com/mattermost/mattermost-webapp/blob/23648a1d3cb8c3f31adcc70a37056111afcc9f24/components/plugin_marketplace/marketplace_item/marketplace_item.jsx#L59), receiving the error back directly and it never references anything in the store.

src/actions/plugins.js
```diff
@@ -17,3 +17,15 @@ export function getMarketplacePlugins(filter) {
     });
 }
 
+export function installMarketplacePlugin(id, version) {
```
marianunez: We wouldn't need this action either. The webapp would reference the new Client4 function direclty.

src/client/client4.js
ali-farooq0: true. However if someone uses this directly by passing the webapp we'll be missing this event. But don't mind removing it :P

src/reducers/requests/plugins.js
ali-farooq0: it's a tiny bit nicer to have the reducer as it has some built in functionality to logout https://github.com/mattermost/mattermost-redux/blob/ef51b0056f5aa6613827e1475cde1339320aeef2/src/actions/helpers.js#L88. Also allows requests to be cancelled if necessary which is a similar problem we ran into when searching through the get API

src/client/client4.js
marianunez: Agreed. We should remove the one in the webapp then.

src/reducers/requests/plugins.js
marianunez: In that case you can still keep the action call which is the one that uses `bindClientFunc` where you have that built in functionality to logout. 

However, I still do not see the value of having this reducer that `handleRequest` and manages the state for the status returned from the client call.

src/client/client4.js
ali-farooq0: @marianunez 
I'm personally OK keeping both, because once 5.16 is released, our analytics tool will contain this event. However >= 5.18 it'll be the other redux event, it might be hard to put together that data if you need to see how many people installed a marketplace plugin. My suggestion would to keep both now. If this change was going into 5.16 then it would've been fine removing that event.

marianunez: There still needs to be a change to add the event to the whitelist [here](https://github.com/mattermost/mattermost-redux/blob/e40072b407f8d437cb10a2edc78ac7972c7587be/src/client/client4.js#L2946) so that it can actually get logged - given that most events right now are bypassed to reduce usage in segment. 

@aaronrothschild could you confirm if we should include this event to the whitelist to get tracked?

ali-farooq0: Ah i see. In that case I shall remove.

marianunez: @ali-farooq0 Which one are you referring that will be removed?

ali-farooq0: @marianunez the webapp one

marianunez: Gotcha, well the whitelist would need to be updated to include your new API event `api_install_marketplace_plugin` instead of [this one](https://github.com/mattermost/mattermost-redux/blob/e40072b407f8d437cb10a2edc78ac7972c7587be/src/client/client4.js#L2942)

ali-farooq0: ah. forgot about that as well. Good catch! Maybe i should revert my other webapp change to remove the tracking there? or leave it the way it is right now?

marianunez: 0/5 either way. We just need to make sure that whichever we keep we add to the whitelist so that it actually gets to segment:

`ui_marketplace_download` and/or
`api_install_marketplace_plugin`

@aaronrothschild thoughts?

ali-farooq0: @marianunez shall we unblock this PR and I can update the tracking events once I get the confirmation from Aaron.

I've created a jira to keep track and I can follow up with Aaron later, does that work?
https://mattermost.atlassian.net/browse/MM-19318

marianunez: Sure, we just would need to add the new event in the whitelist then because now with `ui_marketplace_download` we are not tracking either at this point🤔

ali-farooq0: ah i didn't realize that ui event was whitelisted :(. I've reverted my change in webapp. https://github.com/mattermost/mattermost-webapp/pull/3871/commits/0a5ec12ca7b84b080c8eb42953a94ccd5e4e3c27

ali-farooq0:
@iomodo * **poke** *

ali-farooq0:
oo so fast :P

iomodo:
too slow :D

PR #945, opened 2019-10-17
MM-18356: Update for newly-added ability to paginate channel searches.

mkraft:
#### Summary

Adds ability to paginate channel search results.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-18356

#### Checklist

#### Test Information
Chrome browser, macOS


src/actions/channels.ts
```diff
@@ -1006,7 +1008,7 @@ export function searchAllChannels(term: string, notAssociatedToGroup = '', exclu
             },
         ]), getState);
 
-        return {data: channels};
+        return {data: response};
```
streamer45: Why do we use `response` here instead of `channels`?

mkraft: It is still an array of channels for non-paginated searches, however for paginated searches it is returned as an object like this:
```js
{
  channels: [],
  total_count: 0
}
```

streamer45: Yes, I guess my question was more about the reason behind breaking consistency and returning two different structures. Isn't `total_count` equal to `channels.length`?

mkraft: `total_count` is the total number of channels that match the query regardless of what page of results are being requested and regardless of the number of results per page.

In terms of your consistency question we have two options that I can see:

1. Make a new action so that there is one action for paginated queries and another for non-paginated queries. 
2. Have one action that returns a different structure depending on whether the query is requested paginated or not.

I don't really have a preference so I went with the one that _seemed_ to have less code to write (option 2). That said I'm totally open to option 1.

streamer45: That makes sense. Thank you for clarifying @mkraft :+1:

srkgupta:
Hi @mkraft 
Quick dumb question. The ticket MM-18356 is for including the search and pagination on the System Console. Not sure why it would affect the RN app?

mkraft:
> Hi @mkraft
> Quick dumb question. The ticket MM-18356 is for including the search and pagination on the System Console. Not sure why it would affect the RN app?

Is it affecting it somehow that I'm not aware of?

src/actions/channels.ts
```diff
@@ -996,7 +996,7 @@ export function searchAllChannels(term: string, notAssociatedToGroup = '', exclu
             return {error};
         }
 
-        const channels = paginate ? response.channels : response;
+        const channels = (page == null && perPage == null) ? response : response.channels;
```
streamer45: Just wondering why we are testing against `null`. From my understanding those values will be either `undefined` or some number, is that right?  
Also, any reasons against testing the result (e.g. check if `response.channels` is defined) instead of the input?

mkraft: I think it's better to compare against the API response rather than the function parameter values. Thanks for the idea I have made the change.

PR #946, opened 2019-10-18
MM-19507 - Reactions on Deleted Messages

marianunez:
#### Summary
Added a guard condition to avoid storing reactions in the state for deleted posts.   This was seen in the related bug when switching channels after a message was deleted and the `RECEIVED_POSTS` action was triggering `storeReactionsForPost` for all posts.    

#### Ticket Link
[MM-19507](https://mattermost.atlassian.net/browse/MM-19507)

#### Checklist

#### Test Information
This PR was tested on: Chrome-Mac

### DIFF
--- a/src/reducers/entities/posts.ts
+++ b/src/reducers/entities/posts.ts
@@ -1052,7 +1052,7 @@ export function reactions(state = {}, action) {
 }
 
 function storeReactionsForPost(state, post) {
-    if (!post.metadata || !post.metadata.reactions) {
+    if (!post.metadata || !post.metadata.reactions || post.delete_at > 0) {
         return state;
     }
 
--- a/src/reducers/entities/posts.test.js
+++ b/src/reducers/entities/posts.test.js
@@ -3488,6 +3488,21 @@ describe('reactions', () => {
                     },
                 });
             });
+
+            it('should not save reaction for a deleted post', () => {
+                const state = deepFreeze({});
+                const action = {
+                    type: actionType,
+                    data: {
+                        id: 'post',
+                        delete_at: '1571366424287',
+                    },
+                };
+
+                const nextState = reducers.reactions(state, action);
+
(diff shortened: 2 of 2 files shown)


marianunez:
> LGTM, but shouldn't we also prevent mounting the `<Reactions>` component if the post is in a deleted state?

We could add a condition to get the `post.has_reactions` (which is cleared if its a deleted post) here to not show the `ReactionList` component [here](https://github.com/mattermost/mattermost-webapp/blob/ae6b20c09570648317605b15e951c480e263ef1c/components/post_view/post_body/post_body.jsx#L207) . However, so far it has been handled at the redux level adding/removing reactions when the post is deleted so no condition was needed.

src/reducers/entities/posts.ts
```diff
@@ -1052,7 +1052,7 @@ export function reactions(state = {}, action) {
 }
 
 function storeReactionsForPost(state, post) {
-    if (!post.metadata || !post.metadata.reactions) {
+    if (!post.metadata || !post.metadata.reactions || post.delete_at > 0) {
         return state;
```
enahum: what happens if the post already has reactions, isn't this `state` already including them?

marianunez: This is called when the channel is refreshed and `RECEIVED_POSTS/RECEIVED_POST` is triggered so the deleted message was previously removed (and corresponding reactions) when the channel was changed and is coming back in again with the reactions that should not be added. 

When you are in the channel and the message is deleted, the reaction is correctly removed at the time from the state.

lindy65:
Thanks @marianunez - this can be tested by QA post-merge 👍

src/reducers/entities/posts.ts
hmhealey: It's strange that the server is sending this information for deleted posts. I'm going to file a ticket to look into metadata being provided for deleted posts

PR #951, opened 2019-10-22
Adding new action type REMOVE_MEMBER_FROM_CHANNEL

jespino:
#### Summary
Adding new action type `REMOVE_MEMBER_FROM_CHANNEL` (used to complete the `LEAVE_TEAM` websocket event handling properly).

Before this change the user code wasn't able to remove a channel membership when a user leave the team.

Needed for mattermost/mattermost-webapp#4022

#### Ticket Link
[MM-19414](https://mattermost.atlassian.net/browse/MM-19414)

#### Checklist

### DIFF
--- a/src/reducers/entities/channels.ts
+++ b/src/reducers/entities/channels.ts
@@ -310,7 +310,7 @@ function myMembers(state = {}, action) {
     }
 }
 
-function membersInChannel(state = {}, action) {
+export function membersInChannel(state = {}, action) {
     switch (action.type) {
     case ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER:
     case ChannelTypes.RECEIVED_CHANNEL_MEMBER: {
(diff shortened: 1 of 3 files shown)


src/reducers/entities/channels.ts
```diff
@@ -310,7 +310,7 @@ function myMembers(state = {}, action) {
     }
 }
 
-function membersInChannel(state = {}, action) {
+export function membersInChannel(state = {}, action) {
```
ali-farooq0: Exporting functions that are not being publicly used and is only being exported for testing is considered an anti-pattern. 
https://medium.com/@macsikora/how-to-test-private-functions-of-es6-module-fb8c1345b25f

Perhaps you could import the reducer in the test and invoke that?

kinda like 
```
import reducer from 'reducers/entities/channels';
...
reducer({}, action);

```

jespino: Yes, I agree, now is done. I only changed my implementation because this is a PR that is going to be cherry-picked so I prefer to not generate too much complexity.

ali-farooq0: I think testing using the `reducer` would've been less work because you don't need to export the function, but I can understand that it might increase risk or unknowns, because it's not being done that way currently.

PR #953, opened 2019-10-27
MM-18357: Adds ability to retrieve team search results paginated.

mkraft:
#### Summary

Allows team searches to be paginated.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-18357

#### Checklist

#### Test Information

Chrome browser version 78.0.3904.70 on macOS.

#### Dependent pull requests

* https://github.com/mattermost/mattermost-server/pull/12910


src/actions/teams.ts
```diff
@@ -138,16 +138,36 @@ export function getTeams(page = 0, perPage: number = General.TEAMS_CHUNK_SIZE, i
...
+            {
+                type: TeamTypes.RECEIVED_TEAMS_LIST,
+                data: teams,
+            },
+            {
+                type: TeamTypes.GET_TEAMS_SUCCESS,
+            },
+        ]), getState);
+
+        return {data: response};
```
migbot: Do we want to return `teams` here instead?

mkraft: In the case that this is a non-paginated request `response` will have the same value as `teams`. For a paginated request it will look like:
```json
{"teams": [], "total_count": 0}
```

migbot: OK cool, so we'll expect the caller handle the return object according to whether they paginated or not.

mkraft: Yeah that's how we handled it in another spot too rather than creating two separate actions.

PR #960, opened 2019-10-31
Downgrading typescript version

jespino:
#### Summary
Downgrading typescript version, this is needed until we can figure out how to
make `"declaration": true` works in the last version of typescript.


jespino:
This doesn't fix the problem properly.

PR #965, opened 2019-11-04
[MM-19841] Use action.data or baseState

migbot:
#### Summary
The [comment](https://github.com/mattermost/mattermost-redux/blob/e77647700ad23d0b1d0c98aa6b48f3330c425e4f/src/store/helpers.ts#L30-L32) above the `offlineReducer` says that the data property will be used as the new state but `baseState` was always being used. I've set it to used `action.data` and fallback to `baseState`. On the [mobile app](https://github.com/mattermost/mattermost-mobile/pull/3519) I need to be able to pass in a state through `action.data` in order to preserve some needed user preferences when the state is reset.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19841

#### Checklist

#### Test Information
This PR was tested on:
* iOS simulator, 13.1

### DIFF
--- a/src/store/helpers.ts
+++ b/src/store/helpers.ts
@@ -32,7 +32,7 @@ export function createReducer(baseState, ...reducers) {
     // as the new state.
     function offlineReducer(state = {}, action) {
         if (action.type === General.OFFLINE_STORE_RESET) {
-            return baseReducer(baseState, action);
+            return baseReducer(action.data || baseState, action);
         }
 
         return baseReducer(state, action);


enahum:
@migbot I think what you want to do here and in https://github.com/mattermost/mattermost-mobile/pull/3519 can be achieved by adding the currentTeamId and the team theme preferences here in [this function](https://github.com/mattermost/mattermost-mobile/blob/master/app/store/middleware.js#L201) without changing the reducer or anything else

PR #967, opened 2019-11-05
[MM-19533] unread message count selector

Willyfrog:
#### Summary
Selector to count the number of messages between membership and channel.

#### Ticket Link
[MM-19533](https://mattermost.atlassian.net/browse/MM-19533)


hmhealey:
Is it enough to check this value? I think this value gets cleared at odd times (like when the client sends the websocket event to read a channel), so it might not work quite the way you've expected

Willyfrog:
weems like I didn't notice your comment, sorry about that. This is only needed to know if the channel is unread when the user lands into it, having the ammount of messages was trying to get the initial number alread, but seems like it gets marked as read quite quickly

src/selectors/entities/channels.ts
```diff
@@ -172,6 +172,12 @@ export function isChannelReadOnly(state: GlobalState, channel: Channel): boolean
 export function shouldHideDefaultChannel(state: GlobalState, channel: Channel): boolean {
     return channel && channel.name === General.DEFAULT_CHANNEL && !isCurrentUserSystemAdmin(state) && getConfig(state).ExperimentalHideTownSquareinLHS === 'true';
 }
+export const countCurrentChannelUnreadMessages: (a: GlobalState) => number = createSelector(getCurrentChannel, getMyCurrentChannelMembership, (channel: Channel, membership?: ChannelMembership | null): number => {
+    if (!membership) {
```
sudheerDev: When does this happen? Would we be requesting this value if user is not a member? or Is it just a defensive check?

Willyfrog: it is a defensive check

PR #968, opened 2019-11-05
MM-19482 Stop rendering New Messages line differently for manually unread channels

hmhealey:
As discussed, because we can't properly save that `isManuallyUnread` state between sessions, we're going to remove this special behaviour when using the Mark as Unread feature.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19482

#### Related PRs
Mobile: https://github.com/mattermost/mattermost-mobile/pull/3524
Web: https://github.com/mattermost/mattermost-webapp/pull/4138

### DIFF
--- a/src/utils/post_list.ts
+++ b/src/utils/post_list.ts
@@ -45,8 +45,7 @@ export function makeFilterPostsAndAddSeparators() {
         getCurrentUser,
         shouldShowJoinLeaveMessages,
         isTimezoneEnabled,
-        (state, {channelId}) => state.entities.channels.manuallyUnread[channelId],
-        (posts, lastViewedAt, indicateNewMessages, selectedPostId, currentUser, showJoinLeave, timeZoneEnabled, isManuallyUnread) => {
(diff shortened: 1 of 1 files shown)


deanwhillier:
Quick question: The ticket reads to me like it will never be possible to 'mark as unread' above the current users last post. Is that correct?

For example, if you had something like:

1. John Smith (me) 4:30pm: Lorem Ipsum
2. Jane Doe 4:33pm: Sed sollicitudin porta maximus
3. John Smith (me) 4:35pm: Proin venenatis ex risus
4. Jane Doe 4:38pm: Donec pulvinar, turpis in placerat sodales

Could John Smith set line #2 to unread and have the 'new messages' line appear between line #1 and #2? Or would the 'new messages' line jump down to between line #3 and #4 because John Smith's last post is line #3?

hmhealey:
If you mark your own post as unread, it'll move the New Messages line down to the next message made by another user, so in your example, marking the first or second post would put the New Messages above the second one. if you mark the third or fourth messages as unread, it puts the line above the fourth post.

PR #970, opened 2019-11-06
Removing implicit anys from team selectors and reducers

jespino:
#### Summary
Removing implicit anys from team selectors and reducers

### DIFF
--- a/src/reducers/entities/teams.ts
+++ b/src/reducers/entities/teams.ts
@@ -3,8 +3,12 @@
 import {combineReducers} from 'redux';
 import {ChannelTypes, TeamTypes, UserTypes, SchemeTypes, GroupTypes} from 'action_types';
 import {teamListToMap} from 'utils/team_utils';
+import {Team, TeamMembership} from 'types/teams';
+import {UserProfile} from 'types/users';
+import {RelationOneToOne, IDMappedObjects} from 'types/utilities';
+import {GenericAction} from 'types/actions';
 
-function currentTeamId(state = '', action) {
+function currentTeamId(state = '', action: GenericAction) {
     switch (action.type) {
     case TeamTypes.SELECT_TEAM:
         return action.data;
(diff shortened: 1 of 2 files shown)


src/reducers/entities/teams.ts
```diff
@@ -204,7 +208,7 @@ function myMembers(state = {}, action) {
         return nextState;
     }
     case TeamTypes.UPDATED_TEAM_MEMBER_SCHEME_ROLES: {
-        return updateTeamMemberSchemeRoles(state, action);
+        return updateMyTeamMemberSchemeRoles(state, action);
```
jespino: I think this was a bug before, because the state type wasn't the expected one.

src/reducers/entities/teams.ts
```diff
@@ -349,15 +353,15 @@ function groupsAssociatedToTeam(state = {}, action) {
         for (const group of groups) {
             associatedGroupIDs.delete(group.id);
         }
-        nextState[teamID] = Array.from(associatedGroupIDs);
+        nextState[teamID] = {ids: Array.from(associatedGroupIDs), totalCount: associatedGroupIDs.size};
```
jespino: This looks like a bug returning the state, @mkraft what do you think?

src/selectors/entities/teams.ts
```diff
@@ -48,15 +49,15 @@ export const getCurrentTeam = reselect.createSelector(
...
+export function getTeam(state: GlobalState, id: string): Team {
     const teams = getTeams(state);
     return teams[id];
 }
 
 export const getCurrentTeamMembership = reselect.createSelector(
     getCurrentTeamId,
     getTeamMemberships,
-    (currentTeamId, teamMemberships) => {
+    (currentTeamId: string, teamMemberships: {[x: string]: TeamMembership}): TeamMembership => {
```
reflog: 0/5 on this, just FYI I added a type called Dictionary<T> which defines `{[x: string]: T}`

src/reducers/entities/teams.ts
mkraft: Yes it looks like a bug to me too just not reflected in the UX because of the way it's being used.

PR #973, opened 2019-11-09
[MM-19309] Name Display Lock

hahmadia:
#### Summary
These are changes made for https://github.com/mattermost/mattermost-webapp/pull/4148. It updates the `makeGetDisplayName` function to decide whether we should return name display from admin settings or user preference settings.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19309

#### Checklist


src/selectors/entities/users.test.js
```diff
@@ -443,35 +443,106 @@ describe('Selectors.Users', () => {
...
         // should match full name, since nickname is not present
-        assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'First Last');
+        assert.deepStrictEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'First Last');
 
         // should match nickname since it's present
-        assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser2.id), 'nick2');
+        assert.deepStrictEqual(Selectors.makeGetDisplayName()(newTestState, testUser2.id), 'nick2');
 
-        newTestState.entities.general = {config: {TeammateNameDisplay: General.TEAMMATE_NAME_DISPLAY.SHOW_USERNAME}};
+        newTestState = {
```
lieut-data: Should we add a subtest to group and avoid having to redefine `newTestState`? This would also help below with the other `*config*` changes and subsequent testing.

src/selectors/entities/users.test.js
```diff
@@ -443,35 +443,106 @@ describe('Selectors.Users', () => {
...
+                            value: General.TEAMMATE_NAME_DISPLAY.SHOW_USERNAME,
+                        },
+                    },
+                }
+            },
+        };
 
         // should match username
-        assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'username');
+        assert.deepStrictEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'username');
```
lieut-data: should we also test `testUser2` here?

src/selectors/entities/users.test.js
```diff
@@ -443,35 +443,106 @@ describe('Selectors.Users', () => {
...
+                    myPreferences: {
+                        [`${Preferences.CATEGORY_DISPLAY_SETTINGS}--${Preferences.NAME_NAME_FORMAT}`]: {
+                            value: General.TEAMMATE_NAME_DISPLAY.SHOW_FULLNAME,
+                        },
+                    },
+                }
+            },
+        };
+        // should match full name
+        assert.deepStrictEqual(Selectors.makeGetDisplayName()(newTestState, testUser2.id), 'First2 Last2');
```
lieut-data: Should we also test `testUser1` here?

hahmadia:
Hey @lieut-data ,
I totally agree with what we discussed in our meeting so I decided that it would make the most sense if I further simplified `src/selectors/entities/users.test.js`.  If we have preferences, the selector should return the name using our preferences. If we have `LockTeammateNameDisplay` set  to 'true', then it should evaluate that correctly and pass a boolean value of `true` to `displayUsername`. I think we should only test if the selector evaluates the corrects value from our config and passes it down to `displayUsername`. The return value is something we need to take care off at the `user_utils.test.js` level as you suggested.

I then decided to transfer all of the tests being done at the selector level to `user_utils.test.js` since we are concerning ourself with the output of that function and that shouldn't be done at selector level. I was able to run through different permutations with `LockTeammateNameDisplay` set to `false` and `true`.

PR #976, opened 2019-11-12
Remove more implicit anys

jespino:
#### Summary
Remove more implicit anys to progress in the removal of the allowImplicitAny from tsconfig.json


jespino:
Closed in favor of #977

PR #977, opened 2019-11-12
No implicit anys

reflog:
#### Summary
Enabled noImplicitAny setting and fixed most typing issues


src/actions/gifs.ts
```diff
@@ -214,12 +215,12 @@ export function searchCategory({tagName = '', gfyCount = 30, cursorPos = undefin
...
             }
-        ).catch((err) => dispatch(errorSearching(err, tagName)));
+        ).catch((err: any) => dispatch(errorSearching(err, tagName)));
     };
 }
 
-export function shouldSearch(state, searchText) {
-    const resultsByTerm = state.search.resultsByTerm[searchText];
+export function shouldSearch(state: GlobalState, searchText: string) {
+    const resultsByTerm = state.entities.gifs.search.resultsByTerm[searchText];
```
jespino: nice catch

src/actions/gifs.ts
```diff
@@ -120,29 +121,29 @@ export function receiveSearchById(gfyId, gfyItem) {
...
+export function errorSearchById(err: any, gfyId: string) {
     return {
         type: GifTypes.SEARCH_BY_ID_FAILURE,
         err,
         gfyId,
     };
 }
 
-export function searchScrollPosition(scrollPosition) {
+export function searchScrollPosition(scrollPosition: any) {
```
jespino: scrollPosition, is a number accordingly to mattermost-webapp source code. (we can replace it here and in the other positions in this same file).

reflog: 👍

src/actions/plugins.ts
```diff
@@ -8,7 +8,14 @@ import {PluginTypes} from 'action_types';
 
 import {ActionFunc} from 'types/actions';
 
-export function getMarketplacePlugins(filter): ActionFunc {
+export type MarketplacePluginFilter = {
+    Page: number;
+	PerPage: number;
```
jespino: weird spacing here

src/reducers/entities/channels.ts
```diff
@@ -36,7 +38,7 @@ function currentChannelId(state = '', action) {
     }
 }
 
-function channels(state = {}, action) {
+function channels(state: any = {}, action: GenericAction) {
```
jespino: I think we should have types for this `any`s in the `types/store.ts` file. the types used as "sub-tree" of the `GlobalState` type.

src/selectors/entities/integrations.ts
```diff
@@ -33,7 +33,7 @@ export const getOutgoingHooksInCurrentTeam = reselect.createSelector(
     getCurrentTeamId,
     getOutgoingHooks,
     (teamId, hooks) => {
-        return Object.values(hooks).filter((o) => o.teamId === teamId);
+        return Object.values(hooks).filter((o) => o.team_id === teamId);
```
jespino: another nice catch

src/selectors/entities/roles.ts
```diff
@@ -112,7 +115,7 @@ export const getMyTeamPermissions = reselect.createSelector(
     getMyTeamRoles,
     getRoles,
     getMySystemPermissions,
-    (state, options) => options.team,
+    (state: GlobalState, options: any) => options.team,
```
jespino: We can type easily the permissions options, are always the same, optional `channel` optional `team` and mandatory `permission`

src/selectors/entities/users.ts
```diff
@@ -451,9 +454,9 @@ export function makeGetProfilesByIdsAndUsernames(): (a: GlobalState, b: {allUser
     return createSelector(
         getUsers,
         getUsersByUsername,
-        (state, props) => props.allUserIds,
+        (state: GlobalState, props: any) => props.allUserIds,
```
jespino: Probably we can replace that `any` with an specific typing there (same in the the next line)

src/types/actions.ts
```diff
@@ -44,32 +44,37 @@ export function batchActions(actions: Action[], type = BATCH) {
...
-    function dispatchChildActions(store, action) {
-        if (action.meta && action.meta.batch) {
-            action.payload.forEach((childAction) => {
-                dispatchChildActions(store, childAction);
-            });
-        } else {
-            store.dispatch(action);
-        }
-    }
+// export function batchDispatchMiddleware(store_: GlobalState) {
```
jespino: If is it not used anywhere we can just remove it.

src/types/files.ts
```diff
@@ -22,4 +22,10 @@ export type FilesState = {
     fileIdsByPostId: {
         [x: string]: Array<string>;
     };
+    filePublicLink?: string;
 };
+
+export type FileUploadResponse = {
+	file_infos: FileInfo[];
```
jespino: weird spacing

src/utils/file_utils.ts
```diff
@@ -2,8 +2,11 @@
 // See LICENSE.txt for license information.
 import {Files, General} from '../constants';
 import {Client4} from 'client';
-import mimeDB from 'mime-db';
 import {FileInfo} from 'types/files';
+import {Dictionary} from 'types/utilities';
+
+const mimeDB = require('mime-db');
```
jespino: why `require` instead of `import`?

src/utils/gfycat_sdk.ts
```diff
@@ -1,23 +1,24 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-import Gfycat from 'gfycat-sdk';
+const GfycatConstructor = require('gfycat-sdk');
```
jespino: This is a super weird solution to solve the typing problem, I would prefer to define an "full module any" type definition.

src/utils/file_utils.ts
reflog: due to broken upstream typings.

PR #984, opened 2019-11-13
Pin version of serialize-error to 2.1.0

### LINKED ISSUE #983: Updated dependency fails RN app
#### Summary
Recent version bump (d4b655287768cd72089a3f82e43652a9b4e1008b) to `serialize-errors` prevents RN app from running.

#### Steps to reproduce
mattermost-redux: With current HEAD on `master` (2289b684b65ac1f9fe438a1b7139523a20143102):  
* `make clean`
* `make`

mattermost-mobile: With current HEAD on `master` (https://github.com/mattermost/mattermost-mobile/commit/e5e3bcc84441ed20eb5060bf64f6312b396c9041):  
* `make clean`
* `make pre-run`
* `make run-ios/run-android`

With these steps, the mobile app should fail when building bundle with the following redbox:

Screen Shot 2019-11-13 at 10 19 21

#### Possible fixes

So, [this recent commit](https://github.com/mattermost/mattermost-redux/commit/d4b655287768cd72089a3f82e43652a9b4e1008b) prevents RN app `master` from building, if it is pointed to latest mattermost-redux commit.

We've had the [same issue before](https://github.com/mattermost/mattermost-redux/pull/822), but this time I can't just do a simple revert because there's a pre-commit hook fails the `implicit anys` rule for library dependencies. This prevents the `serialize-error` library from going back to an older version.

I'm looking at options, including possibly removing `serialize-error`, but this is a more involved effort. Another option could be to somehow not run the pre-commit hook on dependency libraries?

chuttam:
#### Summary
Put `serialize-error` back to version 2.1.0. It's a node library and later versions use node packages (e.g. `util`).

#### Ticket Link
Fixes #983 

#### Checklist

#### Test Information
This PR was tested on: iPhone X (simulator)

---

We've had to do this once before in https://github.com/mattermost/mattermost-redux/pull/822 , but this new attempt is a tiny bit more involved because of type checking rules. The PR does introduce an explicit `any` usage, which compiles down to the correct `serializeError.ErrorObject` type.

### DIFF
--- a/src/actions/errors.ts
+++ b/src/actions/errors.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 import {ErrorTypes} from 'action_types';
-import {serializeError, ErrorObject} from 'serialize-error';
+import serializeError from 'serialize-error';
 import {Client4} from 'client';
 import EventEmitter from 'utils/event_emitter';
 import {DispatchFunc, ActionFunc} from 'types/actions';
@@ -22,7 +22,7 @@ export function dismissError(index: number): ActionFunc {
     };
 }
 
-export function getLogErrorAction(error: ErrorObject, displayable = false) {
+export function getLogErrorAction(error: any, displayable = false) {
     return {
         type: ErrorTypes.LOG_ERROR,
         displayable,
--- a/package-lock.json
+++ b/package-lock.json
@@ -6844,19 +6844,9 @@
       "dev": true
     },
     "serialize-error": {
-      "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/serialize-error/-/serialize-error-5.0.0.tgz",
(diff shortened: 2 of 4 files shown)


src/actions/errors.ts
```diff
@@ -22,7 +22,7 @@ export function dismissError(index: number): ActionFunc {
     };
 }
 
-export function getLogErrorAction(error: ErrorObject, displayable = false) {
+export function getLogErrorAction(error: any, displayable = false) {
```
chuttam: Setting this to `error: serializeError.ErrorObject` fails because v2.1.0 of `serializer-error` doesn't export ErrorObject, best I can tell.

Could someone with stronger TypeScript fu than me perhaps guide how to create/export an explicit type definition that matches the structure of `ErrorObject` below?

```
export type ErrorObject = {
	name?: string;
	stack?: string;
	message?: string;
	code?: string;
}
```

reflog:
Closing, since it's already handled by https://github.com/mattermost/mattermost-mobile/pull/3553

PR #989, opened 2019-11-15
Fix sync persisted channels

enahum:
#### Summary
Fixes persisted channels sync by adding a flag to the `fetchChannelsAndMembers` action as there are other actions that dispatch `RECEIVED_CHANNELS` that were causing the store to remove every other channel not received in the requests.

NOTE: also update package-lock.json to include some fixes from `npm audit`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3879,9 +3879,9 @@
       "dev": true
     },
     "handlebars": {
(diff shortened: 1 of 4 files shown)


enahum:
@jwilander pending your review

PR #995, opened 2019-11-30
[GH-10200] Isolate unit tests for channels selectors and make them more thorough

larkox:
#### Summary
Isolate unit tests for selectors and make them more thorough

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/10200

#### Checklist

#### Test Information

### DIFF
--- a/src/selectors/entities/channels.test.js
+++ b/src/selectors/entities/channels.test.js
@@ -518,6 +518,374 @@ describe('Selectors.Channels.getChannelsNameMapInCurrentTeam', () => {
     });
 });
 
+describe('Selectors.Channels.getChannelsNameMapInTeam', () => {
+    const team1 = TestHelper.fakeTeamWithId();
+    const team2 = TestHelper.fakeTeamWithId();
+
+    const channel1 = TestHelper.fakeChannelWithId(team1.id);
+    channel1.name = "Ch1";
+    const channel2 = TestHelper.fakeChannelWithId(team2.id);
+    channel2.name = "Ch2";
+    const channel3 = TestHelper.fakeChannelWithId(team2.id);
+    channel3.name = "Ch3";
+    const channel4 = TestHelper.fakeChannelWithId(team1.id);
+    channel4.name = "Ch4";
+
+    const channels = {};
+    channels[channel1.id] = channel1;
+    channels[channel2.id] = channel2;
+    channels[channel3.id] = channel3;
+    channels[channel4.id] = channel4;
+
+    const channelsInTeam = {};
+    channelsInTeam[team1.id] = [channel1.id, channel4.id];
+    channelsInTeam[team2.id] = [channel2.id, channel3.id];
+
+    const testState = deepFreezeAndThrowOnMutation({
+        entities: {
+            channels: {
+                channels,
+                channelsInTeam,
+            },
+        },
+    });
+    it('get channel map for team', () => {
+        const channelMap = {
+            [channel1.name]: channel1,
+            [channel4.name]: channel4,
+        };
+        assert.deepEqual(Selectors.getChannelsNameMapInTeam(testState, team1.id), channelMap);
+    });
+    it('get empty map for non-existing team', () => {
+        assert.deepEqual(Selectors.getChannelsNameMapInTeam(testState, 'junk'), {});
+    })
+});
+
+describe('Selectors.Channels.getChannelsByCategory', () => {
+    const team1 = TestHelper.fakeTeamWithId();
+    const team2 = TestHelper.fakeTeamWithId();
+
+    const user = TestHelper.fakeUserWithId();
+    const user2 = TestHelper.fakeUserWithId();
+    const user3 = TestHelper.fakeUserWithId();
+
+    const profiles = {};
+    profiles[user.id] = user;
+    profiles[user2.id] = user2;
+    profiles[user3.id] = user3;
+
+    const channel1 = TestHelper.fakeChannelWithId(team1.id);
+    channel1.display_name = 'Channel Name';
+    channel1.type = General.OPEN_CHANNEL;
+    const channel2 = TestHelper.fakeChannelWithId(team2.id);
+    channel2.display_name = 'Channel Name';
+    channel2.type = General.PRIVATE_CHANNEL;
+    const channel3 = TestHelper.fakeChannelWithId(team1.id);
+    channel3.display_name = 'Channel Name';
+    channel3.type = General.OPEN_CHANNEL;
+    const channel4 = TestHelper.fakeChannelWithId('');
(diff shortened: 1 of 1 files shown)


jasonblais:
Thanks @larkox for the pull request! Adding `Work In Progress` label while the PR is in draft stage, let us know when it's ready for a review.

larkox:
Thanks @jasonblais . If you or anyone in the team can take a look to what is done so far, it would be great. The issue will be kind of time consuming and cumbersome (understanding every function, making sure the state is the minimal portion of the state for everything to function correctly...) so it would be great if anyone could confirm what is done so far is what is expected from this PR. Once I have the confirmation, I can put more resources into this while being sure that I am not doing anything unnecessary or directly wrong.
Thanks in advance!

jasonblais:
Sounds like a good plan - @deanwhillier / @hmhealey can you (or someone else in the Apps team) help do a preliminary review as suggested by @larkox? Thanks!

src/selectors/entities/channels.test.js
```diff
@@ -9,6 +9,884 @@ import {sortChannelsByDisplayName, getDirectChannelName} from 'utils/channel_uti
...
+    const team2 = TestHelper.fakeTeamWithId();
+
+    it('should return channels in current team', () => {
+        const user = TestHelper.fakeUserWithId();
+
+        const profiles = {};
+        profiles[user.id] = user;
+
+        const channel1 = TestHelper.fakeChannelWithId(team1.id);
+        channel1.display_name = 'Channel Name';
```
hmhealey: What's the reason for setting the display name on all these?

src/selectors/entities/channels.test.js
```diff
@@ -9,6 +9,884 @@ import {sortChannelsByDisplayName, getDirectChannelName} from 'utils/channel_uti
...
+        const channel2 = TestHelper.fakeChannelWithId(team2.id);
+        channel2.display_name = 'Channel Name';
+        const channel3 = TestHelper.fakeChannelWithId(team1.id);
+        channel3.display_name = 'Channel Name';
+        const channel4 = TestHelper.fakeChannelWithId('');
+        channel4.display_name = 'Channel Name';
+
+        const channels = {};
```
hmhealey: Feel free to keep doing this how you're doing it, but you could avoid repeating the name of the object if you wanted to
```javascript
const channels = {
    [channel1.id]: channel1,
    [channel2.id]: channel2,
    [channel3.id]: channel3,
    [channel4.id]: channel4,
};
```

src/selectors/entities/channels.test.js
```diff
@@ -9,6 +9,884 @@ import {sortChannelsByDisplayName, getDirectChannelName} from 'utils/channel_uti
...
+                    channelsInTeam,
+                },
+            },
+        });
+
+        const channelsInCurrentTeam = [channel1, channel3].sort(sortChannelsByDisplayName.bind(null, []));
+        assert.deepEqual(Selectors.getChannelsInCurrentTeam(testState), channelsInCurrentTeam);
+    });
+
+    it('should order by user locale', () => {
```
hmhealey: Good to finally get a test for these. I know locale can affect sorting, but I didn't know of any cases where that happened before.

src/selectors/entities/channels.test.js
```diff
@@ -9,6 +9,884 @@ import {sortChannelsByDisplayName, getDirectChannelName} from 'utils/channel_uti
...
+
+        const channelsInCurrentTeamDe = [channel1, channel2].sort(sortChannelsByDisplayName.bind(null, userDe.locale));
+        const channelsInCurrentTeamSv = [channel1, channel2].sort(sortChannelsByDisplayName.bind(null, userSv.locale));
+
+        assert.deepEqual(Selectors.getChannelsInCurrentTeam(testStateDe), channelsInCurrentTeamDe);
+        assert.deepEqual(Selectors.getChannelsInCurrentTeam(testStateSv), channelsInCurrentTeamSv);
+    });
+});
+
+describe('Selectos.Channels.getMyChannels', () => {
```
hmhealey:
```suggestion
describe('Selectors.Channels.getMyChannels', () => {
```
Although I think you could also consider dropping the `Selectors.Channels` part of the test suite name anyway since that's implied by the location of the file anyway.

src/selectors/entities/channels.test.js
```diff
@@ -9,6 +9,884 @@ import {sortChannelsByDisplayName, getDirectChannelName} from 'utils/channel_uti
...
+});
+
+describe('Selectors.Channels.getMembersInCurrentChannel', () => {
+    const channel1 = TestHelper.fakeChannelWithId('');
+
+    const user = TestHelper.fakeUserWithId();
+    const user2 = TestHelper.fakeUserWithId();
+    const user3 = TestHelper.fakeUserWithId();
+
+    const membersInChannel = {};
```
hmhealey: Using an object literal for this whole definition as I mentioned above would definitely help readability here

larkox:
Thanks for the comments @hmhealey . Regarding dividing this in several PR, I think that one good approach would be finish the isolation by myself without going deep in the test cases (just reproduce the ones that are already in the file) and then you could open smaller issues to complete the tests for each function. What do you think? Does it make sense?

hmhealey:
Yeah, we could certainly do that. I thought it might be easier to do them at the same time, but if you find it easier to isolate the tests first, that's totally fine with me.

src/selectors/entities/channels.test.js
larkox: I set them because the ordering use in theory the display_name. Nevertheless, I removed them and the result is the same, so they are not needed.

src/selectors/entities/channels.test.js
larkox: I was mainly following the patter they were using on the original test. I can do it this other way without any problem.

src/selectors/entities/channels.test.js
larkox: All merit to the [function documentation](https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/String/localeCompare). I got the example they had there and used it.

src/selectors/entities/channels.test.js
larkox: If we remove Selectors because is implied by the file, shall we drop also Channels?

larkox:
I think this is already ready for review. Some test may still not do exactly what they should, but at least most should be working. If you create a new campaign to improve the tests, I am pretty sure the community can make a great test out of this.

src/selectors/entities/channels.test.js
hmhealey: Yeah, I'd lean towards that myself. Feel free to leave it as-is though since we aren't really consistent with that naming.

PR #998, opened 2019-12-02
[MM-20164] Dispatch redux when users role is changed from member to system admin or vice versa

hahmadia:
#### Summary
Adds a dispatch to load new roles when a users roles is changed (from member to system admin or vice versa)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20164


amyblais:
This should be cherry-picked if this is for 5.18.

PR #1002, opened 2019-12-03
[MM-18177] Emit INCREASE_POST_VISIBILITY_BY_ONE for current channel

migbot:
#### Summary
Added a new websocket event to be emitted in `handleNewPostEvent` when a new post is seen for the current channel. The mobile app will react to this event by incrementing the post visibility of the current channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-18177

### DIFF
--- a/src/actions/websocket.ts
+++ b/src/actions/websocket.ts
@@ -326,7 +326,12 @@ function handleEvent(msg: WebSocketMessage) {
 
 function handleNewPostEvent(msg: WebSocketMessage) {
     return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
         const post = JSON.parse(msg.data.post);
+
+        if (getCurrentChannelId(state) === post.channel_id) {
+            EventEmitter.emit(WebsocketEvents.INCREASE_POST_VISIBILITY_BY_ONE);
+        }
 
         dispatch(handleNewPost(msg));
         getProfilesAndStatusesForPosts([post], dispatch, getState);
@@ -337,6 +342,7 @@ function handleNewPostEvent(msg: WebSocketMessage) {
                 data: [{user_id: post.user_id, status: General.ONLINE}],
             });
         }
+
         return {data: true};
     };
 }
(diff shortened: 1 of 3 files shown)


src/actions/websocket.ts
```diff
@@ -326,8 +326,13 @@ function handleEvent(msg: WebSocketMessage) {
...
 function handleNewPostEvent(msg: WebSocketMessage) {
     return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
         const post = JSON.parse(msg.data.post);
 
+        if (getCurrentChannelId(state) === post.channel_id) {
+            EventEmitter.emit(WebsocketEvents.INCREASE_POST_VISIBILITY_BY_ONE);
+        }
+
         dispatch(handleNewPost(msg));
```
enahum: I still want to avoid getting the new post to the store if we already have it

migbot: Ah gotcha.

migbot: Done.

src/actions/websocket.ts
```diff
@@ -326,17 +326,26 @@ function handleEvent(msg: WebSocketMessage) {
 
 function handleNewPostEvent(msg: WebSocketMessage) {
     return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
         const post = JSON.parse(msg.data.post);
 
-        dispatch(handleNewPost(msg));
-        getProfilesAndStatusesForPosts([post], dispatch, getState);
+        if (getCurrentChannelId(state) === post.channel_id) {
```
enahum: this should be in if statement cause you should have already incremented the post visibility if you already have the post.

migbot: Hmm, in my testing I did not see the post visibility incremented without this.

enahum: you should handle it in the app, instead of relying on this event when the websocket gets the post back.

Reasoning: you create a post in the client, WS is disconnected so you miss the broadcast event but you still should increase the visibility

PR #1005, opened 2019-12-08
MM-20743  - CI uses npm ci instead of npm install

reflog:
#### Summary
Switched to using `npm ci` on CI instead of `npm install`
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20743

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
       - checkout
       - run: ls -la
       - *restore_cache
-      - run: npm install
+      - run: npm ci
       - run: ls -la
       - *save_cache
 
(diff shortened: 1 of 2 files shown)


sudheerDev:
More of an FYI. AFAIK there is an issue with `npm ci` and prepare scripts last time i checked. So, this will be a problem if we do it on web or mobile repos https://github.com/mattermost/mattermost-server/issues/10334

reflog:
thanks @sudheerDev , I think that issue you referenced doesn't exist anymore. I think we ironed out that problem when migrating to Typescript.

sudheerDev:
@reflog Awesome good to know. 👍

PR #1011, opened 2019-12-12
Enable skipped tests and fix memoization of createIdsSelector

saturninoabril:
#### Summary
1. Enable skipped unit tests in utils/post_list.test.js
2. On enabling those tests, observed that it's failing on memoization of `makeCombineUserActivityPosts`. Fixed by reverting change [here](https://github.com/mattermost/mattermost-redux/pull/936/files#diff-f813fe557eb39d392c00fb67ecd64d48R9). This is most likely causing some performance issues since recomputation on `createIdsSelector` is always true.

Before:
Screen Shot 2019-12-12 at 4 22 22 PM

After:
Screen Shot 2019-12-12 at 5 19 40 PM


sudheerDev:
This is cherrypicked to 5.19v branch

PR #1013, opened 2019-12-13
MM-21153 Count mentions in DM channels using mention_count 

hmhealey:
This is to fix a bug where marking a DM channel as unread would incorrectly set the unread count in the browser tab title and in the mobile app badge. The unread count was including the user's own messages because the logic for counting unread mentions in a DM used the channel member's `msg_count` field instead of its `mention_count`.

The actual fix for this PR is the line I commented on below. The rest of this is cleaning up other tests because a bunch of them were broken due to missing fields (like `msg_count` and `notify_props`) in the objects that were being tested.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21153

### DIFF
--- a/src/selectors/entities/channels.ts
+++ b/src/selectors/entities/channels.ts
@@ -305,33 +305,48 @@ export const getMembersInCurrentChannel: (a: GlobalState) => UserIDMappedObjects
 export const getUnreads: (a: GlobalState) => {
     messageCount: number;
     mentionCount: number;
-} = createSelector(getAllChannels, getMyChannelMemberships, getUsers, getCurrentUserId, getCurrentTeamId, getMyTeams, getTeamMemberships, (channels: IDMappedObjects<Channel>, myMembers: RelationOneToOne<Channel, ChannelMembership>, users: IDMappedObjects<UserProfile>, currentUserId: string, currentTeamId: string, myTeams: Array<Team>, myTeamMemberships: RelationOneToOne<Team, TeamMembership>): {
-    messageCount: number;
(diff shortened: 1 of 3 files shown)


src/selectors/entities/channels.ts
```diff
@@ -305,62 +305,80 @@ export const getMembersInCurrentChannel: (a: GlobalState) => UserIDMappedObjects
...
-        if (channel && m && (channel.team_id === currentTeamId || channel.type === General.DM_CHANNEL || channel.type === General.GM_CHANNEL)) {
             let otherUserId = '';
 
-            if (channel.type === 'D') {
+            if (channel.type === General.DM_CHANNEL) {
                 otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
 
                 if (users[otherUserId] && users[otherUserId].delete_at === 0) {
-                    mentionCountForCurrentTeam += channel.total_msg_count - m.msg_count;
+                    mentionCountForCurrentTeam += m.mention_count;
```
hmhealey: This line is the actual fix here. The rest is cleanup

saturninoabril:
@enahum, done.  PTAL.

 I did preserve as much as I could the new tests added by HH.  Tests like:
- `describe('Selectors.Channels.getUnreadChannelIds', () => {...})`
- `describe('Selectors.Channels.getUnreads', () => {...});`

This will definitely cause conflict on 5.18 branch on cherry-pick.  I guess, we'll need to submit a separate PR to 5.18 and that would just copying the original tests from HH.

PR #1021, opened 2019-12-19
MM-21262 - Removed prepackaged plugin flags

marianunez:
#### Summary
Removed flag that would label a plugin as prepackaged from a fix list of plugins given that there is no longer any restriction to uninstall from the system console.

#### Ticket Link
[MM-21262](https://mattermost.atlassian.net/browse/MM-21262)

#### Related Pull Requests
- Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/4560

### DIFF
--- a/src/reducers/entities/admin.test.js
+++ b/src/reducers/entities/admin.test.js
@@ -31,7 +31,6 @@ describe('reducers.entities.admin', () => {
                         state: PluginState.PLUGIN_STATE_NOT_RUNNING,
                         name: 'Plugin 0',
                         description: 'The plugin 0.',
-                        is_prepackaged: false,
                     },
                     {
                         plugin_id: 'plugin_1',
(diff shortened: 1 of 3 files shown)


src/reducers/entities/admin.test.js
```diff
@@ -186,9 +175,7 @@ describe('reducers.entities.admin', () => {
                     version: '0.0.1',
                     state: PluginState.PLUGIN_STATE_RUNNING,
                     name: 'Plugin 1',
-                    description: 'The plugin.',
-                    is_prepackaged: false,
-                    active: true,
+                    description: 'The plugin.',                    active: true,
```
lieut-data: Looks like this line got merged together accidentally?

marianunez: Done! Surprised that the check style didn't catch that 🤔

PR #1024, opened 2019-12-21
MM-21290: Checks for license AND config prior to overriding user's di…

mkraft:
…splay name account setting.

#### Summary

Checks for license *and* config setting prior to overriding the user's display name account setting.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21290

#### Dependency

https://github.com/mattermost/mattermost-server/pull/13431

### DIFF
--- a/src/selectors/entities/users.ts
+++ b/src/selectors/entities/users.ts
@@ -490,8 +490,9 @@ export function makeGetDisplayName(): (a: GlobalState, b: $ID<UserProfile>, c: b
         getTeammateNameDisplaySetting,
         (state, _, useFallbackUsername = true) => useFallbackUsername,
         getConfig,
-        (user, teammateNameDisplaySetting, useFallbackUsername, config) => {
-            const useAdminTemmateNameDisplaySetting = config.LockTeammateNameDisplay === 'true';
+        getLicense,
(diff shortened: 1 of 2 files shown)


src/selectors/entities/users.ts
```diff
@@ -490,8 +490,9 @@ export function makeGetDisplayName(): (a: GlobalState, b: $ID<UserProfile>, c: b
         getTeammateNameDisplaySetting,
         (state, _, useFallbackUsername = true) => useFallbackUsername,
         getConfig,
-        (user, teammateNameDisplaySetting, useFallbackUsername, config) => {
-            const useAdminTemmateNameDisplaySetting = config.LockTeammateNameDisplay === 'true';
+        getLicense,
+        (user, teammateNameDisplaySetting, useFallbackUsername, config, license) => {
+            const useAdminTemmateNameDisplaySetting = license.LockTeammateNameDisplay === 'true' && config.LockTeammateNameDisplay === 'true';
```
enahum: What if license or config is undefined? Actually the unit tests are failing exactly cause of that reason

mkraft: license and config are keys present on the initial global state so they should only even be undefined in mocks.

PR #1029, opened 2020-01-03
MM-19976 & MM-19978 - Guest accounts, guest user should not see members that share no channels with them

reflog:
#### Summary

When a member leaves/being removed, and is not sharing any channels/teams with the current guest user - the guest user should not see that member or access it's info.
This PR adds a new event that is fired when this occurs and cleans up the state.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-19978 and https://mattermost.atlassian.net/browse/MM-19976

**Might need additional work on the mobile side to fix offline sync**

### DIFF
--- a/src/actions/websocket.ts
+++ b/src/actions/websocket.ts
@@ -21,13 +21,13 @@ import {ActionFunc, DispatchFunc, GetStateFunc, PlatformType, batchActions} from
 
 import {getTeam, getMyTeamUnreads, getMyTeams, getMyTeamMembers} from './teams';
 import {getPost, getPosts, getProfilesAndStatusesForPosts, getCustomEmojiForReaction, getUnreadPostData, handleNewPost, postDeleted, receivedPost} from './posts';
-import {fetchMyChannelsAndMembers, getChannelAndMyMember, getChannelStats, markChannelAsRead, getChannelMembers} from './channels';
+import {fetchMyChannelsAndMembers, getChannelAndMyMember, getChannelStats, markChannelAsRead} from './channels';
 import {checkForModifiedUsers, getMe, getProfilesByIds, getStatusesByIds, loadProfilesForDirect} from './users';
 import {loadRolesIfNeeded} from './roles';
 import {Channel, ChannelMembership} from 'types/channels';
 import {Dictionary} from 'types/utilities';
 import {PreferenceType} from 'types/preferences';
-import {rolesIncludePermission} from 'utils/user_utils';
+import {isGuest} from 'utils/user_utils';
 let doDispatch: DispatchFunc;
 export function init(platform: PlatformType, siteUrl: string | undefined | null, token: string | undefined | null, optionalWebSocket: any, additionalOptions: any = {}) {
     return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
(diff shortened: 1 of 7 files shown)


src/actions/websocket.ts
```diff
@@ -17,16 +17,17 @@ import {fromAutoResponder} from 'utils/post_utils';
...
 import {getTeam, getMyTeamUnreads, getMyTeams, getMyTeamMembers} from './teams';
 import {getPost, getPosts, getProfilesAndStatusesForPosts, getCustomEmojiForReaction, getUnreadPostData, handleNewPost, postDeleted, receivedPost} from './posts';
-import {fetchMyChannelsAndMembers, getChannelAndMyMember, getChannelStats, markChannelAsRead} from './channels';
+import {fetchMyChannelsAndMembers, getChannelAndMyMember, getChannelStats, markChannelAsRead, getChannelMembers} from './channels';
 import {checkForModifiedUsers, getMe, getProfilesByIds, getStatusesByIds, loadProfilesForDirect} from './users';
 import {loadRolesIfNeeded} from './roles';
 import {Channel, ChannelMembership} from 'types/channels';
 import {Dictionary} from 'types/utilities';
 import {PreferenceType} from 'types/preferences';
+import {rolesIncludePermission} from 'utils/user_utils';
```
jespino: can we add a different function here, something like `isGuest` which is more explicit and clear (see `isSystemAdmin` function in `utils/user_utils.ts` as example)

enahum:
What happens if the WS is disconnected? How is this handled on clients like the mobile app that persist data in local storage?

reflog:
@enahum - we discussed this with @jespino , and (as I mentioned in the description of this PR) this would need to be handled on mobile separately during the re-hydration step, it's out of redux's purview

enahum:
@reflog is there a ticket for that?

reflog:
@enahum not yet. Once this is merged - I'll open a ticket for mobile team

reflog:
@jespino apologies, forgot to commit 🤦‍♂️

enahum:
> @enahum not yet. Once this is merged - I'll open a ticket for mobile team

What version is this for?

reflog:
@enahum - no idea. version not set. @amyblais  ?

src/reducers/entities/users.ts
```diff
@@ -341,7 +366,11 @@ function statuses(state: RelationOneToOne<UserProfile, string> = {}, action: Gen
     }
     case UserTypes.LOGOUT_SUCCESS:
         return {};
-
+    case UserTypes.PROFILE_NO_LONGER_VISIBLE: {
+        const newState = {...state};
+        delete newState[action.data.user_id];
```
enahum: you should check that the state actually contains the userId before deleting, just to avoid returning a newState if nothing has changed.

src/reducers/entities/users.ts
```diff
@@ -178,7 +186,11 @@ function profiles(state: IDMappedObjects<UserProfile> = {}, action: GenericActio
             },
         };
     }
-
+    case UserTypes.PROFILE_NO_LONGER_VISIBLE: {
+        const newState = {...state};
+        delete newState[action.data.user_id];
```
enahum: you should check that the state actually contains the userId before deleting, just to avoid returning a newState if nothing has changed.

src/reducers/entities/users.ts
```diff
@@ -67,6 +67,14 @@ function addProfileToSet(state: RelationOneToMany<Team, UserProfile>, action: Ge
     };
 }
 
+function removeProfileFromTeams(state: RelationOneToMany<Team, UserProfile>, action: GenericAction) {
+    const newState = {...state};
+    Object.keys(state).forEach((key) => {
```
enahum: Maybe consider having an additional variable to keep track if anything actually change in the state to avoid returning a new state if nothing was changed.

PR #1035, opened 2020-01-15
MM-21726 Add theme type and function to set theme defaults

jwilander:
#### Summary
This will be used in an upcoming webapp PR to fix the bug in the ticket. Essentially we need to make sure all the theme properties are set before applying themes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21726


hmhealey:
What are the corresponding changes on the web app for this? The `getTheme` selector should be setting any missing fields already when the user's theme is pulled from redux

PR #1041, opened 2020-01-23
MM-21890: add team_display_name to Channel object

sbishel:
#### Summary
Adds missing 'team_display_name' to channel so team_display_name is shown rather than the team_id

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21890

#### Related PRs
https://github.com/mattermost/mattermost-webapp/pull/4731

### DIFF
--- a/src/types/channels.ts
+++ b/src/types/channels.ts
@@ -20,7 +20,6 @@ export type Channel = {
     update_at: number;
     delete_at: number;
     team_id: string;
-    team_display_name: string;
     type: ChannelType;
     display_name: string;
     name: string;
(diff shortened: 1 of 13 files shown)


enahum:
Perhaps you want a selector that does this? We are already suffering from randomly modifying the immutability of the channel object

src/types/channels.ts
```diff
@@ -20,6 +20,7 @@ export type Channel = {
     update_at: number;
     delete_at: number;
     team_id: string;
+    team_display_name: string;
```
reflog: should this be a required field?

hmhealey: Alternatively, should this be a separate type that extends `Channel`? I don't think most of the APIs return channels with team information included

PR #1048, opened 2020-02-10
[MM-22205] [MM-22206] Add GetChannelModerations and PatchChannelModerations

fm2munsh:
#### Summary
- Adds actions for `GetChannelModerations` and `PatchChannelModerations` 
- Adds a selector for `GetChannelModerations`
- Adds `ChannelModeration` and `ChannelModerationPatch` types
- Tests for added actions and selector

- The newly added endpoints wont work without the following server PR https://github.com/mattermost/mattermost-server/pull/13845 (And the two PRs that depend on it) 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22205
https://mattermost.atlassian.net/browse/MM-22206


src/actions/channels.ts
```diff
@@ -16,15 +16,15 @@ import {getConfig} from 'selectors/entities/general';
...
-import {Channel, ChannelNotifyProps, ChannelMembership} from 'types/channels';
+import {Channel, ChannelNotifyProps, ChannelMembership, ChannelModeration, ChannelModerationPatch} from 'types/channels';
 
 import {PreferenceType} from 'types/preferences';
 
 import {logError} from './errors';
 import {bindClientFunc, forceLogoutIfNecessary} from './helpers';
 import {getMissingProfilesByIds} from './users';
 import {loadRolesIfNeeded} from './roles';
-import {UserProfile} from 'types/users';
```
fm2munsh: this was not being used in the file so i removed it

src/actions/channels.ts
```diff
@@ -1445,6 +1445,32 @@ export function membersMinusGroupMembers(channelID: string, groupIDs: Array<stri
...
+    });
+}
+
+export function patchChannelModerations(channelId: string, patch: Array<ChannelModerationPatch>): ActionFunc {
+    return bindClientFunc({
+        clientFunc: async () => {
+            const moderations = await Client4.patchChannelModerations(channelId, patch);
+            return {channelId, moderations};
+        },
+        onSuccess: ChannelTypes.RECEIVED_CHANNEL_MODERATIONS,
```
fm2munsh: should this have a failure action too?

src/selectors/entities/channels.ts
```diff
@@ -892,3 +892,11 @@ export function isManuallyUnread(state: GlobalState, channelId?: string): boolea
 
     return Boolean(state.entities.channels.manuallyUnread[channelId]);
 }
+
+export function getChannelModerations(state: GlobalState, channelId: string): Array<ChannelModeration> | object {
```
hahmadia:
```suggestion
export function getChannelModerations(state: GlobalState, channelId: string): Array<ChannelModeration> | Array<[]> {
```

src/selectors/entities/channels.ts
```diff
@@ -892,3 +892,11 @@ export function isManuallyUnread(state: GlobalState, channelId?: string): boolea
...
+    if (!channelId) {
+        return {};
```
hahmadia:
```suggestion
        return [];
```

src/selectors/entities/channels.ts
fm2munsh: Actually just dropping the object and responding with an array works since `[]` fulfills `Array<ChannelModeration>`

fm2munsh: will also add a test for this state

src/types/channels.ts
```diff
@@ -73,4 +73,27 @@ export type ChannelsState = {
...
+};
+
+export type ChannelModeration = {
+    name: string;
+    roles: {
+        guests?: {
+            value: boolean;
+            enabled: boolean;
+        };
+        members: {
```
hahmadia:
```suggestion
        members?: {
```

fm2munsh: I think a moderation will always have a member attribute - since all channel moderations apply to the members role but not necessarily to guests (manage members permission)

src/actions/channels.test.js
```diff
@@ -2278,4 +2278,52 @@ describe('Actions.Channels', () => {
...
+        reply(200,  [{
+                "name": "create_posts",
+                "roles": {
+                    "members": true,
+                    "guests": false,
+                },
+            }]
+        );
+
+        const {error} = await Actions.getChannelModerations(channelID)(store.dispatch, store.getState);
```
hmhealey: Actions should actually be passed into `dispatch` instead of called directly. I'm going to create a ticket to go do a mass refactor of these since we've been slowly changing that as we come across it, but that hasn't stopped more from being added.

```suggestion
        const {error} = await store.dispatch(Actions.getChannelModerations(channelID));
```

src/actions/channels.test.js
```diff
@@ -2278,4 +2278,52 @@ describe('Actions.Channels', () => {
...
+                    "name": "create_reactions",
+                    "roles": {
+                        "members": true,
+                        "guests": false,
+                    },
+                },
+            ]
+        );
+
+        const {error} = await Actions.patchChannelModerations(channelID, {})(store.dispatch, store.getState);
```
hmhealey:
```suggestion
        const {error} = await store.dispatch(Actions.patchChannelModerations(channelID, {}));
```

src/actions/channels.ts
hmhealey: It only needs an `onFailure` if we specifically need to dispatch an action for that. If the caller is just waiting for the result of `patchChannelModerations`, then it shouldn't be needed.

src/reducers/entities/channels.test.js
```diff
@@ -482,4 +494,94 @@ describe('channels', () => {
...
+            const nextState = channelsReducer(state, {
+                type: ChannelTypes.RECEIVED_CHANNEL_MODERATIONS,
+                sync: true,
+                currentChannelId: 'channel1',
+                teamId: 'team',
+                data: {
+                    channelId: 'channel1',
+                    moderations: [
+                        {
+                            "name": "create_posts",
```
hmhealey: This looks incorrect. I think the style checker might not be running correctly? It'd usually want this to be
```suggestion
                            name: 'create_posts',
```

hmhealey: Also, should `'create_posts'` be a constant?

src/selectors/entities/channels.ts
```diff
@@ -892,3 +892,7 @@ export function isManuallyUnread(state: GlobalState, channelId?: string): boolea
 
     return Boolean(state.entities.channels.manuallyUnread[channelId]);
 }
+
+export function getChannelModerations(state: GlobalState, channelId: string): Array<ChannelModeration> {
+    return state.entities.channels.channelModerations[channelId] || [];
```
hmhealey: This returns a new array each time it's called which can cause some bad behaviour when this is used in `mapStateToProps`. Can you either allow this to return undefined or make this into a selector? If made into a selector, it needs to be a selector factory (like `makeGetChannelModerations` that returns an instance of `createSelector`) since it has a parameter that needs to be memoized.

src/reducers/entities/channels.test.js
hmhealey: PR to fix ESLint: https://github.com/mattermost/mattermost-redux/pull/1050

src/selectors/entities/channels.ts
fm2munsh: @hahmadia I'm going to opt for making this return an `undefined` if there is no `channelModerations[channelId]` found from the selector - let me know if this causes any issues for you and if you'd prefer to go the selector factory route

PR #1049, opened 2020-02-11
Performance improvements for mobile app

enahum:
#### Summary
To improve the mobile app performance considerably there is a need to set/update multiple reducers while dispatching only one action type.

When adding all the data needed to one action there is only one dispatch, thus the `mapStateToProps` run only once instead of multiple times.

in this PR we add actions for:
* **LOGIN**: includes user data, teams, teams members, team unreads, preferences, config and license. All of this fetched while logging in or loading the current user data.
* **RECEIVED_MY_CHANNELS_WITH_MEMBERS**: Instead of  dispatching one action for the channels and another action for the channel members, we are dispatching only one with  the data needed for both reducers.
* **SET_CONFIG_AND_LICENSE**: as the name describes
* **RECEIVED_BATCHED_PROFILES_IN_CHANNEL**: To load the sidebar DM/GM's we execute one request per DM/GM with this action we will fetch all the results and update the state just once instead of updating thee state for each response thus alleviating the amount of times components need to recompute for each added profile to the store.

Also on this PR we got rid of the `postDate.setHours(0, 0, 0, 0)` as it is incredibly slow and not really needed.

Note: candidate for 5.21 to be included with mobile app 1.29

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22084

### DIFF
--- a/src/utils/post_list.test.js
+++ b/src/utils/post_list.test.js
@@ -30,8 +30,7 @@ describe('makeFilterPostsAndAddSeparators', () => {
     it('filter join/leave posts', () => {
         const filterPostsAndAddSeparators = makeFilterPostsAndAddSeparators();
         const time = Date.now();
-        const today = new Date();
-        today.setHours(0, 0, 0, 0);
+        const today = new Date(time);
 
         let state = {
             entities: {
@@ -145,8 +144,7 @@ describe('makeFilterPostsAndAddSeparators', () => {
     it('new messages indicator', () => {
         const filterPostsAndAddSeparators = makeFilterPostsAndAddSeparators();
         const time = Date.now();
-        const today = new Date();
-        today.setHours(0, 0, 0, 0);
+        const today = new Date(time);
 
         const state = {
             entities: {
@@ -180,7 +178,7 @@ describe('makeFilterPostsAndAddSeparators', () => {
             '1010',
             '1005',
             '1000',
-            'date-' + today.getTime(),
+            'date-' + (today.getTime() + 1000),
         ]);
 
         now = filterPostsAndAddSeparators(state, {postIds, indicateNewMessages: true});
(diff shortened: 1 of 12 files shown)


src/utils/post_list.ts
```diff
@@ -58,6 +58,7 @@ export function makeFilterPostsAndAddSeparators() {
                 return [];
             }
 
+            const dt = Date.now();
```
migbot: Where is `dt` used?

src/utils/post_list.test.js
```diff
@@ -145,8 +144,7 @@ describe('makeFilterPostsAndAddSeparators', () => {
     it('new messages indicator', () => {
         const filterPostsAndAddSeparators = makeFilterPostsAndAddSeparators();
         const time = Date.now();
-        const today = new Date();
-        today.setHours(0, 0, 0, 0);
```
migbot: What was the reason for setting/no longer setting HMS to 0?

src/utils/post_list.test.js
```diff
@@ -180,23 +178,23 @@ describe('makeFilterPostsAndAddSeparators', () => {
             '1010',
             '1005',
             '1000',
-            'date-' + today.getTime(),
+            'date-' + (today.getTime() + 1000),
```
migbot: Why add 1000 ms?

src/utils/post_list.test.js
enahum: Set hours was so the date is set to the start of the day, which is not needed at all and also extremely expensive

src/utils/post_list.test.js
enahum: Is based on the test data

src/utils/post_list.ts
enahum: Nice catch I can remove it, weird that eslint did not catched it. I added that in for a log, will remove.

enahum:
```suggestion
```

enahum:
Need a rebase after https://github.com/mattermost/mattermost-redux/pull/1051 goes in

PR #1051, opened 2020-02-12
Synchronize archived channels with storage

ethervoid:
#### Summary

This PR solves the synchronization problems we are facing due to the persistent nature of the RN storage. In order to synchronize the archived channels in the RN applications we need to ask for the archived channels when we ask for the current user channels so we can update the `delete_at` attribute for the stored channel entities

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22201

### Related PRs

https://github.com/mattermost/mattermost-server/pull/13877

### DIFF
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -12,7 +12,8 @@ import {
     isManuallyUnread,
 } from 'selectors/entities/channels';
 import {getCurrentTeamId} from 'selectors/entities/teams';
-import {getConfig} from 'selectors/entities/general';
+import {getConfig, getServerVersion} from 'selectors/entities/general';
+import {isMinimumServerVersion} from 'utils/helpers';
 
 import {Action, ActionFunc, batchActions, DispatchFunc, GetStateFunc} from 'types/actions';
 
@@ -496,7 +497,12 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
         let channels;
         let channelMembers;
         try {
-            const channelRequest = Client4.getMyChannels(teamId, true);
+            let channelRequest;
+            if (isMinimumServerVersion(getServerVersion(getState()), 5, 21)) {
+                channelRequest = Client4.getMyChannels(teamId, true);
+            } else {
+                channelRequest = Client4.getMyChannels(teamId);
+            }
             const memberRequest = Client4.getMyChannelMembers(teamId);
             channels = await channelRequest;
             channelMembers = await memberRequest;
@@ -526,6 +532,9 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
             {
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBERS,
                 data: channelMembers,
+                sync: true,
+                channels,
+                serverVersion: getServerVersion(getState()),
                 remove: getChannelsIdForTeam(getState(), teamId),
                 currentUserId,
                 currentChannelId,
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -496,7 +496,7 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
         let channels;
         let channelMembers;
         try {
-            const channelRequest = Client4.getMyChannels(teamId);
+            const channelRequest = Client4.getMyChannels(teamId, true);
             const memberRequest = Client4.getMyChannelMembers(teamId);
             channels = await channelRequest;
             channelMembers = await memberRequest;
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -496,13 +496,9 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
 
         let channels;
         let channelMembers;
+        const shouldFetchArchived = (isMinimumServerVersion(getServerVersion(getState()), 5, 21));
         try {
-            let channelRequest;
-            if (isMinimumServerVersion(getServerVersion(getState()), 5, 21)) {
-                channelRequest = Client4.getMyChannels(teamId, true);
-            } else {
-                channelRequest = Client4.getMyChannels(teamId);
-            }
+            const channelRequest = Client4.getMyChannels(teamId, shouldFetchArchived);
             const memberRequest = Client4.getMyChannelMembers(teamId);
             channels = await channelRequest;
             channelMembers = await memberRequest;
@@ -532,7 +528,7 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
             {
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBERS,
                 data: channelMembers,
-                sync: true,
+                sync: !shouldFetchArchived,
                 channels,
                 serverVersion: getServerVersion(getState()),
                 remove: getChannelsIdForTeam(getState(), teamId),
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -496,7 +496,8 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
 
         let channels;
         let channelMembers;
-        const shouldFetchArchived = (isMinimumServerVersion(getServerVersion(getState()), 5, 21));
+        const state = getState();
+        const shouldFetchArchived = isMinimumServerVersion(getServerVersion(state), 5, 21);
         try {
             const channelRequest = Client4.getMyChannels(teamId, shouldFetchArchived);
             const memberRequest = Client4.getMyChannelMembers(teamId);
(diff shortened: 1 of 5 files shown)


enahum:
Also this PR does not take care of synching those channels that I left in another client while the mobile app was disconnected. The approach to remove the channel members with the sync flag does.

ethervoid:
@enahum 

> Why do we need to fetch the archived channels at this point?

Well I've been investigating and the main problem is that we have out of sync deleted channels in our storage. So getting the archived channels would update those channels and change the `deleted_at` property properly and are going to get synced and not appear in the interface.

If we remove the membership of the archived channels we're doing something "wrong" because even though you're archiving a channel you're not going to lose your membership on that channel and is needed to perform searchs for example, if the option is enabled. Doing that could lead to more side effects IMHO 

> Also this PR does not take care of syncing those channels that I left in another client while the mobile app was disconnected. The approach to remove the channel members with the sync flag does.
Yes it should do it. I've tried in RN for Android and is doing it properly. In the end, you're not deleting channels from the state so you're syncing everything you get, even the archived channels.

enahum:
I’m going to test and benchmark both approaches with at least 200 archived channels.

ethervoid:
Yes that would be helpful. I'm sure removing the membership is better for performance but having consistent state, based in the our logic, is better to avoid side effects and bugs so is a trade-off. But I'm eager to see the results. How can I help you doing it?

enahum:
> and is needed to perform searchs for example,

Search resulta should be controlled by the server though.

The aim of the mobile team is to improve perf so if we add something that will degrade it, that is a concern, although the only problem with this should really be the size of the payload in the response nothing more.

> How can I help you doing it?

No worries, thanks for offering. I’ll take care of it in a couple of hours

ethervoid:
@enahum Great. Thank you. Let me know if you need anything

enahum:
@ethervoid who is the QA tester on this?

ethervoid:
@enahum I usually first check for the dev review and then I add a QA tester to avoid test it before the review is done. I"ll add one now

src/actions/channels.ts
```diff
@@ -496,7 +497,12 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
         let channels;
         let channelMembers;
         try {
-            const channelRequest = Client4.getMyChannels(teamId);
+            let channelRequest;
+            if (isMinimumServerVersion(getServerVersion(getState()), 5, 21)) {
+                channelRequest = Client4.getMyChannels(teamId, true);
+            } else {
+                channelRequest = Client4.getMyChannels(teamId);
+            }
```
enahum: Maybe this whole block can be simplified.

```js
const state = getState();
const serverVersion = getServerVersion(state);
const shouldFetchArchived = isMinimumServerVersion(serverVersion, 5, 21)l

try {
    const channelRequest = Client.getMyChannels(teamId, shouldFetchArchived); // requests to previous servers should not fail if the flag is set in the query string
...
}

File: src/actions/channels.ts [hunk 2]
```diff
@@ -527,6 +532,9 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
             {
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBERS,
                 data: channelMembers,
+                sync: true,
+                channels,
+                serverVersion: getServerVersion(getState()),
```
enahum: this can also be simplified

```js
{
    type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBERS,
    data: channelMembers,
    sync: shouldFetchArchived,
    channels,
    remove: getChannelsIdForTeam(state, teamId) <- do we still need this???
    currentUserId,
    ....
}
````

File: src/client/client4.ts [hunk 3]
```diff
@@ -1449,7 +1449,14 @@ export default class Client4 {
         );
     };
 
-    getMyChannels = async (teamId: string) => {
+    getMyChannels = async (teamId: string, includeDeleted = false) => {
+        const serverVersion = this.getServerVersion();
+        if (isMinimumServerVersion(serverVersion, 5, 21)) {
+            return this.doFetch(
+                `${this.getUserRoute('me')}/teams/${teamId}/channels${buildQueryString({include_deleted: includeDeleted})}`,
```
enahum: you can keep this without needed to check for the server version as adding a query string should have no  effect  on previous servers, or am I wrong?

src/reducers/entities/channels.ts
```diff
@@ -7,6 +7,7 @@ import {GenericAction} from 'types/actions';
 import {Channel, ChannelMembership, ChannelStats} from 'types/channels';
 import {RelationOneToMany, RelationOneToOne, IDMappedObjects, UserIDMappedObjects} from 'types/utilities';
 import {Team} from 'types/teams';
+import {isMinimumServerVersion} from 'utils/helpers';
```
enahum: if the `sync` flag is set based on fetching or not the archived channels this is not needed here.

src/reducers/entities/channels.ts
```diff
@@ -249,6 +222,19 @@ function myMembers(state: RelationOneToOne<Channel, ChannelMembership> = {}, act
         for (const cm of action.data) {
             nextState[cm.channel_id] = cm;
         }
+
+        // We remove membership from archived channels. In version 5.21 we are
+        // sending the archived channels so we have those channel synchronized
+        // and we don't need this code (MM-22201)
+        if (action.sync && action.channels && !isMinimumServerVersion(action.serverVersion, 5, 21)) {
```
enahum:
```suggestion
        if (action.sync && action.channels) {
```

src/client/client4.ts
ethervoid: Yes, I've tested and no problem sending an unused query parameter. I was playing safe here

src/actions/channels.ts
```diff
@@ -495,8 +496,9 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
 
         let channels;
         let channelMembers;
+        const shouldFetchArchived = (isMinimumServerVersion(getServerVersion(getState()), 5, 21));
```
enahum: ` getState()` is being called three times and `getServerVersion` to times in this action, are we expecting different result values for each time we call them? if not I would suggest using a variable to store the value returned by the function and then use the variable where needed.

src/actions/channels.ts
```diff
@@ -527,6 +528,9 @@ export function fetchMyChannelsAndMembers(teamId: string): ActionFunc {
             {
                 type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBERS,
                 data: channelMembers,
+                sync: !shouldFetchArchived,
+                channels,
+                serverVersion: getServerVersion(getState()),
```
enahum: do we still need to pass this to the reducer, if so why? the same with the `remove` property

ethervoid: `remove` was introduced [here](https://github.com/mattermost/mattermost-redux/commit/019dd7625b9f3960371829d0e075aed6e5ce8a1a) and I thought it was already needed. I can't go back to the initial reasoning to include the remove there is only a commit for that change without a ticket.

src/actions/channels.ts
ethervoid: Changed to use a variable :)

ethervoid:
@enahum thanks to you for your time and dealing with a newbie :heart:

ethervoid:
@srkgupta Hi I think I know what is happening. This PR should be tested in two different ways based on the server version.,

#### Server < 5.21

Server version prior to 5.21 should work without having [these changes](https://github.com/mattermost/mattermost-server/pull/13877) deployed. 

This means that it has to be tested against version 5.20 or lesser of the server, because we're checking [here](https://github.com/mattermost/mattermost-redux/blob/MM-22201/src/actions/channels.ts#L500) which server version we're connected to. 

In your case you have a mix, is detecting a server version 5.22 and is trying to use the new endpoint which is not deployed and that's is why is syncing correctly. 

To know if you're using the correct version check in the Chrome Dev Tools > Network tab and you should see this kind of calls to the server: `http://localhost:9005/api/v4/users/me/teams/ygs8imc7zty4trpc68md58j4ue/channels?include_deleted=false`

#### Server >= 5.21

In this case you need to deploy this [server changes](https://github.com/mattermost/mattermost-server/pull/13877) and test using it.

In this case the calls to the server should be of this kind: `http://localhost:9005/api/v4/users/me/teams/ygs8imc7zty4trpc68md58j4ue/channels?include_deleted=true`

enahum:
Is the webapp build referencing the commit on this PR? If not changes are not visible

ethervoid:
Maybe, given that he's testing in his enviroment, he used the `npm run dev` command to copy into the `node_modules`

PR #1052, opened 2020-02-13
MM-21584 Add initial state for channel categories

hmhealey:
This adds the basic state needed for channel categories in phase 1 of the new sidebar organization.

Right now, the channels are hardcoded on the client, but to mimic how they'll work once they're provided by the server, they're added to `state.entities.channelCategories.byId` when a team member is received and removed when the team is deleted. Eventually, the array of channel IDs will be part of that state, but for now, it's derived from channels state by the new selectors.

One thing to note here is that the autoclosing of DM channels is not supported in this version since there's still discussion of how that will work with the sidebar and because we want to rethink how it's supported since it's really complicated right now (as you can see by looking at the TODOs I left for it).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21584


hmhealey:
I had to remove the collapsed state since it needs to be stored in local storage on the device.

src/reducers/entities/channel_categories.test.js
```diff
@@ -0,0 +1,156 @@
...
+        expect(state['team1-public']).toBeDefined();
+        expect(state['team1-private']).toBeDefined();
+        expect(state['team1-direct_messages']).toBeDefined();
+        expect(state['team2-favorites']).toBeDefined();
+        expect(state['team2-public']).toBeDefined();
+        expect(state['team2-private']).toBeDefined();
+        expect(state['team2-direct_messages']).toBeDefined();
+    });
+
+    test('should remove correspoding categories when leaving a team', () => {
```
devinbinnie: correspoding

src/selectors/entities/channel_categories.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+export function makeFilterChannelsByFavorites() {
+    return createSelector(
+        (state: GlobalState, channels: Channel[]) => channels,
+        (state: GlobalState, channels: Channel[], categoryType: string) => categoryType,
+        (state: GlobalState) => state.entities.preferences.myPreferences,
+        (channels, categoryType, myPreferences) => {
+            const filtered = channels.filter((channel) => {
+                if (categoryType === CategoryTypes.FAVORITES) {
```
devinbinnie: Could this just be simplified to:
```suggestion
                return categoryType === CategoryTypes.FAVORITES && isFavoriteChannel(myPreferences, channel.id);
```

src/selectors/entities/channel_categories.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        channels = filterChannelsByFavorites(state, channels, category.type);
+
+        channels = filterChannelsByType(state, channels, category.type);
+
+        channels = filterAutoclosedDMs(state, channels, category.type);
+
+        channels = filterManuallyClosedDMs(state, channels);
+
+        if (channels.some((channel) => channel.type === General.DM_CHANNEL || channel.type === General.GM_CHANNEL)) {
```
devinbinnie: How do we do filtering per different custom categories? Or is that logic not happening yet?

src/reducers/entities/channel_categories.ts
```diff
@@ -0,0 +1,185 @@
...
+        return members.reduce((nextState, member) => {
+            // Note that this adds new categories before state to prevent overwriting existing categories
+            return {
+                ...makeDefaultCategories(member.team_id),
+                ...nextState,
+            };
+        }, state);
+    }
+
+    // case ChannelCategoryTypes.RECEIVED_CATEGORIES: {
```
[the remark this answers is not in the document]
hmhealey: I'll uncomment these once we start getting categories from the server. I started to write them, but I wanted to get the rest of this in and usable to unblock everyone else before I spend time testing it and making sure it works

src/selectors/entities/channel_categories.ts
hmhealey: No, that wouldn't work since it would make it always return false when the category isn't Favourites. `(categoryType === CategoryTypes.FAVORITES) === isFavoriteChannel(myPreferences, channel.id)` is more like what you're looking for, but that just looks awful with the nested equals.

src/selectors/entities/channel_categories.ts
hmhealey: That logic isn't in yet. Right now, the "categories" in the store just act as hardcoded placeholders. Once the channel IDs for each category come from the server, then this function will do that sorting. It might actually make it a bit simpler than the current logic since the stored IDs will be mostly sorted and filtered for us.

PR #1059, opened 2020-02-25
[MM-21408] Get correct display name for DM and modals

hahmadia:
#### Summary

When the System Admin locks the `lock teammate name display` and updates the teammate name display option in the system console, the DM names and the manage members modal names were not updating accordingly. This PR aims to resolve this issue.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21408

#### Screenshots (The names in the red squares update accordingly when system admin updates the teammate name display and locks it in system console)

image


src/selectors/entities/preferences.ts
```diff
@@ -89,14 +89,14 @@ export const getVisibleGroupIds = reselect.createSelector(
...
+    (config, preferences, license) => {
+        const useAdminTemmateNameDisplaySetting = (license && license.LockTeammateNameDisplay === 'true') && config.LockTeammateNameDisplay === 'true';
         const key = getPreferenceKey(Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT);
-        if (preferences[key]) {
+        if (preferences[key] && !useAdminTemmateNameDisplaySetting) {
             return preferences[key].value;
-        } else if (config.TeammateNameDisplay) {
-            return config.TeammateNameDisplay;
         }
-        return General.TEAMMATE_NAME_DISPLAY.SHOW_USERNAME;
```
hahmadia: I'm removing this because `config.TeammateNameDisplay` will always have a value associated to it as it gets set in `setDefaults` in server.

https://github.com/mattermost/mattermost-server/blob/master/model/config.go#L2634

src/selectors/entities/preferences.ts
```diff
@@ -89,14 +89,14 @@ export const getVisibleGroupIds = reselect.createSelector(
 export const getTeammateNameDisplaySetting = reselect.createSelector(
     getConfig,
     getMyPreferences,
-    (config, preferences) => {
+    getLicense,
+    (config, preferences, license) => {
+        const useAdminTemmateNameDisplaySetting = (license && license.LockTeammateNameDisplay === 'true') && config.LockTeammateNameDisplay === 'true';
```
lieut-data: Should this be `useAdminTeammateNameDisplaySetting`? (typo)

src/selectors/entities/preferences.ts
lieut-data: It's true that this will default to being that value, but I can configure `config.json` to set this value to an empty string and the server won't rewrite it. Should we keep the default to handle this case?

enahum:
Did not review it, but have you confirm this works on the mobile app as well as nothing else breaks?

src/selectors/entities/users.test.js
```diff
@@ -458,7 +458,19 @@ describe('Selectors.Users', () => {
         // Should show full name since preferences is being used and LockTeammateNameDisplay is false
         assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'First Last');
 
-        newTestState.entities.general.config.LockTeammateNameDisplay = 'true';
+        newTestState = {
```
lieut-data: Can you elaborate on this change?

src/selectors/entities/preferences.ts
```diff
@@ -89,14 +89,14 @@ export const getVisibleGroupIds = reselect.createSelector(
 export const getTeammateNameDisplaySetting = reselect.createSelector(
```
lieut-data: Can you add unit tests to verify the new license-handling behaviour? I suspect some of the tests removed from `makeGetDisplayName` would apply here now instead.

src/selectors/entities/preferences.ts
lieut-data: Actually, I stand corrected on the substance of this point: the server fails to start if this field is empty. I wrote this comment before testing, then forgot to remove it :D

0/5 on where the default lives, though I think it might still be valuable to be explicit here.

hahmadia: Ah my bad! I totally forgot we can do that so I wasn't taking that into account. Addressed!

src/selectors/entities/preferences.ts
hahmadia: Done! Thanks

src/selectors/entities/users.test.js
lieut-data: Talked with Hossein offline, and concluded this is necessary due to memoization: Redux was caching the results here.

hahmadia: I have updated the way I'm doing this to make it cleaner. Thanks a lot for clarifying why this is the correct solution! Appreciate it.

src/selectors/entities/users.test.js
```diff
@@ -458,19 +458,8 @@ describe('Selectors.Users', () => {
...
-                general: {
-                    ...newTestState.entities.general,
-                    config: {
-                        ...newTestState.entities.general.config,
-                        LockTeammateNameDisplay: 'true',
-                    },
-                },
-            },
-        };
+        newTestState = JSON.parse(JSON.stringify(newTestState));
```
lieut-data: FWIW, I actually think the old approach was "cleaner", vs. exploiting the side effects of serialization here. Even better, perhaps, would be a separate test case that just duplicates the state definition altogether, vs. trying to "mutate".

src/selectors/entities/preferences.ts
hahmadia: Yes you are right! I have added some test cases to cover having no license and license being set to false to ensure that we handle it properly.

src/selectors/entities/users.test.js
hahmadia: Great idea! I just made a `describe(..)` with many `it(..)` and just had a new state in each one. Was better than mutating the state and making comments cause now I was able to write proper descriptions in the `it(..)`. Thanks for the explanation on the mutation stuff as well. Appreciate it 😄

hahmadia:
> Did not review it, but have you confirm this works on the mobile app as well as nothing else breaks?

@enahum 
I just finished playing with the mobile and the changes seem to reflect perfectly on the mobile.  Prior to this change, the lock teammate name display would not have been working on mobile but after this change it will work. Thanks for the heads up.

hahmadia:
@srkgupta 

I have addressed the mobile problem you mentioned in https://github.com/mattermost/mattermost-mobile/pull/4021

While addressing that issue, I also found a small minor inconsistency in the web app which I also addressed. Note that this inconsistency was not happening in the mobile app so I thought I would fix it to make the two consistent with each other. https://github.com/mattermost/mattermost-webapp/pull/5024

PR #1060, opened 2020-02-28
[MM-22276] Add roles constants

hahmadia:
#### Summary

This ticket is a part of https://github.com/mattermost/mattermost-webapp/pull/4881 where it was suggested to create a constants for roles.

### DIFF
--- a/src/types/roles.ts
+++ b/src/types/roles.ts
@@ -1,6 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export type Roles = 'members' | 'guests' | 'admins';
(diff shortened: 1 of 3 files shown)


src/types/roles.ts
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+export type ChannelModerationRoles = 'members' | 'guests';
```
mkraft: Can this use the constants?

hahmadia: I tried but it doesn't seem to work. I also couldn't find anywhere where we use constants for types. 
image

PR #1061, opened 2020-02-29
Fix the type definition for emoji

cometkim:
#### Summary

Moved some utils and fixed type definitions during working on the emoji picker.

### DIFF
--- a/src/utils/emoji_utils.ts
+++ b/src/utils/emoji_utils.ts
@@ -1,14 +1,23 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+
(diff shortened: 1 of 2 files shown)


src/utils/emoji_utils.ts
```diff
@@ -1,14 +1,23 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+
 import {Client4} from 'client';
 import {Emoji, SystemEmoji, CustomEmoji} from 'types/emojis';
+
+export function isSystemEmoji(emoji: Emoji): emoji is SystemEmoji {
```
reflog: nice!

PR #1062, opened 2020-03-01
MM-10272: Handle pinned post count when a post is pinned ot unpinned.

sowmiyamuthuraman:
#### Summary
This PR updates the count of the pinned posts if the post is pinned or unpinned by the user.

### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-10272
 https://github.com/mattermost/mattermost-server/issues/13817

### DIFF
--- a/src/actions/posts.test.js
+++ b/src/actions/posts.test.js
@@ -6,6 +6,7 @@ import assert from 'assert';
 import nock from 'nock';
 
 import * as Actions from 'actions/posts';
+import {getChannelStats} from 'actions/channels';
 import {login} from 'actions/users';
 import {setSystemEmojis, createCustomEmoji} from 'actions/emojis';
 import {Client4} from 'client';
@@ -1174,6 +1175,12 @@ describe('Actions.Posts', () => {
         const {dispatch, getState} = store;
 
         nock(Client4.getBaseRoute()).
+        get(`/channels/${TestHelper.basicChannel.id}/stats`).
+        reply(200, {channel_id: TestHelper.basicChannel.id, member_count: 1, pinnedpost_count: 0});
+
+        await dispatch(getChannelStats(TestHelper.basicChannel.id));
+
+        nock(Client4.getBaseRoute()).
             post('/posts').
             reply(201, TestHelper.fakePostWithId(TestHelper.basicChannel.id));
         const post1 = await Client4.createPost(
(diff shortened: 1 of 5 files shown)


Willyfrog:
adding qa, although it needs https://github.com/mattermost/mattermost-webapp/pull/4917 to be fully tested

Willyfrog:
I'm wondering if this case is covered as I might be missing something: suppose I have two clients open, if I pin the post in one, will the other client be updated as well?

src/actions/posts.ts
```diff
@@ -482,7 +482,11 @@ export function pinPost(postId: string) {
                 ...post,
                 is_pinned: true,
                 update_at: Date.now(),
-            }));
+            }),
+            {
```
hmhealey: Would you mind fixing the indentation here so that the arguments are indented relative to the function call? I'm going to make a note to look at if ESLint can enforce that for us since I know we're not always consistent with it
```javascript
actions.push(
    receivedPost({
        // ...
    }),
    {
        type: ChannelTypes.INCREMENT_PINNED_POST_COUNT,
        id: post.channel_id,
    }
);
```

src/actions/posts.ts
```diff
@@ -519,7 +523,12 @@ export function unpinPost(postId: string) {
                 ...post,
                 is_pinned: false,
                 update_at: Date.now(),
-            }));
+            }),
+            {
```
hmhealey: Same here

src/reducers/entities/channels.ts
```diff
@@ -524,6 +524,40 @@ function stats(state: RelationOneToOne<Channel, ChannelStats> = {}, action: Gene
 
         return state;
     }
+    case ChannelTypes.INCREMENT_PINNED_POST_COUNT: {
+        const nextState = {...state};
+        const id = action.id;
+        const nextStat = nextState[id];
```
hmhealey: I understand the purpose of this, but it seems like a typo that's going to cause us problems down the line. Perhaps we could just refer to it as `nextState[id]` all over to avoid the weird naming?

hmhealey: Oh, that's the existing wording used in the rest of the reducer. This is fine to leave as-is then since it's better to be consistent.

Willyfrog:
> I'm wondering if this case is covered as I might be missing something: suppose I have two clients open, if I pin the post in one, will the other client be updated as well?

@sowmiyamuthuraman have you had the chance to test this?

sowmiyamuthuraman:
> > I'm wondering if this case is covered as I might be missing something: suppose I have two clients open, if I pin the post in one, will the other client be updated as well?
> 
> @sowmiyamuthuraman have you had the chance to test this?

@hmhealey suggested me to update the websocket event for the pin/unpin post to handle this. I have pushed a commit to web-app making the required changes.The conversations regarding this are in this [thread](https://github.com/mattermost/mattermost-server/issues/13817).
Can yot please review it?

hmhealey:
@jgilliam17 For testing this one, the changes are included with https://github.com/mattermost/mattermost-webapp/pull/4917, so they can be tested together

src/actions/posts.test.js
```diff
@@ -293,9 +294,9 @@ describe('Actions.Posts', () => {
         assert.ok(!reactions[post1.id]);
     });
 
-    it('removePost', async () => {
+    it.only('removePost', async () => {
```
hmhealey: Could you remove the `.only` here to ensure all the tests in this file are ran?
```suggestion
    it('removePost', async () => {
```

sowmiyamuthuraman: oops! My bad:( will do that!

esethna:
@amyblais @hmhealey reminder to resolve the JIRA tickets and assign the fix version there when we merge community PRs.

amyblais:
@esethna The webapp PR is still open: https://github.com/mattermost/mattermost-webapp/pull/4917 so I don't think the Jira ticket should be resolved yet?

Also, the person who merges a PR should resolve the Jira ticket (I don't necessarily know what repro steps should be added to the QA test steps).

esethna:
^ thanks, missed that the webapp PR wasn't merged yet

PR #1063, opened 2020-03-02
MM-22752 Preserve other teams channel membership

enahum:
#### Summary
With the recent changes to the mobile performance we need to keep previous channel memberships so whenever you switch teams while being offline the channel list contains all your channels (even when not in sync).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22752

### DIFF
--- a/src/reducers/entities/channels.ts
+++ b/src/reducers/entities/channels.ts
@@ -246,18 +246,6 @@ function myMembers(state: RelationOneToOne<Channel, ChannelMembership> = {}, act
             nextState[cm.channel_id] = cm;
         }
 
-        // We remove membership from archived channels. In version 5.21 we are
-        // sending the archived channels so we have those channel synchronized
-        // and we don't need this code (MM-22201)
-        if (action.sync && action.channels) {
-            const channelsId = action.channels.map((c: any) => c.id);
-            Object.keys(nextState).forEach((channelId) => {
-                if (!channelsId.includes(channelId)) {
-                    Reflect.deleteProperty(nextState, channelId);
-                }
-            });
-        }
-
         return nextState;
     }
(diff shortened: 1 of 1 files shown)


src/reducers/entities/channels.ts
```diff
@@ -389,7 +377,7 @@ function myMembers(state: RelationOneToOne<Channel, ChannelMembership> = {}, act
         if (sync) {
             current.forEach((member: ChannelMembership) => {
                 const id = member.channel_id;
-                if (!channelMembers.find((cm: ChannelMembership) => cm.channel_id === id)) {
+                if (channelMembers.find((cm: ChannelMembership) => cm.channel_id === id)) {
```
enahum: this accomplishes the same thing as `remove` in the `RECEIVED_MY_CHANNEL_MEMBERS` action

src/reducers/entities/channels.ts
```diff
@@ -246,18 +246,6 @@ function myMembers(state: RelationOneToOne<Channel, ChannelMembership> = {}, act
             nextState[cm.channel_id] = cm;
         }
 
-        // We remove membership from archived channels. In version 5.21 we are
-        // sending the archived channels so we have those channel synchronized
-        // and we don't need this code (MM-22201)
-        if (action.sync && action.channels) {
```
enahum: when the websocket reconnects it will remove all channel memberships from other teams that you loaded previously.

chuttam:
* Any risk of regression of #1051 ?
* Any impact when running on older server versions?

enahum:
@chuttam as of #1051 there should not be any as the channel should be marked as archived.

older server versions would not have an impact other than not marking a channel as archived, something that happens with the current version.

PR #1069, opened 2020-03-07
MM-22975|MM-22934 Fix mobile archive channel on action and WS reconnect

enahum:
#### Summary
When a channel was archived the WebSocket handler did not set the `delete_at` property leaving it `undefined` and it was using the wrong `team_id` so that one was also `undefined` thus the channel was not being set as archived.

Also as part of the WS reconnect logic we determine if the user is no longer a member of the current channel by inspecting the response of `getMyChannels`. In servers prior to 5.21 or for those with the `ExperimentalViewArchivedChannels` config set to true, the request does not include the channel so the user was being redirected to Town-Square if for some reason their were sitting at an archived channel.

This PR ensures that the right values for `delete_at` and `team_id` are set when a channel gets archived and that if the user is viewing an archive channel ad the setting is set to **true** then they will remain in the channel.

Note: This WS implementation is only used by the mobile app

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22934
https://mattermost.atlassian.net/browse/MM-22975

### DIFF
--- a/src/actions/websocket.ts
+++ b/src/actions/websocket.ts
@@ -6,7 +6,7 @@ import websocketClient from '../client/websocket_client';
 
 import {ChannelTypes, GeneralTypes, EmojiTypes, PostTypes, PreferenceTypes, TeamTypes, UserTypes, RoleTypes, AdminTypes, IntegrationTypes} from 'action_types';
 import {General, WebsocketEvents, Preferences} from '../constants';
-import {getAllChannels, getChannel, getChannelsNameMapInTeam, getCurrentChannelId, getMyChannelMember, getRedirectChannelNameForTeam, getCurrentChannelStats, getMyChannels, getChannelMembersInChannels, isManuallyUnread} from 'selectors/entities/channels';
+import {getAllChannels, getChannel, getChannelsNameMapInTeam, getCurrentChannelId, getCurrentChannel, getMyChannelMember, getRedirectChannelNameForTeam, getCurrentChannelStats, getMyChannels, getChannelMembersInChannels, isManuallyUnread} from 'selectors/entities/channels';
 import {getConfig} from 'selectors/entities/general';
 import {getAllPosts, getPost as getPostSelector} from 'selectors/entities/posts';
 import {getDirectShowPreferences} from 'selectors/entities/preferences';
@@ -138,6 +138,7 @@ export function doReconnect(now: number) {
             await dispatch(getMyTeamMembers());
             const currentTeamMembership = getCurrentTeamMembership(getState());
             if (currentTeamMembership) {
+                const currentChannel = getCurrentChannel(state);
                 const fethcResult = await dispatch(fetchMyChannelsAndMembers(currentTeamId));
                 const data = (fethcResult as any).data || null;
                 dispatch(loadProfilesForDirect());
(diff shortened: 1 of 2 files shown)


agnivade:
I don't think I have sufficient knowledge in the codebase to make a review.

enahum:
Assigned @migbot instead

src/actions/websocket.ts
```diff
@@ -611,11 +616,8 @@ function handleChannelDeletedEvent(msg: WebSocketMessage) {
                 }
                 EventEmitter.emit(General.DEFAULT_CHANNEL, '');
             }
-
-            dispatch({type: ChannelTypes.RECEIVED_CHANNEL_DELETED, data: {id: msg.data.channel_id, team_id: msg.data.team_id, viewArchivedChannels}});
-
-            dispatch(fetchMyChannelsAndMembers(currentTeamId));
```
ethervoid: Why did you remove this dispatch? Is not needed?

src/actions/websocket.ts
```diff
@@ -611,11 +616,8 @@ function handleChannelDeletedEvent(msg: WebSocketMessage) {
                 }
                 EventEmitter.emit(General.DEFAULT_CHANNEL, '');
             }
-
```
ethervoid: What is the purpose of change the code from here to above?

src/actions/websocket.ts
```diff
@@ -138,15 +138,18 @@ export function doReconnect(now: number) {
...
             if (currentTeamMembership) {
+                const currentChannel = getCurrentChannel(state);
                 const fethcResult = await dispatch(fetchMyChannelsAndMembers(currentTeamId));
                 const data = (fethcResult as any).data || null;
                 dispatch(loadProfilesForDirect());
 
                 if (data && data.channels && data.members) {
                     const channelStillExists = data.channels.find((c: Channel) => c.id === currentChannelId);
                     const stillMemberOfCurrentChannel = data.members.find((m: ChannelMembership) => m.channel_id === currentChannelId);
+                    const config = getConfig(getState());
```
ethervoid: Shouldn't be used `state` variable here instead of `getState` again?

enahum: As this is happening on reconnect, there is another event that gets triggered if the server version has changed here https://github.com/mattermost/mattermost-mobile/blob/master/app/init/global_event_handler.js#L247 So I want to make sure we are using the latest config.

src/actions/websocket.ts
enahum: Basically we were only dispatching if the channel being removed was part of the current team, so if a channel from another team was removed we were not updating the store and on previous servers that do not return the archived channels then this removed channel was never archived.

src/actions/websocket.ts
enahum: We already have the channels at this point, so I don't see a need to `re-fetch` them, if you think otherwise, let me know why and I can change it back.

chuttam:
Assigning to @enahum for cherry-pick merge conflicts.

While we're stopped, should we add QA Review?

enahum:
@chuttam QA review is assigned to the Pr that makes use of this code, which I will update with the right refs once I resolve this conflicts

PR #1070, opened 2020-03-09
[MM-23103] Update selectTeam function to accept teamId string and Team object

hahmadia:
#### Summary

Update `selectTeam` function to accept a `string` along with a `Team` object and make it align more with `selectChannel` which accepts a `string` of `channelId`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23103

#### This change is for 

WebApp https://github.com/mattermost/mattermost-webapp/pull/5019

### DIFF
--- a/src/action_types/channels.ts
+++ b/src/action_types/channels.ts
@@ -63,7 +63,6 @@ export default keyMirror({
     UPDATE_CHANNEL_PURPOSE: null,
     CHANNEL_MEMBER_ADDED: null,
     CHANNEL_MEMBER_REMOVED: null,
-    RESET_CURRENT_CHANNEL_ID: null,
 
     INCREMENT_TOTAL_MSG_COUNT: null,
     INCREMENT_UNREAD_MSG_COUNT: null,
--- a/src/action_types/channels.ts
+++ b/src/action_types/channels.ts
@@ -63,6 +63,7 @@ export default keyMirror({
     UPDATE_CHANNEL_PURPOSE: null,
     CHANNEL_MEMBER_ADDED: null,
     CHANNEL_MEMBER_REMOVED: null,
+    RESET_CURRENT_CHANNEL_ID: null,
 
     INCREMENT_TOTAL_MSG_COUNT: null,
     INCREMENT_UNREAD_MSG_COUNT: null,
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -33,6 +33,12 @@ export function selectChannel(channelId: string) {
     };
 }
 
+export function resetCurrentChannelId() {
+    return {
+        type: ChannelTypes.RESET_CURRENT_CHANNEL_ID,
+    };
+}
+
 export function createChannel(channel: Channel, userId: string): ActionFunc {
     return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         let created;
--- a/src/actions/channels.ts
+++ b/src/actions/channels.ts
@@ -33,12 +33,6 @@ export function selectChannel(channelId: string) {
     };
 }
 
-export function resetCurrentChannelId() {
-    return {
-        type: ChannelTypes.RESET_CURRENT_CHANNEL_ID,
-    };
-}
-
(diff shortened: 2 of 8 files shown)


enahum:
@hahmadia can you explain what is it that we are trying to solve here? I could not really understand the ticket.

Also if you need to clear the channel selection is not enough to do
```js
dispatch({type: ChannelTypes.SELECT_CHANNEL, data: ''});
```

hahmadia:
> @hahmadia can you explain what is it that we are trying to solve here? I could not really understand the ticket.
> 
> Also if you need to clear the channel selection is not enough to do
> 
> ```js
> dispatch({type: ChannelTypes.SELECT_CHANNEL, data: ''});
> ```

@enahum 

I could do that and it is what I had initially but it's a little confusing. The problem is the following:

You view a channel and then click go to `System Console > Channels` and select the last channel you viewed. You edit any of the settings and press `Save` , you will be re-directed to that channel after the save. This is due to the fact that https://github.com/mattermost/mattermost-webapp/blob/master/actions/websocket_actions.jsx#L451 gets triggered. So I fix the redirection by resetting the the current channel id so it doesn't get triggered. 

The reason I created a new action is because in the web app PR ( https://github.com/mattermost/mattermost-webapp/pull/5019 ) , it was a little confusing to put `actions.selectChannel('')` and I thought it would be more clear to have something like `actions.resetCurrentChannelId` instead. However, if you think this is not appropriate, I'm more than happy to do what you suggested.

enahum:
This seems risky either way.. what else gets impacted with this change. 0/5 on the action name.

enahum:
Adding @hmhealey just in case he can think of potential side effects

enahum:
Another question is why do we have a redirect in the WS action.

hahmadia:
> Another question is why do we have a redirect in the WS action.

@enahum 
The change happened in this PR https://github.com/mattermost/mattermost-webapp/pull/3077 . Seems to be for channel renaming issues.

enahum:
@hmhealey maybe the WS approach to fix the linked webapp PR was not the best solution? what about redirecting only if the channel name changed somewhere in the Channel view component ?

hmhealey:
I'd lean towards using an existing redux action like @enahum suggested above when you go into the system console instead of creating a new action that does the same thing just without an argument. We may have other code looking for SELECT_CHANNEL actions (like some stuff I've been looking at recently), so I don't think we should have two ways of doing that. I like the idea of clearing the currentChannelId (and maybe the currentTeamId too?) when you leave the channel view though.

Regarding the previous PR, it does seem like there might be a better way to do that, but getting redirected by websocket events happens elsewhere as well (like when a team/channel is deleted or the user is removed from either of those), so I think just clearing the currentChannelId and currentTeamId when you're not in the channel view would be easier.

hahmadia:
@hmhealey @enahum 
To addressed Harrison's comment on my other PR about resetting the `currentTeamId`, I needed to update the `selectTeam` to align it with our `selectChannel` action. In order to avoid refactoring a lot of our code which is already passing a `Team` type, I decided to add a extra `string` type and check to see if the passed in parameter should use `team` or `team.id` as the `teamId`. Thanks!

enahum:
I lost track of what this PR tries to do, at first it was about unselect the current channel, now is about the current team, and I only see two lines of code being changed.

@hahmadia can you please clarify?

Thanks my friend.

hahmadia:
@enahum 
Sure! So initially it was adding a new action to reset the `currentChannelId` but as you and Harrison mentioned, it would be better to use the current `selectChannel('')` to reset that which I did https://github.com/mattermost/mattermost-webapp/pull/5019 and Harrison suggested it wouldn't be a bad idea to also reset the `currentTeamId` as well but `selectTeam` function accepts a `Team` object however I wanted to make it so it can accept both a `Team` and `string` so I could pass in a empty `TeamId` string to reset the `currentTeamId`. This also aligns up with the way `selectChannel` function works which is accepting a `channelId` string.

PR #1074, opened 2020-03-10
Update Dependencies

grundleborg:
#### Summary
Update dependencies.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21581


hmhealey:
Are the changes to the trailing commas in function calls coming from the new ESLint version? We might want to disable that to prevent so many changes. I know the trailing commas rule was entirely disabled on the web app repo as well, so not everyone seems to be a fan of it.

grundleborg:
Oh hmm good question @hmhealey . I don't think it's a particularly disruptive change, but if it's controversial then happy to change the eslint rule and revert (personally I'm a fan of the dangling commas as they make subsequent diffs cleaner though).

grundleborg:
@hmhealey @jespino (re-)requesting a quick dev review on this as I've re-done the updates. This time I will try and get it merged quickly so it doesn't go out of date!

grundleborg:
Changes are separated out into separate commits if you prefer to review that way.

PR #1075, opened 2020-03-10
MM-22577 Re-add autoclosing DMs for new selectors

hmhealey:
This implements the missing channel selector from my last PR. As discussed, we're just matching the previous behaviour (implemented in `isAutoClosed`), so the code may be a bit difficult to parse.

If it makes it easier to follow, the original code is [here](https://github.com/mattermost/mattermost-redux/blob/fe064495b21157b729761b924771fd64c81f79dd/src/utils/channel_utils.ts#L157) and I also made a flow chart for it [here](https://drive.google.com/file/d/1ReyJR8X8tGMyYtaD_pbTS-MKjyX78cTx/view?usp=sharing)

To ensure that this should be tested correctly (since testing on the original function is light), I've made it so that the tests double check their results against the original function as well.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22577

### DIFF
--- a/src/selectors/entities/channel_categories.test.js
+++ b/src/selectors/entities/channel_categories.test.js
@@ -4,6 +4,13 @@
 import {General, Preferences} from '../../constants';
 import {CategoryTypes} from '../../constants/channel_categories';
 
+import {getCurrentChannelId, getMyChannelMemberships} from 'selectors/entities/channels';
+import {getConfig} from 'selectors/entities/general';
+import {getLastPostPerChannel} from 'selectors/entities/posts';
+import {getMyPreferences} from 'selectors/entities/preferences';
+import {getCurrentUserId} from 'selectors/entities/users';
+
+import {isGroupOrDirectChannelVisible} from 'utils/channel_utils';
 import {getPreferenceKey} from 'utils/preference_utils';
 
 import * as Selectors from './channel_categories';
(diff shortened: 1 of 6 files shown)


src/selectors/entities/channel_categories.test.js
```diff
@@ -191,6 +200,418 @@ describe('makeFilterChannelsByType', () => {
...
+            now,
+        );
+    }
+
+    test('should hide an inactive GM channel', () => {
+        const filterAutoclosedDMs = Selectors.makeFilterAutoclosedDMs(() => cutoff);
+
+        const channel1 = {id: 'channel1', type: General.GM_CHANNEL};
+
+        const state = mergeObjects(baseState, {
```
hmhealey: Also, I'm trying out a new method of merging state objects to see if that makes these test cases more readable when they slightly modify the existing state. Let me know if you have any thoughts on how this looks. You can see the difference between this version and the full state here: https://github.com/mattermost/mattermost-redux/pull/1075/commits/dfda2ac8461fbf0e368f46de49bc7eacacbc7ef4

test/merge_objects.js
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+function isObject(obj) {
+    return obj && typeof obj === 'object' && !Array.isArray(obj);
+}
+
+// Returns the result of merging two objects. If a field is specified in both a and b, the value from b takes precedence
+// unless both values are objects in which case mergeObjects will be called recursively.
+export default function mergeObjects(a, b, path = '.') {
```
devinbinnie: Neat!

src/selectors/entities/channel_categories.test.js
devinbinnie: I think if anything we just need to be louder about the fact that the method exists so that more people use it.
Also document any limitations if they exist

sudheerDev: Nice 👍

src/selectors/entities/channel_categories.ts
```diff
@@ -91,36 +92,98 @@ export function makeFilterChannelsByType(): (state: GlobalState, channels: Chann
...
 
-            // TODO check if Preferences.CATEGORY_CHANNEL_APPROXIMATE_VIEW_TIME > now - 7 days to keep open
-            // TODO check if Preferences.CATEGORY_CHANNEL_OPEN_TIME > now - 7 days to keep open
-            // TODO check if teammate.delete_at > Preferences.CATEGORY_CHANNEL_OPEN_TIME to auto close if not current channel
-            // TODO check if lastPost.create_at > now - 7 days to keep open
-            // TODO check if channel.last_post_at > now - 7 days to keep open
-            // TODO don't autoclose favorited channel, except if teammate was deleted since last opened and not current channel
-            // TODO don't autoclose if autocloseDMs is false, except if teammate was deleted since last opened and not current channel
+            // Ideally, this would come from a selector, but that would cause the filter to recompute too often
+            const cutoff = getAutocloseCutoff();
```
sudheerDev: Nothing to act but This might be funny if user is looking at LHS and a DM just drops off because it exceeded.

src/selectors/entities/channel_categories.ts
```diff
@@ -91,36 +92,98 @@ export function makeFilterChannelsByType(): (state: GlobalState, channels: Chann
     );
 }
 
-export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channel[], categoryType: string) => Channel[] {
+function getDefaultAutocloseCutoff() {
+    return Date.now() - 7 * 24 * 60 * 60 * 1000;
+}
+
+export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutocloseCutoff): (state: GlobalState, channels: Channel[], categoryType: string) => Channel[] {
```
sudheerDev: Nit: Does is makes sense to change this to be `makeFilterAutoclosedDMsGMs` or something of that sorts?

src/selectors/entities/preferences.ts
```diff
@@ -260,7 +260,6 @@ export function shouldAutocloseDMs(state: GlobalState) {
         return false;
     }
 
-    const preference = get(state, Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.CHANNELS_SIDEBAR_AUTOCLOSE_DMS, '');
-
-    return preference.value === Preferences.AUTOCLOSE_DMS_ENABLED;
+    const preference = get(state, Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.CHANNEL_SIDEBAR_AUTOCLOSE_DMS, '');
```
sudheerDev: I know this is not a new change but it is Weird that it always said DMs but infact this supports GMs as well 🤷‍♂. When ever we talk about this func it is hard to relate we do this for GMs as well.

src/selectors/entities/channel_categories.ts
hmhealey: That's actually the current behaviour too. It just only happens if something else causes the selector to recalculate.

src/selectors/entities/channel_categories.ts
hmhealey: I considered that, but most of the other logic just refers to autoclosing DMs even though it handles both GMs and DMs

src/selectors/entities/preferences.ts
hmhealey: I think it's because the setting predates GMs if I remember correctly.

src/selectors/entities/channel_categories.ts
sudheerDev: Never mind. Better to leave it for now IMO too. When i first saw the notification about the review i didnt even realise that it is about GMs too.

PR #1076, opened 2020-03-11
[MM-23095] - Update action to take into consideration system posts

nevyangelova:
#### Summary
Fixes a JS error related to marking a system post as unread. There are two types of posts and system posts have an array of posts ID's, which needs to be taken into account in the mattermost-redux action. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23095

#### Related Pull Requests
- Has webapp changes (https://github.com/mattermost/mattermost-webapp/pull/5042)

### DIFF
--- a/src/actions/posts.ts
+++ b/src/actions/posts.ts
@@ -14,6 +14,7 @@ import {getCurrentUserId, getUsersByUsername} from 'selectors/entities/users';
 import {getUserIdFromChannelName} from 'utils/channel_utils';
 import {parseNeededCustomEmojisFromText} from 'utils/emoji_utils';
 import {isFromWebhook, isSystemMessage, shouldIgnorePost} from 'utils/post_utils';
+import {isCombinedUserActivityPost} from 'utils/post_list';
 
 import {getMyChannelMember, markChannelAsUnread, markChannelAsRead, markChannelAsViewed} from './channels';
 import {systemEmojis, getCustomEmojiByName, getCustomEmojisByName} from './emojis';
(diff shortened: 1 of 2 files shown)


src/actions/posts.ts
```diff
@@ -429,9 +428,11 @@ export function setUnreadPost(userId: string, postId: string) {
                 channelId: post.channel_id,
             },
         });
-
         try {
-            unreadChan = await Client4.markPostAsUnread(userId, postId);
+            if (post.system_post_ids) {
```
hmhealey: We should probably do this before we dispatch the ADD_MANUALLY_UNREAD. Dispatching that adds some behaviour that stops the channel from being marked unread, so if we're preventing you from marking the system message unread, we also should avoid doing that.

src/actions/posts.ts
```diff
@@ -417,10 +417,9 @@ export function getUnreadPostData(unreadChan: ChannelUnread, state: GlobalState)
     return data;
 }
 
-export function setUnreadPost(userId: string, postId: string) {
+export function setUnreadPost(userId: string, post: ExtendedPost) {
```
hmhealey: Changing the parameters of this function might be a bit tricky since we need to update the web and mobile app code that uses this at the same time that this PR goes in or else it might temporarily break that code. You have a PR open to change the code on the web app that uses this, but we'd need another PR to update the mobile app as well.

Alternatively, instead of changing the parameters so that we can check for `post.system_post_ids`, we might instead be able to call [`isCombinedUserActivityPost`](https://github.com/mattermost/mattermost-redux/blob/master/src/utils/post_list.ts#L192) on the post ID to see what type of post it is.

nevyangelova:
Updated

PR #1080, opened 2020-03-12
[MM-23235] Handle user typing event only for current channel

migbot:
#### Summary
Only dispatch actions in `handleUserTypingEvent` for the current channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23235


hmhealey:
@migbot Leaving this to you to merge in case there needs to be an accompanying mobile PR created

PR #1066, opened 2020-03-04
MM-23015: Enable or disable group mentions

catalintomai:
#### Summary
Enable or disable mentions for a group.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23015

Related PRs:
webapp: https://github.com/mattermost/mattermost-webapp/pull/4977
server: https://github.com/mattermost/mattermost-server/pull/14010

PR #1081, opened 2020-03-12
[MM-23159] Highlight Group Names

hahmadia:
## Last commit is the one with my changes!
Builds on top of https://github.com/mattermost/mattermost-redux/pull/1066

#### Summary

This change adds groups that are allowed to be referenced to the keys array of mention keys so that it is highlighted like we do for `@here` and `@channel` and so on. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23159

#### Screenshot
image


src/selectors/entities/users.test.js
```diff
@@ -186,12 +192,50 @@ describe('Selectors.Users', () => {
...
+                        },
+                    },
+                    groups: {
+                        groups: {
+                            test1: {
+                                display_name: 'I_AM_THE_BEST!',
+                                allow_reference: true,
+                            },
+                            test2: {
+                                display_name: 'Do you love me?',
```
fm2munsh: are we operating under the assumption that groups can have spaces in between them?

hahmadia: > are we operating under the assumption that groups can have spaces in between them?

Just a test, I didn't really think of that. I just made a random display name. I can change it to a  non-spaced display name if it's better.

hahmadia: I've made the display names non-spaced.

fm2munsh: It was more of a question - since my backend code will not be able to parse a group display name like this

hahmadia: > It was more of a question - since my backend code will not be able to parse a group display name like this

@fm2munsh 
I agree that it shouldn't be spaced cause that can be confusing. I don't think we can have groups with spaces in them and it's better to align the tests more with how they would actually be.

src/selectors/entities/users.ts
```diff
@@ -119,36 +121,41 @@ export type UserMentionKey= {
...
-        keys.push({key: usernameKey});
-    }
+        const usernameKey = '@' + user.username;
+        if (keys.findIndex((key) => key.key === usernameKey) === -1) {
+            keys.push({key: usernameKey});
+        }
 
-    return keys;
-}
+        groups.forEach((group) => keys.push({key: `@${group.display_name}`}));
```
catalintomai: use "group.name" instead, for now?

hahmadia: yep! I will address this after I finish the confirmation dialogue

PR #1082, opened 2020-03-12
MM-23239 Only show the user's channels in the sidebar

hmhealey:
I missed implementing this when copying from the old selectors. Just checking for a channel member to exist here is enough to ensure that we're a member of the channel since channel members are one of the few things that we delete entirely when a user leaves a channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23239

### DIFF
--- a/src/selectors/entities/channel_categories.ts
+++ b/src/selectors/entities/channel_categories.ts
@@ -6,12 +6,12 @@ import {createSelector} from 'reselect';
 import {General, Preferences} from '../../constants';
 import {CategoryTypes} from '../../constants/channel_categories';
 
-import {getCurrentChannelId} from 'selectors/entities/channels';
+import {getCurrentChannelId, getMyChannelMemberships} from 'selectors/entities/channels';
 import {getCurrentUserLocale} from 'selectors/entities/i18n';
 import {getTeammateNameDisplaySetting, shouldAutocloseDMs} from 'selectors/entities/preferences';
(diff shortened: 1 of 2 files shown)


hmhealey:
I've tried this with the web app feature branch, and it seems like it's not working. I'm currently investigating

src/selectors/entities/channel_categories.ts
```diff
@@ -42,9 +42,13 @@ export function makeGetCategoriesForTeam(): (state: GlobalState, teamId: string)
 export function makeGetChannelsForAllCategories(): (state: GlobalState, teamId: string) => Channel[] {
     return createSelector(
         (state: GlobalState) => state.entities.channels.channels,
+        getMyChannelMemberships,
         (state: GlobalState, teamId: string) => teamId,
-        (allChannels: IDMappedObjects<Channel>, teamId: string) => {
-            return Object.values(allChannels).filter((channel) => channel.team_id === teamId || channel.team_id === '');
+        (allChannels: IDMappedObjects<Channel>, myMembers: RelationOneToOne<Channel, ChannelMembership>, teamId: string) => {
+            return Object.values(allChannels).
+                filter((channel) => channel.delete_at === 0).
```
nevyangelova: 0/5 I would roll the filters into one using ```&&```

hmhealey: Does that have a significant performance difference? I chose to leave them separate for readability, but we could certainly combine them at some point

PR #1083, opened 2020-03-13
MM-23266 Add selector to get sorted channels for all categories

hmhealey:
As explained in the Jira ticket, the sidebar needs to have a sorted list of all channels in the sidebar so that it can use it for things like the alt+arrow hotkeys. We have existing code to return a list like this, but it performs poorly in some cases which we've thought might cause some performance issues when there are a large number of channels in the sidebar, so this is another attempt at rewriting it to be performant.

The strategy of this selector is to:
1. Not use `createSelector` since it will cause everything to recalculate if even one dependency might have changed.
2. Store an instance of `makeFilterAndSortChannelsForCategory` for each category so that getting the channels for each category can be memoized independently. This is something I've wanted to try for a while, but I haven't been able to come up with a generic way to do it. I'm not entirely sure if this will work better, but this seems like a good enough time to try it.
3. If the results of each instance of `makeFilterAndSortChannelsForCategory` don't change, then return the same object. `createSelector` can't do this, so this is another reason not to use it.

This PR is based on multiple other PRs, so the changes specific to it are here (https://github.com/mattermost/mattermost-redux/commit/ff6dea639275b803e2ad961d6358f737b958ce8b). The important changes are in `makeGetChannelsByCategory` 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23266

#### Related PRs
https://github.com/mattermost/mattermost-redux/pull/1075
https://github.com/mattermost/mattermost-redux/pull/1082


src/selectors/entities/channel_categories.test.js
```diff
@@ -71,34 +73,98 @@ describe('makeGetChannelsForAllCategories', () => {
...
+        const getUnsortedUnfilteredChannels = Selectors.makeGetUnsortedUnfilteredChannels();
+
+        const state = {
+            entities: {
+                channels: {
+                    channels: {
+                        channel,
+                    },
+                    myMembers: {
+                        [channel.id]: {},
```
devinbinnie: Should this include membership for `deletedChannel`? I think you're testing this same behaviour elsewhere.

PR #1084, opened 2020-03-13
[MM-22212] Add channel scheme updated socket event

fm2munsh:
#### Summary
Adds a websocket event to dispatch `getMyChannelMember` when a channel_scheme is either created or deleted to update the list of roles tied to that channel.

Related Server PR: https://github.com/mattermost/mattermost-server/pull/14057

Test steps: 
1.  Login to mobile app and open a channel as regular member
2. Login as sysadmin in different tab and make the channel read only
3. Ensure that the channel changes to read only without reload for the regular member


srkgupta:
I was seeing the issues because the following server PR was missed. 
https://github.com/mattermost/mattermost-server/pull/14057

Since the above mentioned PR will be merged to Martin's server PR:
https://github.com/mattermost/mattermost-server/pull/13813

This PR will be revalidated once all the issues mentioned in the Martin's server PR is fixed.

PR #1089, opened 2020-03-17
Channel Sidebar Organization Stage 1

hmhealey:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21296

#### Related PRs
https://github.com/mattermost/mattermost-webapp/pull/4999


src/reducers/entities/channel_categories.ts
```diff
@@ -0,0 +1,185 @@
...
+        return members.reduce((nextState, member) => {
+            // Note that this adds new categories before state to prevent overwriting existing categories
+            return {
+                ...makeDefaultCategories(member.team_id),
+                ...nextState,
+            };
+        }, state);
+    }
+
+    // case ChannelCategoryTypes.RECEIVED_CATEGORIES: {
```
devinbinnie: Can we remove this code? Or are we keeping it for Phase 2?

src/selectors/entities/channel_categories.ts
```diff
@@ -0,0 +1,375 @@
...
+                return true;
+            });
+
+            return filtered.length === channels.length ? channels : filtered;
+        }
+    );
+}
+
+function getDefaultAutocloseCutoff() {
+    return Date.now() - 7 * 24 * 60 * 60 * 1000;
```
devinbinnie: Non-blocking, but maybe the 7 days should be a constant somewhere?

src/reducers/entities/channel_categories.ts
hmhealey: We'll keep it for phase 2 for now

src/selectors/entities/channel_categories.ts
hmhealey: That would certainly be helpful, but I'd say it's out of scope of this PR. Maybe `moment` provides some sort of constants for this

src/reducers/entities/channel_categories.ts
sudheerDev: Nit: A comment to indicate that it will be use later for phase 2?

src/selectors/entities/channel_categories.ts
```diff
@@ -0,0 +1,375 @@
...
+    return state.entities.channelCategories.orderByTeam[teamId];
+}
+
+export function makeGetCategoriesForTeam(): (state: GlobalState, teamId: string) => ChannelCategory[] {
+    return createSelector(
+        getCategoryIdsForTeam,
+        (state: GlobalState) => state.entities.channelCategories.byId,
+        (categoryIds, categoriesById) => {
+            if (!categoryIds) {
+                return [];
```
sudheerDev: Out of curiosity,  when would we not have  `getCategoryIdsForTeam`?

src/types/channel_categories.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type ChannelCategoryType = 'favorites' | 'public' | 'private' | 'direct_messages' | 'custom';
+
+export type ChannelCategory = {
+    id: string;
+    team_id: $ID<Team>;
+    type: ChannelCategoryType;
+    display_name: string;
+
+    // channel_ids: $ID<Channel>;
```
sudheerDev: isn't needed anymore?

src/selectors/entities/channel_categories.ts
hmhealey: It might happen in the future when these are loaded from the server, but they should always be set now. It's also just error proofing since you could call the selector with any random string

src/types/channel_categories.ts
hmhealey: This will be added in phase 2 once the server supplies the IDs. I'll add a couple comments

PR #1091, opened 2020-03-19
[MM-21517] Add new options parameter to getTeamMembers to fix users not having their roles updated

hahmadia:
#### Summary

The backend will accept parameters for how to sort the team members and if to exclude deleted users or not.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21517

The backend PR which implements this is https://github.com/mattermost/mattermost-server/pull/14043

### DIFF
--- a/src/actions/teams.ts
+++ b/src/actions/teams.ts
@@ -299,7 +299,7 @@ export function getMyTeamMembers(): ActionFunc {
     };
 }
 
-export function getTeamMembers(teamId: string, page = 0, perPage: number = General.TEAMS_CHUNK_SIZE): ActionFunc {
+export function getTeamMembers(teamId: string, page = 0, perPage: number = General.TEAMS_CHUNK_SIZE, options = {}): ActionFunc {
     return bindClientFunc({
         clientFunc: Client4.getTeamMembers,
         onRequest: TeamTypes.GET_TEAM_MEMBERS_REQUEST,
@@ -309,6 +309,7 @@ export function getTeamMembers(teamId: string, page = 0, perPage: number = Gener
             teamId,
             page,
             perPage,
+            options,
         ],
     });
 }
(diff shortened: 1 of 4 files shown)


src/actions/teams.ts
```diff
@@ -299,7 +299,7 @@ export function getMyTeamMembers(): ActionFunc {
     };
 }
 
-export function getTeamMembers(teamId: string, page = 0, perPage: number = General.TEAMS_CHUNK_SIZE): ActionFunc {
+export function getTeamMembers(teamId: string, page = 0, perPage: number = General.TEAMS_CHUNK_SIZE, options = {}): ActionFunc {
```
fm2munsh: I would opt for creating a type `GetTeamMembersOpts` instead of `options = {}` 
something like 
```
type GetTeamMembersOpts = {
    sort: 'Username' | undefined;
    exclude_deleted_users: bool | undefined;
}
```

sbishel: Also 'Username' and 'UserId' should be constants.

fm2munsh: I dont know if constants can be used in a type definition i seem to remember having issues with that

hahmadia: Thanks for the suggestion! 
@sbishel I did make Username a constant since I do use it in the webapp however seems like you cannot use it within a type definition. Thanks for the suggestion.

@fm2munsh I did use the type you recommended however I did not include 'UserId' as that is default behaviour when you don't pass anything to sort.

hahmadia:
@fm2munsh @sbishel 
Any opinions on the `SORT_USERNAME_OPTION` and me putting it in the `teams/constants`? I wasn't sure what to name it nor where to place to it. Felt that was most appropriate but maybe you guys have other thoughts?

PR #1094, opened 2020-03-24
Adding call for purge bleve indexes

jespino:
#### Summary
Adding call for purge bleve indexes

#### Ticket Link
[MM-23144](https://mattermost.atlassian.net/browse/MM-23144)


src/action_types/admin.ts
```diff
@@ -135,6 +135,10 @@ export default keyMirror({
     PURGE_ELASTICSEARCH_INDEXES_SUCCESS: null,
     PURGE_ELASTICSEARCH_INDEXES_FAILURE: null,
 
+    PURGE_BLEVE_INDEXES_REQUEST: null,
+    PURGE_BLEVE_INDEXES_SUCCESS: null,
+    PURGE_BLEVE_INDEXES_FAILURE: null,
+
```
enahum: Any reason why you are tracking the request ?

src/actions/admin.ts
```diff
@@ -379,6 +379,15 @@ export function purgeElasticsearchIndexes(): ActionFunc {
     });
 }
 
+export function purgeBleveIndexes(): ActionFunc {
```
enahum: does this action need to live in this library? cannot be added to the Webapp actions instead?

jespino: I added it here because consistency, but probably makes sense to move the whole `admin.ts` to `mattermost-webapp`

src/action_types/admin.ts
jespino: No, is not needed, I'll remove it. I was following the pattern of the other purge command (for elasticsearch).

src/actions/admin.ts
enahum: I cannot agree more with that!

src/action_types/admin.ts
enahum: Yes, if you  don't need to track is better not to add them, as each action will be dispatch causing multiple unnecessary mapStateToProps to run

src/actions/admin.ts
jespino: For now I going to close this PR and move this changes to the webapp, but I think we can't remove the other actions from here before 6.0.

jespino: Actually, I still need the changes for the client, so I going to remove the actions and keep the client function.

PR #1105, opened 2020-03-27
MM-23569: Add rudder to diagnostics feature.

grundleborg:
#### Summary
Adds Rudder to diagnostics functionality in addition to Segment (Segment will be removed later).

Once this is in, there are WebApp changes to follow.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23569


enahum:
@grundleborg should this be added to RN as well ?

grundleborg:
@enahum you guys have the ticket for Rudder support on mobile, so up to you to figure that out :)

grundleborg:
Less smart-arse answer: I don't actually know how Segment works on mobile, but at some point when you guys get to adding Rudder support there, you'll need to do whatever is necessary - if you use the `trackEvent()` function on the Client class like the webapp does, then something similar to this will be needed (but probably just ripping out Segment altogether and just adding Rudder, since we'll already have piloted it on the server/webapp by then I'd imagine.

enahum:
Well, this will certainly be a thing, now the ticket is too vague, I've asked BizOps to add any necessary details to the ticket.

PR #1106, opened 2020-03-27
fix fakeDmChannel to ensure correct user id order

lieut-data:
#### Summary
Ensure `fakeDMChannel` generates a channel name that matches the semantics of putting the smaller (lexicographically) user id first, avoiding a flaky test failure when comparing against `fakeUserId1`.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-22923


lieut-data:
Skipping `QA Review`, since this change only touches the test case utility functions.

PR #1113, opened 2020-04-01
MM-23395: Add websockets support for update group for group mentions

catalintomai:
#### Summary
Ensuring that when we delete or update a MM group, the websocket event is triggered and we are updating the referenceable groups (et al) in Redux for group-mentions + fixed bug that was updating the groups state with incomplete group metadata(only the id) when mapping a team/channel to a group.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23395 Note: this addresses items(events) 1-9 of the ticket - see ticket comments.

#### Related PRs
webapp: https://github.com/mattermost/mattermost-webapp/pull/5245
server: https://github.com/mattermost/mattermost-server/pull/14202

Note: 1) removed the mobile changes: https://github.com/mattermost/mattermost-mobile/pull/4116 from this PR since they require extra consideration and are not blocking for the current set for webapp.

Note 2) Re-requested dev review since significant amount of code has been added following QA feedback.


enahum:
@catalintomai interestingly the WS implementation on this repo was only being used by the mobile app, but as I announced before, the mobile app no longer uses this repo to hold the code base.

catalintomai:
> @catalintomai interestingly the WS implementation on this repo was only being used by the mobile app, but as I announced before, the mobile app no longer uses this repo to hold the code base.

Thanks @enahum , If I get this correctly, that means that for mobile, if we need to support group mentioning/invites we'll have to add websockets support in the new repo(mm-mobile/app/redux?) - or do we? - I see that you've removed websockets from mobile redux:
https://github.com/mattermost/mattermost-mobile/pull/4088/commits/4af22f2545e79ecb821b87366427ddf6e7430499?file-filters%5B%5D=.ts<
i

enahum:
@catalintomai to answer your question yes! anything on this repo that needs to be implemented in mobile, has to be done directly in mobile.

as for the Websocket it lives here https://github.com/mattermost/mattermost-mobile/blob/master/app/actions/websocket.ts

catalintomai:
> @catalintomai to answer your question yes! anything on this repo that needs to be implemented in mobile, has to be done directly in mobile.
> 
> as for the Websocket it lives here https://github.com/mattermost/mattermost-mobile/blob/master/app/actions/websocket.ts

Thanks @enahum, will move these changes to the mobile redux. Is there a recommended way to validate websocket-related redux changes on mobile?

src/constants/general.ts
```diff
@@ -68,4 +68,5 @@ export default {
     DISABLED: 'disabled',
     DEFAULT_ON: 'default_on',
     DEFAULT_OFF: 'default_off',
+    DEFAULT_GROUP: 'board',
```
mkraft, line 71: What's `'board'`?

catalintomai: > What's `'board'`?

@mkraft - It is the name of a dummy group (note: DEFAULT_GROUP is used in websocket.test.js only,). There doesn't seem to be a separate file for redux constants for test only, so we can either create one or give this a different name like DEFAULT_TEST_GROUP?

mkraft: Ok, it is the test 'board'. It threw me off that it was in a non-test file. 0/5 from my perspective.

catalintomai:
> Tested the PR along with its dependent server and webapp PRs and the websocket support is provided as expected when Group Mentions are enabled/disabled/unlinked/removed during LDAP Group sync. Approving the PR.

@srkgupta, are you doing more testing to validate the comments of the original ticket (https://mattermost.atlassian.net/browse/MM-23395) - basically that we do not need websocket support for the rest in the list or are we ok with merging this PR in and treat that validation effort separately.

src/actions/channels.ts
```diff
@@ -249,7 +249,6 @@ export function patchChannel(channelId: string, patch: Channel): ActionFunc {
             ]));
             return {error};
         }
-
```
hahmadia, line 252: nit: could we add this back in?

src/reducers/entities/groups.ts
```diff
@@ -28,13 +28,14 @@ function syncables(state: Dictionary<GroupSyncables> = {}, action: GenericAction
...
+    case GroupTypes.PATCHED_GROUP_TEAM:
     case GroupTypes.LINKED_GROUP_TEAM: {
         let nextGroupTeams: GroupTeam[] = [];
-
-        if (!state[action.data.group_id] || !state[action.data.group_id].teams) {
+        if (!state[action.data.group_id] || !state[action.data.group_id].teams || state[action.data.group_id].teams.length === 0) {
             nextGroupTeams = [action.data];
         } else {
-            nextGroupTeams = {...state}[action.data.group_id].teams;
+            nextGroupTeams = {...state}[action.data.group_id].teams.slice();
```
hahmadia, line 37: I'm confused, why do we do the `.slice()` ?

hahmadia: Not a blocker so I approved but I'm not quite clear on why the `.slice()`. Especially cause we don't pass anything into it.

catalintomai: @hahmadia, slice() creates a new object. I was hitting this weird issue where I could not update the `nextGroupTeams` if I recall correctly, probably because it is a subset of `state`(?). The way out was to create a new version of it to be updated for the next state.

hahmadia: If you need a copy of the array, why not do something like 
`[...{...state}[action.data.group_id].teams]`. Does this not work?

catalintomai: could be, but slice() was the existing pattern - see `UNLINKED_GROUP_TEAM`, `UNLINKED_GROUP_CHANNEL` , not sure of the pros/cons of using one vs. the other.

hahmadia: I'm not sure of the pros and cons either but to me slice seems hacky for some reason. I just know the spread operator is used more commonly but seems there is more ways than one to do the same. All good. I suppose it's personal preference.

PR #1117, opened 2020-04-04
Clean up code style in selectors

hmhealey:
There's a lot of stuff going on here to make the code a bit easier to follow. I'm making the following changes:
1. I reorganized the imports so that they're alphabetical and separated by folder. I'm going to look at including ESLint rules to do this better in the future since we do have a plugin that should be enforcing import order.
2. I fixed some places where `createSelector` was imported as `reselect.createSelector` after the TS migration
3. I added names to all the parameters for selector functions so that instead of them having a type like `(a: GlobalState, b: boolean, c: boolean) => string`, they'll now have types like `(state: GlobalState, unreadsAtTop: boolean, favoritesAtTop: boolean) => string`
4. I made it so that `createSelector` is always on its own line so that you don't end up with really long lines like `createSelector(getCurrentChannelId, getChannels, (state: GlobalState, channelId: string) => channelId), (currentChannelId, channels, channelId) => ...`. This is something else I'll look at loosely enforcing with ESLint.
5. I removed some really old unused channel selectors that weren't used in either app.

None of this should have functional changes.


hmhealey:
Yeah, I assumed it was from the automatic conversion. I might go through the other files at some point since I enjoy doing this sort of mindless cleanup, but these were definitely the ones that were affected the most by it.

src/selectors/entities/files.ts
```diff
@@ -1,20 +1,19 @@
...
 import {GlobalState} from 'types/store';
 
-function getAllFiles(state: types.store.GlobalState) {
+import {sortFileInfos} from 'utils/file_utils';
+
+function getAllFiles(state: GlobalState) {
     return state.entities.files.files;
 }
 
-function getFilesIdsForPost(state: types.store.GlobalState, postId: string) {
```
jespino, line 17: this `types.whatever.Type` references here, aren't an accident, that was introduced to fix a problem to use it from the mattermost-webapp. If you remove them, maybe you break the mattermost-webapp build, or... maybe is already fixed and this is no longer needed.

jespino:
@hmhealey you didn't add the QA label, so I didn't add it either, but I don't not if this is ready to merge.

src/selectors/entities/files.ts
jespino: I made a quick try and is working locally.

hmhealey:
It shouldn't need QA since it's just changing whitespace and variable names pretty much. I avoided any functional changes.

PR #1118, opened 2020-04-06
[MM-5881] Add support for searching by position in user lists

mbecca:
#### Summary

Add support for searching by position in user lists

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-5881

### DIFF
--- a/src/utils/user_utils.ts
+++ b/src/utils/user_utils.ts
@@ -150,6 +150,7 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
         const full = first + ' ' + last;
         profileSuggestions.push(first, last, full);
         profileSuggestions.push((user.nickname || '').toLowerCase());
+        profileSuggestions.push((user.position || '').toLowerCase());
         const email = (user.email || '').toLowerCase();
         profileSuggestions.push(email);
         profileSuggestions.push((user.nickname || '').toLowerCase());


enahum:
This filter seems to be applied only at the client level.

Should this be added to mobile as well?

mbecca:
> 
> 
> This filter seems to be applied only at the client level, ticket seems to about applying this at the server level.
> 
> Should this be added to mobile as well?

I had not realized that it was from the server, also the changes are on the client side. At the moment I did not make the changes for mobile

src/utils/user_utils.ts
```diff
@@ -150,6 +150,7 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
         const full = first + ' ' + last;
         profileSuggestions.push(first, last, full);
         profileSuggestions.push((user.nickname || '').toLowerCase());
+        profileSuggestions.push((user.position || '').toLowerCase());
         const email = (user.email || '').toLowerCase();
         profileSuggestions.push(email);
         profileSuggestions.push((user.nickname || '').toLowerCase());
```
Willyfrog, line 156: not part of your PR, but I wonder why are we adding twice the nickname. @mgdelacroix any ideas?

enahum:
@wiersgallak are we ok with this working only on one client qnd not the other?

src/utils/user_utils.ts
mgdelacroix: Good catch! It's definitely not necessary, and as the function uses `some`, it does not alter the results, probably why it went unnoticed.

Addressed in https://github.com/mattermost/mattermost-redux/pull/1123

Willyfrog: thanks!

wiersgallak:
@enahum I created a ticket for the mobile support: https://mattermost.atlassian.net/browse/MM-24309

@mbecca please let us know if you are interested in adding support for that client as well.  

Thanks!

mbecca:
> 
> 
> @enahum I created a ticket for the mobile support: https://mattermost.atlassian.net/browse/MM-24309
> 
> @mbecca please let us know if you are interested in adding support for that client as well.
> 
> Thanks!

Yes, I can do it !! (in the near future)

Thank you all!

Regards, M

PR #1127, opened 2020-04-14
[MM-23159] Highlight Group Names

hahmadia:
**Unblocks:**
https://github.com/mattermost/mattermost-webapp/pull/5330

#### Summary

This change adds groups that are allowed to be referenced to the keys array of mention keys so that it is highlighted like we do for `@here` and `@channel` and so on. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23159

#### Screenshot
image

### DIFF
--- a/src/selectors/entities/groups.ts
+++ b/src/selectors/entities/groups.ts
@@ -113,7 +113,7 @@ const getChannelGroupIDSet = createSelector(
     (channelIDs) => new Set(channelIDs),
 );
 
-export const getGroupsNotAssociatedToTeam = createSelector(
+export const getGroupsNotAssociatedToTeam: (state: GlobalState, teamID: string) => Group[] = createSelector(
     getAllGroups,
     (state: GlobalState, teamID: string) => getTeamGroupIDSet(state, teamID),
     (allGroups, teamGroupIDSet) => {
@@ -121,7 +121,7 @@ export const getGroupsNotAssociatedToTeam = createSelector(
     },
 );
 
-export const getGroupsAssociatedToTeam = createSelector(
+export const getGroupsAssociatedToTeam: (state: GlobalState, teamID: string) => Group[] = createSelector(
     getAllGroups,
     (state: GlobalState, teamID: string) => getTeamGroupIDSet(state, teamID),
     (allGroups, teamGroupIDSet) => {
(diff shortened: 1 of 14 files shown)


src/selectors/entities/groups.ts
```diff
@@ -165,4 +166,13 @@ export const getAllAssociatedGroupsForReference = createSelector(
...
+);
+
+export const getCurrentUserGroupMentionKeys = createSelector(
+    getAllAssociatedGroupsForReference,
+    (groups: Array<Group>) => {
+        const keys: UserMentionKey[] = [];
+        groups.forEach((group) => keys.push({key: `@${group.name}`}));
+        return keys;
+    },
+);
```
fmunshi, line 178: Do we want to add a unit test for this selector specifically?

hahmadia: Good idea! Added

src/selectors/entities/groups.ts
```diff
@@ -168,11 +168,16 @@ export const getAllAssociatedGroupsForReference: (state: GlobalState) => Group[]
...
-    (groups: Array<Group>) => {
+    (state: GlobalState, groupMentionHighlightDisabled = false) => groupMentionHighlightDisabled,
+    (groups: Array<Group>, groupMentionHighlightDisabled) => {
         const keys: UserMentionKey[] = [];
-        groups.forEach((group) => keys.push({key: `@${group.name}`}));
+
+        if (!groupMentionHighlightDisabled) {
+            groups.forEach((group) => keys.push({key: `@${group.name}`}));
+        }
+
```
fmunshi: I dont think you want to be doing this - this will disable highlight only for the user without the group mentions permission not for other people.

hahmadia: I'm confused then. Maybe I didn't understand the requirement correctly? So as a guest user if I don't have permission to mention a group and I mention a group, shouldn't it not be highlighted for only me? Do you have suggestions what the right way to tackle this problem is?

hahmadia: It seems that it will be better for me to address this issue in another PR since we have the issue on Jira. So I reverted the changes I made and I'll be creating a new PR and we can discuss the solution there.

PR #1131, opened 2020-04-17
MM-23991 Add ability to sort DM channels in the new sidebar

hmhealey:
The sorting for DM channels may be moved onto the server eventually, but we can do it on the client for now since the client already sorts them alphabetically in the current version of the code.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23991


Willyfrog:
@hmhealey does this need qa?

hmhealey:
The QA review will have to be done when the feature is available in the web app since there's nothing to test for now

PR #1138, opened 2020-04-21
Fixing users cleanup on leave/remove guest from channel

jespino:
#### Summary
When you remove a guest from a channel the cleanup of the users was too
aggresive an was generating certain problems, no is handled at server side, so
is more reliable.

#### Ticket Link
[MM-24080](https://mattermost.atlassian.net/browse/MM-24080)

#### Related PRs
Server: mattermost/mattermost-server#14332


enahum:
mobile needed?

jespino:
@enahum yes, there is other changes that I have to submit, almost there

PR #1142, opened 2020-05-01
[MM-24376] Show channel (and team) as unread if system message is adding current user to channel

bradjcoughlin:
#### Summary
When a user is added to a channel on another team, they are not immediately shown an unread badge that they have been added. Until the web app reconnects with the server (it periodically syncs every 15 mins), the channel and team remains as read.

The web client receives the post via the websocket but currently ignores it as it ignores all system messages. This change makes an exception to allow the "system_add_to_team" post type through and increment the unread count.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24376

### DIFF
--- a/src/utils/post_utils.ts
+++ b/src/utils/post_utils.ts
@@ -25,6 +25,10 @@ export function isSystemMessage(post: Post): boolean {
     return Boolean(post.type && post.type.startsWith(Posts.SYSTEM_MESSAGE_PREFIX));
 }
 
+export function isSystemMessageAddToChannel(post: Post, userId: $ID<UserProfile>): boolean {
+    return Boolean(post.type && post.type === Posts.POST_TYPES.ADD_TO_CHANNEL && post.props.addedUserId === userId);
+}
+
 export function isMeMessage(post: Post): boolean {
     return Boolean(post.type && post.type === Posts.POST_TYPES.ME);
 }
--- a/src/utils/post_utils.ts
+++ b/src/utils/post_utils.ts
@@ -38,7 +38,9 @@ export function isPostEphemeral(post: Post): boolean {
 }
 
 export function shouldIgnorePost(post: Post, userId?: $ID<UserProfile>): boolean {
-    if ((post.type && post.type === Posts.POST_TYPES.ADD_TO_CHANNEL) && (post.props.addedUserId && post.props.addedUserId === userId)) {
+    const postTypeCheck = post.type && (post.type === Posts.POST_TYPES.ADD_TO_CHANNEL);
+    const userIdCheck = post.props && post.props.addedUserId && (post.props.addedUserId === userId);
+    if (postTypeCheck && userIdCheck) {
         return false;
     }
     return Posts.IGNORE_POST_TYPES.includes(post.type);
--- a/src/utils/post_utils.ts
+++ b/src/utils/post_utils.ts
@@ -25,10 +25,6 @@ export function isSystemMessage(post: Post): boolean {
     return Boolean(post.type && post.type.startsWith(Posts.SYSTEM_MESSAGE_PREFIX));
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/post_utils.ts
```diff
@@ -25,6 +25,10 @@ export function isSystemMessage(post: Post): boolean {
     return Boolean(post.type && post.type.startsWith(Posts.SYSTEM_MESSAGE_PREFIX));
 }
 
+export function isSystemMessageAddToChannel(post: Post, userId: $ID<UserProfile>): boolean {
```
hmhealey, line 28: I think this generally works, but what would you think about combining this with `shouldIgnorePost` instead of having two separate functions? If both of the functions are only ever used together, than keeping them together would likely make sense.

bradjcoughlin: I considered that, I just wasn't sure if it was worth adding complexity the `shouldIgnorePost` function if we may remove it from redux. `isSystemMessage` and `shouldIgnorePost` are already very similar and it feels like they could be merged.

hmhealey: They are slightly different in that not all types of system messages are ignored. `shouldIgnorePost` could get moved from redux to the web app, but I doubt we'll outright remove it. Overall, I added `shouldIgnorePost` to encapsulate all the conditions for ignoring a post (aka being a join/leave message), but I gave it the generic name so that its purpose would be very clear.

bradjcoughlin: That approach make sense - I updated the function to include this case.

src/utils/post_utils.ts
```diff
@@ -37,7 +37,10 @@ export function isPostEphemeral(post: Post): boolean {
     return post.type === Posts.POST_TYPES.EPHEMERAL || post.type === Posts.POST_TYPES.EPHEMERAL_ADD_TO_CHANNEL || post.state === Posts.POST_DELETED;
 }
 
-export function shouldIgnorePost(post: Post): boolean {
+export function shouldIgnorePost(post: Post, userId?: $ID<UserProfile>): boolean {
+    if ((post.type && post.type === Posts.POST_TYPES.ADD_TO_CHANNEL) && (post.props.addedUserId && post.props.addedUserId === userId)) {
```
sudheerDev, line 41: Can we add tests for the exception of `ADD_TO_CHANNEL` type here?

bradjcoughlin: Yep, I'll add that.

bradjcoughlin:
@hmhealey I think I've addressed your comments if you wouldn't mind taking another look. Thanks.

amyblais:
@hmhealey Assigned this to you if you're able to help cherry-pick this along with the webapp PR https://github.com/mattermost/mattermost-webapp/pull/5427.

amyblais:
@sudheerDev Would you be open to cherry-picking this along with the webapp PR https://github.com/mattermost/mattermost-webapp/pull/5427, or would it be easier to push this to v5.26?

sudheerDev:
> would it be easier to push this to v5.26

@amyblais that would require no work so it is definitely easy. 

Well, I think these PR's are fine and should not create any conflicts. I will cherrypick then later today and let you know when they are done

sudheerDev:
@amyblais Cherrypicked web PR, this PR is merged before 5.25v is created hence it is already in. We don't need to cherry-pick this.

amyblais:
@sudheerDev All v5.25 PRs need to be cherry-picked because v5.25 branch is based off from v5.24 branch.

sudheerDev:
@amyblais My bad, this was merged before 5.24v was created so we don't need to cherry-pick this. This commit exists on the release branch I cross checked couple of times

sudheerDev:
Adding label cherrypick done for consistency though it does not require cherrypick

PR #1143, opened 2020-05-06
Add metric warning support (announcement bar and DM)

catalintomai:
#### Summary
Add sysadmin warning support (announcement bar and DM) for multiple metrics (here - if the number of active users metric is above a certain limit).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25042

#### Related PRs
webapp: https://github.com/mattermost/mattermost-webapp/pull/5447
server: https://github.com/mattermost/mattermost-server/pull/14483


src/reducers/requests/admin.ts
```diff
@@ -408,6 +408,16 @@ function disablePlugin(state: RequestStatusType = initialRequestState(), action:
     );
 }
 
+function sendWarnMetricAck(state: RequestStatusType = initialRequestState(), action: GenericAction): RequestStatusType {
```
jespino, line 411: We are no longer tracking the request unless is strictly needed, for example use the request completion to trigger something in other side of the code. But if the code making the call is the one that is reacting to the changes we only need to use the async/await (or promise) result for showing that.

jespino: Removing this is making some changes here and there, not only this reducer.

src/actions/admin.test.js
```diff
@@ -1209,4 +1209,15 @@ describe('Actions.Admin', () => {
 
         const state = store.getState();
     });
+
+    it('sendWarnMetricAck', async () => {
+        const warnMetricAck = {
+            metric1: true,
+        };
+        nock(Client4.getBaseRoute()).
+            get('/warn_metrics/ack').
```
jespino, line 1218: Shouldn't it be mocking the `post` call instead of the `get` call?

src/actions/general.ts
```diff
@@ -185,6 +185,24 @@ export function getRedirectLocation(url: string): ActionFunc {
...
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        let data;
+        try {
+            data = await Client4.getWarnMetricsStatus();
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            return {error};
+        }
+
+        dispatch(batchActions([
```
jespino, line 198: this `batchActions` is not needed

src/actions/general.test.js
```diff
@@ -132,6 +132,24 @@ describe('Actions.General', () => {
...
+            metric2: false,
+        };
+
+        nock(Client4.getBaseRoute()).
+            get('/warn_metrics/status').
+            query(true).
+            reply(200, responseData);
+
+        await Actions.getWarnMetricsStatus()(store.dispatch, store.getState);
+        await TestHelper.wait(100);
```
jespino, line 147: I'm not sure this `wait` is needed here, for sure shouldn't be needed, but I'm not sure if we need more changes anywhere to make it work properly.

src/actions/admin.test.js
catalintomai: yes it should, will update

src/actions/general.test.js
catalintomai: I'll remove it, but right now it looks to be used in some places and not in others, so I erred on the caution side.

src/actions/general.ts
catalintomai: sure

src/reducers/requests/admin.ts
catalintomai: removed the request tracking defines+related.

srkgupta:
Hi @catalintomai 
I am beginning to test this now. Can you please resolve all the merge conflicts here. Thanks.

srkgupta:
This PR is awaiting some of the issues address in the dependent PR:
https://github.com/mattermost/mattermost-server/pull/14483

PR #1146, opened 2020-05-06
MM-24916 Add redux actions to create/rename custom categories

hmhealey:
This PR is based on https://github.com/mattermost/mattermost-redux/pull/1144. The changes specific to this PR can be found here: https://github.com/mattermost/mattermost-redux/compare/MM-20895_custom-categories...MM-24916_add-rename-categories

These actions still aren't hooked up to the server, but along with the PR that this is based on, it should allow full testing of custom categories with the proper back-of-webapp infrastructure.

Renaming a category is completely straightforward, but creating a category is more complicated because there are some funny things with how it positions new categories and adding channels to the new category requires removing them from old ones as well.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24916
https://mattermost.atlassian.net/browse/MM-20895

### DIFF
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -262,14 +262,13 @@ export function createCategory(teamId: string, displayName: string, channelIds: 
             for (const categoryId of categoryIds) {
                 const category = getCategory(state, categoryId);
 
-                if (!category.channel_ids.some((channelId) => channelIds.includes(channelId))) {
-                    continue;
+                // Only modify categories that have to be changed
+                if (category.channel_ids.some((channelId) => channelIds.includes(channelId))) {
+                    categoriesToUpdate.push({
+                        ...category,
+                        channel_ids: category.channel_ids.filter((channelId) => !channelIds.includes(channelId)),
+                    });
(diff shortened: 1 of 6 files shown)


src/actions/channel_categories.ts
```diff
@@ -245,3 +247,86 @@ function removeItem<T>(array: T[], item: T) {
...
+        }
+
+        const categoriesToUpdate = [newCategory];
+
+        // Remove the provided channels from any existing categories they may have existed in
+        if (channelIds.length > 0) {
+            for (const categoryId of categoryIds) {
+                const category = getCategory(state, categoryId);
+
+                if (!category.channel_ids.some((channelId) => channelIds.includes(channelId))) {
```
nevyangelova, line 289: 3/5 Can we revert this logic for readability, the 'continue' is not necessary since we can push only when the condition is true

hmhealey: Yeah, I can do that. At some point, we decided that early returns and early continues were preferable when writing branching logic, but having the condition inverted here does hurt that.

PR #1150, opened 2020-05-11
[MM-14906] Add a new selector to getProfilesNotInTeam

fmunshi:
#### Summary
- Add a new selector to `getProfilesNotInTeam` - to be used with the team members block

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-14906


src/selectors/entities/users.ts
```diff
@@ -306,6 +306,16 @@ export const getProfilesInTeam: (state: GlobalState, teamId: $ID<Team>) => Array
     },
 );
 
+export const getProfilesNotInTeam: (state: GlobalState, teamId: $ID<Team>) => Array<UserProfile> = createSelector(
+    getUsers,
+    getUserIdsNotInTeams,
+    (state: GlobalState, teamId: string) => teamId,
+    (state: GlobalState, teamId, filters) => filters,
```
hahmadia: I guess I'm a little confused where `filters` is coming from? I don't quite understand where we get filters from. Also, in the same way we are doing above, should we do the following? 
```suggestion
    (state: GlobalState, teamId: string, filters: {object or whatever suppose be}) => filters,
```

src/selectors/entities/users.ts
```diff
@@ -344,6 +353,17 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
+export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false): Array<UserProfile> {
+    const doGetProfilesInChannel = makeGetProfilesInChannel();
+    const profiles = filterProfilesMatchingTerm(doGetProfilesInChannel(state, channelId, false), term);
+
+    if (skipCurrent) {
+        removeCurrentUserFromList(profiles, getCurrentUserId(state));
```
hahmadia, line 361: Out of curiosity, we don't need to reassign profiles here? It will change profiles on the spot without need to re-assign? I just thought javascript was always pass by value and pass by reference wasn't a thing although I may be incorrect.

src/selectors/entities/users.ts
hahmadia: Thanks for quick fix!

src/selectors/entities/users.ts
```diff
@@ -474,6 +473,38 @@ export function makeGetProfilesForReactions(): (state: GlobalState, reactions: A
     );
 }
 
+export const getProfilesInChannel: (state: GlobalState, channelId: $ID<Channel>, skipInactive: boolean) => Array<UserProfile> = createSelector(
```
fmunshi: turns out I didnt need these two new selectors and I can get by with just using `makeGetProfiles...`

src/selectors/entities/users.ts
```diff
@@ -344,6 +353,17 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
+export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false): Array<UserProfile> {
+    const doGetProfilesInChannel = makeGetProfilesInChannel();
+    const profiles = filterProfilesMatchingTerm(doGetProfilesInChannel(state, channelId, false), term);
```
hahmadia, line 358:
```suggestion
    const profiles = filterProfilesMatchingTerm(doGetProfilesInChannel(state, channelId), term);
```
I think you can remove the `false` since the function by default is false for that parameter.
image

src/selectors/entities/users.ts
hahmadia: I noticed we use this elsewhere but I'm still confused how removeCurrentUserFromList effects `profiles` when Javascript is a pass by value language. Do you happen to know how it works cause I don't quite understand it although we are doing the same exact thing in other parts of the codebase.

fmunshi:
@hahmadia Arrays are passed by reference in JavaScript
https://www.dyn-web.com/javascript/arrays/value-vs-reference.php

hahmadia:
> https://www.dyn-web.com/javascript/arrays/value-vs-reference.php

Thanks Farhan for that! I had a read through the website and learned quite a bit. Appreciate it!

srkgupta:
Waiting for the dev review to be completed on the dependent PR:
https://github.com/mattermost/mattermost-webapp/pull/5469

PR #1152, opened 2020-05-11
[MM-24695] Only highlight groups you are apart of yellow

hahmadia:
#### Summary

This ticket addressed the issue of all groups getting highlighted for all users regardless. Now groups only get highlighted yellow ONLY if that user is apart of that group.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24695

### DIFF
--- a/src/action_types/groups.ts
+++ b/src/action_types/groups.ts
@@ -37,5 +37,7 @@ export default keyMirror({
     PATCHED_GROUP_TEAM: null,
     PATCHED_GROUP_CHANNEL: null,
 
+    RECEIVED_MY_GROUPS: null,
+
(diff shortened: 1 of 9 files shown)


devinbinnie:
@hahmadia Do we need QA on this? I don't think there's much we can review on this PR. We can have them look at the webapp one.

hahmadia:
> @hahmadia Do we need QA on this? I don't think there's much we can review on this PR. We can have them look at the webapp one.

Yah that's true but in our team we tend to tag our QA on all the PRs even though this is tested through webapp. It's mostly just for coordination but yah this would be tested through the webapp.

PR #1153, opened 2020-05-13
MM-24682: don't mutate on RECEIVED_CHANNELS

lieut-data:
#### Summary
Fix the reducer not to mutate the incoming action on `RECEIVED_CHANNELS`. While this isn't a problem in production, it frequently throws an exception during plugin development.

Fixes: https://mattermost.atlassian.net/browse/MM-24682

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24682


lieut-data:
@mickmister, yep! Repeatedly resetting the plugin (enable/disable) no longer crashes the webapp.

src/reducers/entities/channels.ts
```diff
@@ -65,9 +65,9 @@ function channels(state: IDMappedObjects<Channel> = {}, action: GenericAction) {
     case SchemeTypes.RECEIVED_SCHEME_CHANNELS: {
         const nextState = {...state};
 
-        for (const channel of action.data) {
-            if (state[channel.id] && channel.type === General.DM_CHANNEL) {
-                channel.display_name = channel.display_name || state[channel.id].display_name;
+        for (let channel of action.data) {
+            if (state[channel.id] && channel.type === General.DM_CHANNEL && !channel.display_name) {
+                channel = {...channel, display_name: state[channel.id].display_name};
```
hmhealey, line 70: Do you know when we're dispatching a RECEIVED_CHANNELS action with data that's already in the store? That's weird to me that we can't mutate `action.data` because it's something that came from the store previously.

lieut-data: I think Redux requires you not to modify any of the incoming parameters, including the action (https://redux.js.org/troubleshooting/#never-mutate-reducer-arguments):
> Never mutate reducer arguments
> It is tempting to modify the state or action passed to you by Redux. Don't do this!

hmhealey: Oh, yeah. That makes sense now that you mention it since the action will also be handled by other reducers as well.

lieut-data:
@DHaussermann -- this should finally fix all those weird plugin deploy issues (assuming they don't still register their own reducer in the wrong order).

lieut-data:
Removing cherry-pick status, since there's technically no urgency to get this into v5.24.

hanzei:
@lieut-data Just to double check: Can you confirm that no plugins fails on load due to this bug?

lieut-data:
> @lieut-data Just to double check: Can you confirm that no plugins fails on load due to this bug?

@hanzei, my understanding of the bug being fixed here is that it only affects plugins loading in development environments, since the deep-freeze library checks for the mutation and throws an exception as a result. A plugin not using a reducer probably wouldn't be affected. No plugin in production should be affected.

lieut-data:
@DHaussermann, any objection to merging this and just letting it soak in v5.26? I imagine it's going to be hard to meaningfully qualify this change given how low-level it occurs. The unit tests are in place to help automate that.

lieut-data:
Thanks, @DHaussermann! Merged, and will let the next webapp update pull this change in organically.

PR #1159, opened 2020-05-20
Fix new message indicator not showing for webhook messages to webhook owner

WGH-:
#### Summary
The new message indicator has a special case for messages posted by the
current user. Currently, webhook messages are attributed to its creator,
which breaks the new message indicator.

#### Ticket Link
 https://github.com/mattermost/mattermost-server/issues/14612

### DIFF
--- a/src/utils/post_list.ts
+++ b/src/utils/post_list.ts
@@ -8,7 +8,7 @@ import {getBool} from 'selectors/entities/preferences';
 import {isTimezoneEnabled} from 'selectors/entities/timezone';
 import {getCurrentUser} from 'selectors/entities/users';
 import {createIdsSelector, memoizeResult} from 'utils/helpers';
-import {isUserActivityPost, shouldFilterJoinLeavePost} from 'utils/post_utils';
+import {isUserActivityPost, shouldFilterJoinLeavePost, isFromWebhook} from 'utils/post_utils';
 import {getUserCurrentTimezone} from 'utils/timezone_utils';
 import * as types from 'types';
 export const COMBINED_USER_ACTIVITY = 'user-activity-';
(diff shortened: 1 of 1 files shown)


jasonblais:
@hmhealey @devinbinnie please help review when you have the chance?

@WGH- deep apologies for this long review time..

hmhealey:
@WGH- To get this tested by QA and in-use in the web app, we'll need a PR submitted against https://github.com/mattermost/mattermost-webapp to update the version of mattermost-redux in the package.json. Would you be available to submit that? If not, one of us can do it easily enough.

WGH-:
@hmhealey I'll be able to do that not earlier than Monday. I'll only have to update the commit hash in the mattermost-redux line, correct?

WGH-:
By the way, when your webhook mentions someone who is not in the channel, you'll see "were mentioned but they are not in the channel. Would you like to add them to the channel?" message (which I think was intended when you mention someone yourself, not your webhook). This's a similar problem caused by webhook messages being attributed to its owner. Maybe in the long run it would be better to break this attribution. Just saying.

hmhealey:
> @hmhealey I'll be able to do that not earlier than Monday. I'll only have to update the commit hash in the mattermost-redux line, correct?

Yeah, that's correct. For now, you can point it the latest commit in this branch, and then once this PR is tested and merged, we can point it to the merged commit.

> By the way, when your webhook mentions someone who is not in the channel, you'll see "were mentioned but they are not in the channel. Would you like to add them to the channel?" message (which I think was intended when you mention someone yourself, not your webhook). This's a similar problem caused by webhook messages being attributed to its owner. Maybe in the long run it would be better to break this attribution. Just saying.

Unfortunately, having webhooks owned by users is a key part of our Slack-compatibility that we try for, although it obviously causes a lot of problems like these 😛. I didn't know that it was causing the out-of-channel-mention message though. I don't think any of us noticed since most of our webhooks are owned by bots or by dedicated accounts. I'll file another ticket for that.

hmhealey:
@jgilliam17 This PR can be tested here: https://github.com/mattermost/mattermost-webapp/pull/6116

PR #1161, opened 2020-05-21
MM-25040: Omits some groups from channels where the team is group-con…

mkraft:
…strained.

### DIFF
--- a/src/selectors/entities/groups.test.js
+++ b/src/selectors/entities/groups.test.js
@@ -90,6 +90,9 @@ describe('Selectors.Groups', () => {
                 },
             },
             teams: {
+                teams: {
+                    [teamID]: {group_constrained: false},
+                },
                 groupsAssociatedToTeam: {
                     [teamID]: {ids: teamAssociatedGroupIDs},
                 },
@@ -134,7 +137,12 @@ describe('Selectors.Groups', () => {
 
     it('getGroupsNotAssociatedToChannel', () => {
         const expected = Object.entries(testState.entities.groups.groups).filter(([groupID]) => !channelAssociatedGroupIDs.includes(groupID)).map(([, group]) => group);
-        assert.deepEqual(Selectors.getGroupsNotAssociatedToChannel(testState, channelID), expected);
+        assert.deepEqual(Selectors.getGroupsNotAssociatedToChannel(testState, channelID, teamID), expected);
+
+        const cloneState = JSON.parse(JSON.stringify(testState));
+        cloneState.entities.teams.teams[teamID].group_constrained = true;
+        cloneState.entities.teams.groupsAssociatedToTeam[teamID].ids = [expectedAssociatedGroupID1];
+        assert.equal(Selectors.getGroupsNotAssociatedToChannel(cloneState, channelID, teamID).length, expected.length - 1);
     });
 
     it('getGroupsAssociatedToTeamForReference', () => {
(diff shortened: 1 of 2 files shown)


gsagula:
This PR is out of my domain of expertise. Let me know if you think that my review here is still valid or relevant.

mkraft:
> This PR is out of my domain of expertise. Let me know if you think that my review here is still valid or relevant.

I'll re-assign it.

mkraft:
> Code LGTM. I would just say that we write 2 more test that don't use only the length but also check the actual groups we get back. Ensure that the groups we are getting back are what we are expecting. Once for a team that is not group constrained and once for a team that is group constrained. Thoughts?

Yes I think you're right. I'll add them.

src/selectors/entities/groups.test.js
```diff
@@ -134,7 +137,12 @@ describe('Selectors.Groups', () => {
 
     it('getGroupsNotAssociatedToChannel', () => {
         const expected = Object.entries(testState.entities.groups.groups).filter(([groupID]) => !channelAssociatedGroupIDs.includes(groupID)).map(([, group]) => group);
-        assert.deepEqual(Selectors.getGroupsNotAssociatedToChannel(testState, channelID), expected);
+        assert.deepEqual(Selectors.getGroupsNotAssociatedToChannel(testState, channelID, teamID), expected);
+
+        const cloneState = JSON.parse(JSON.stringify(testState));
+        cloneState.entities.teams.teams[teamID].group_constrained = true;
+        cloneState.entities.teams.groupsAssociatedToTeam[teamID].ids = [expectedAssociatedGroupID1];
```
fmunshi, line 144: can we just use destructuring to clone the new state?

mkraft: > can we just use destructuring to clone the new state?

I tried that first but our `deepFreezeAndThrowOnMutation` function rejected it. Let me try again because it *should* work.

mkraft: Looking into it a bit more it looks like the custom `__defineSetter__` methods are cloned across when doing `const cloneState = {...testState}`, which explains why it doesn't work. To clone the object like that I think we would have to re-define the methods, something like this:

```js
cloneState.entities.teams.teams[teamID].__defineSetter__("group_constrained", someFunc);
cloneState.entities.teams.groupsAssociatedToTeam[teamID].__defineSetter__("ids", someFunc);
```

Doesn't seem worth it.

mkraft:
@hahmadia Good call on the `Object.values` instead of `Object.entries` 👍

src/selectors/entities/groups.test.js
fmunshi: I meant something like this @mkraft - up to you either way im 0/5 on it 
```
const modifiedState = {
            ...testState,
            entities: {
                ...testState.entities,
                teams: {
                    ...testState.entities.teams,
                    teams: {
                        ...testState.entities.teams.teams,
                        [teamID]: {
                            ...testState.entities.teams.teams[teamID],
                            group_constrained: true,
                        },
                    },
                    groupsAssociatedToTeam: {
                    ...testState.entities.teams.groupsAssociatedToTeam,
                        [teamID]: {
                            ...testState.entities.teams.groupsAssociatedToTeam[teamID],
                            ids: [expectedAssociatedGroupID1],
                        },
                    },
                },
            },
        };
```

srkgupta:
Waiting for the dependent PR to be Dev approved
https://github.com/mattermost/mattermost-server/pull/14619

amyblais:
@mkraft This one as well needs cherry picking

mkraft:
> @mkraft This one as well needs cherry picking

Done now. Thanks Amy.

PR #1162, opened 2020-05-21
Fully type Client4 methods

hmhealey:
I wanted to add proper return types for the new channel category API methods, and I made the mistake of doing that by parameterizing `Client4.doFetch` which caused every other method to start throwing type errors because the type parameter was unknown, so I gave up and just started typing everything in there.

This also touched a few other areas of the type definitions. Of note:
- I renamed the `Config` type to `ClientConfig` and added a placeholder `Config` type to match the server. I'm going to file a ticket to properly type `Config` since I was too lazy to do that all.
- I added some missing types from the redux state where needed since TypeScript was yelling at me.
- I removed the `async` off most of the store methods. It's not necessary when you return the results of another `async` call without `await`ing it. Either way, the return type for that method is the same `Promise` regardless of `async` being there.

### DIFF
--- a/src/client/client4.ts
+++ b/src/client/client4.ts
@@ -1,27 +1,96 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {buildQueryString, isMinimumServerVersion} from 'utils/helpers';
-import {cleanUrlForLogging} from 'utils/sentry';
 import {General} from '../constants';
 
+import {ClusterInfo, AnalyticsRow} from 'types/admin';
+import {Audit} from 'types/audits';
+import {UserAutocomplete, AutocompleteSuggestion} from 'types/autocomplete';
+import {Bot, BotPatch} from 'types/bots';
+import {
+    Channel,
+    ChannelMemberCountsByGroup,
+    ChannelMembership,
+    ChannelModeration,
+    ChannelModerationPatch,
+    ChannelStats,
+    ChannelsWithTotalCount,
+    ChannelUnread,
+    ChannelViewResponse,
+    ChannelWithTeamData,
+} from 'types/channels';
+import {Options, StatusOK, ClientResponse} from 'types/client4';
+import {Compliance} from 'types/compliance';
+import {
+    ClientConfig,
+    ClientLicense,
+    Config,
+    DataRetentionPolicy,
+    EnvironmentConfig,
+    License,
+} from 'types/config';
+import {CustomEmoji} from 'types/emojis';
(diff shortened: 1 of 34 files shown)


src/client/client4.ts
```diff
@@ -421,58 +491,58 @@ export default class Client4 {
...
             {method: 'post', body: JSON.stringify({token, new_password: newPassword})},
         );
     }
 
-    getKnownUsers = async () => {
+    getKnownUsers = () => {
         this.trackEvent('api', 'api_get_known_users');
 
-        return this.doFetch(
+        return this.doFetch<$ID<UserProfile>>(
```
devinbinnie, line 521: Should this not return an array? Or am I misunderstanding the `$ID` type?

src/client/client4.ts
```diff
@@ -667,58 +733,58 @@ export default class Client4 {
...
             {method: 'get'},
         );
     };
 
-    getProfilesInGroupChannels = async (channelsIds: string[]) => {
+    getProfilesInGroupChannels = (channelsIds: string[]) => {
         this.trackEvent('api', 'api_profiles_get_in_group_channels', {channelsIds});
 
-        return this.doFetch(
+        return this.doFetch<{[channelId: string]: UserProfile[]}>(
```
devinbinnie, line 745: You can shorten this to `Record<string, UserProfile[]>`, it's basically the TypeScript helper for a dictionary.

src/client/client4.ts
```diff
@@ -807,210 +876,210 @@ export default class Client4 {
...
             {method: 'post', body: JSON.stringify({current_service: 'email', new_service: 'ldap', email, password: emailPassword, ldap_id: ldapId, new_password: ldapPassword, mfa_code: mfaCode})},
         );
     };
 
-    switchLdapToEmail = async (ldapPassword: string, email: string, emailPassword: string, mfaCode = '') => {
+    switchLdapToEmail = (ldapPassword: string, email: string, emailPassword: string, mfaCode = '') => {
         this.trackEvent('api', 'api_users_ldap_to_email');
 
-        return this.doFetch(
+        return this.doFetch<{follow_link: string}>(
```
devinbinnie, line 936: Since we're redefining this type over a few times, should we just create a type for it?

src/client/client4.ts
```diff
@@ -2308,150 +2379,150 @@ export default class Client4 {
...
             {method: 'get'},
         );
     };
 
     // Jobs Routes
 
-    getJob = async (id: string) => {
-        return this.doFetch(
+    getJob = (id: string) => {
+        return this.doFetch<Job[]>(
```
devinbinnie, line 2391: This should just return `Job`

PR #1165, opened 2020-06-01
MM-24917 Update channel object last_post_at when new message is received

sudheerDev:
#### Summary
This PR updates the value of `last_post_at` in channel object. This will be used for determining if delay is needed when prefetching posts. If `last_post_at` is within the last minute then a jitter is added before prefetching posts so all clients don't start requesting at the same time.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24924

Related PR: https://github.com/mattermost/mattermost-webapp/pull/5547

### DIFF
--- a/src/reducers/entities/channels.ts
+++ b/src/reducers/entities/channels.ts
@@ -170,7 +170,7 @@ function channels(state: IDMappedObjects<Channel> = {}, action: GenericAction) {
             ...state,
             [channel_id]: {
                 ...channel,
-                last_post_at: create_at,
+                last_post_at: Math.max(create_at, channel.last_post_at),
             },
         };
     }
--- a/src/reducers/entities/channels.ts
+++ b/src/reducers/entities/channels.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 import {combineReducers} from 'redux';
-import {ChannelTypes, UserTypes, SchemeTypes, GroupTypes} from 'action_types';
+import {ChannelTypes, UserTypes, SchemeTypes, GroupTypes, PostTypes} from 'action_types';
 import {General} from '../../constants';
 import {GenericAction} from 'types/actions';
 import {Channel, ChannelMembership, ChannelStats, ChannelMemberCountByGroup, ChannelMemberCountsByGroup} from 'types/channels';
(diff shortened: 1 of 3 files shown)


src/reducers/entities/channels.ts
```diff
@@ -157,6 +157,24 @@ function channels(state: IDMappedObjects<Channel> = {}, action: GenericAction) {
...
+
+        if (!channel) {
+            return state;
+        }
+
+        return {
+            ...state,
+            [channel_id]: {
+                ...channel,
+                last_post_at: create_at,
```
hmhealey, line 173: Would there be any benefit to doing something like 
```suggestion
                last_post_at: Math.min(create_at, channel.last_post_at),
```
to ensure that this never gets out of sync if a new post event is ever received at a weird time?

sudheerDev: If you meant to suggest `Math.max`? yes, I think it makes it better in case of any network races.

sudheerDev:
> This is certainly easier than having to look at either the last_post_at or the create_at of the most recent post laoded, so I'm all for it, but I had a couple questions first:
> 
> 1. Updating the `last_post_at` doesn't change the position of the new messages line, does it?
> 2. See below

I checked this in places of where we use `last_post_at` and I couldn't see any issues that it could cause. Although I forgot, this changes one thing in UX.

A DM/GM is removed from LHS after `x` amount of days, This is done only on load of webapp but with these changes, it will remove the DM/GM dynamically in the run time of app.

@esethna FYI, ☝️ Let me know if you think this is odd.

sudheerDev:
FYI, @Willyfrog I added a type definition `ExperimentalDataPrefetch` needed for the related PR.

src/reducers/entities/channels.ts
hmhealey: I remember writing `Math.min` on purpose, but I can't remember why that would've been. `Math.max` definitely sounds correct though.

hmhealey:
> A DM/GM is removed from LHS after x amount of days, This is done only on load of webapp but with these changes, it will remove the DM/GM dynamically in the run time of app.

I think that actually happens dynamically. That code is incredibly complicated, but if I recall correctly, it checks for the `create_at` of the most recent post if that exists instead of the `last_post_at`

sudheerDev:
> I think that actually happens dynamically. That code is incredibly complicated, but if I recall correctly, it checks for the create_at of the most recent post if that exists instead of the last_post_at

yes, but how often do we visit a DM/GM channel before it is about to expire from LHS to have loaded posts?. Anyway now irrespective of posts loaded or not it works in same way. We can also remove the logic to be dependent on `create_at` after this change gets in

hmhealey:
Sorry, I meant to say that I don't think that's a big deal. I forgot to approve the PR last time I looked at it,

PR #1169, opened 2020-06-05
[MM-25771] Add filter for selecting profiles that are not deleted

fmunshi:
#### Summary
- Adds a filter to the get profiles redux actions to "skip inactive" users

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25771

Related Web PR: https://github.com/mattermost/mattermost-webapp/pull/5662

### DIFF
--- a/src/selectors/entities/users.ts
+++ b/src/selectors/entities/users.ts
@@ -357,9 +357,9 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
-export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false): Array<UserProfile> {
+export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false, skipInactive = false): Array<UserProfile> {
     const doGetProfilesInChannel = makeGetProfilesInChannel();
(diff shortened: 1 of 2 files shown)


src/selectors/entities/users.ts
```diff
@@ -354,9 +357,9 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
-export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false): Array<UserProfile> {
+export function searchProfilesInChannel(state: GlobalState, channelId: $ID<Channel>, term: string, skipCurrent = false, skipInactive = false): Array<UserProfile> {
     const doGetProfilesInChannel = makeGetProfilesInChannel();
-    const profiles = filterProfilesMatchingTerm(doGetProfilesInChannel(state, channelId, false), term);
+    const profiles = filterProfilesMatchingTerm(doGetProfilesInChannel(state, channelId, skipInactive), term);
```
hmhealey, line 362: Not something specific to this PR, but making a new `doGetProfilesInChannel` selector as well as filtering each time this is called gets rid of any memoization benefits from the selector, so it may cause performance issues due to constant re-rendering of components. Depending on where this is used, it could cause be worse, but it'll probably have less effect in the System Console.

PR #1177, opened 2020-06-12
MM-24914 Attach sidebar logic to server

hmhealey:
Here's the big PR to link the server logic for custom categories to the redux. I originally thought that adding server calls would allow us to remove almost all the client-side logic here, but it turns out that that doesn't work because the API responses from the server don't reflect all the side effects of the APIs (for example, deleting a custom category changes the channels in the Channels category even though the API response is only about the deleted category).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24914

### DIFF
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -3,7 +3,11 @@
 
 import {ChannelCategoryTypes} from 'action_types';
 
-import {fetchMyChannelsAndMembers, unfavoriteChannel, favoriteChannel} from 'actions/channels';
+import {Client4} from 'client';
+
+import {unfavoriteChannel, favoriteChannel} from 'actions/channels';
+import {logError} from 'actions/errors';
+import {forceLogoutIfNecessary} from 'actions/helpers';
 
 import {General} from '../constants';
 import {CategoryTypes} from 'constants/channel_categories';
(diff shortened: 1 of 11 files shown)


src/actions/channel_categories.ts
```diff
@@ -51,94 +50,66 @@ export function collapseCategory(categoryId: string) {
...
     };
 }
 
 // addChannelToInitialCategory returns an action that can be dispatched to add a newly-joined or newly-created channel
 // to its either the Channels or Direct Messages category based on the type of channel. New DM and GM channels are
 // added to the Direct Messages category on each team.
-export function addChannelToInitialCategory(channel: Channel): ActionFunc {
+//
+// Unless setOnServer is true, this only affects the categories on this client. If it is set to true, this updates
+// categories on the server too.
```
reflog, line 111: out of curiosity - what's the use case for setOnServer=false?

hmhealey: It's for when adding a channel to Channels is a side effect of a server-side action, like when a custom category is deleted and all of its channels get moved back to the Channels/DMs category. When we delete a category, the server just deletes it and then returns that the category was deleted successfully, but on the client, we need to move the now-orphaned channels in the category to where they belong.

PR #1178, opened 2020-06-16
MM-26066 Changes to use rudder npm package

sudheerDev:
#### Summary
Changes to use rudder npm package instead of a global object. Exporting rudder so we can use this dependency for webapp as well. Exporting was a suggestion from Rudder docs https://github.com/rudderlabs/rudder-sdk-js#step-1-install-rudderstack-using-the-code-snippet   

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26066.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "redux-thunk": "2.3.0",
     "remote-redux-devtools": "0.5.16",
     "reselect": "4.0.0",
+    "rudder-sdk-js": "^1.0.4",
     "serialize-error": "6.0.0",
     "shallow-equals": "1.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "redux-thunk": "2.3.0",
     "remote-redux-devtools": "0.5.16",
     "reselect": "4.0.0",
-    "rudder-sdk-js": "^1.0.4",
+    "rudder-sdk-js": "1.0.4",
     "serialize-error": "6.0.0",
     "shallow-equals": "1.0.0"
   },
--- a/src/client/client4.ts
+++ b/src/client/client4.ts
@@ -1,7 +1,5 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-
-import * as rudderAnalytics from 'rudder-sdk-js';
 import {General} from '../constants';
 
 import {ClusterInfo, AnalyticsRow} from 'types/admin';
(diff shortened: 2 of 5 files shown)


src/client/client4.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import * as rudderAnalytics from 'rudder-sdk-js';
```
lieut-data, line 4: Should we pull this into a `rudder.ts` file instead of as a "side export" of the client? (Then this file can just import that file so we maintain the "singleton" nature.)

sudheerDev: I thought about it but for these two lines, it seemed odd creating a file.

```
import * as rudderAnalytics from 'rudder-sdk-js';
export {rudderAnalytics};
```

Do you see more than this in a file?

sudheerDev: Hmm, Let me know if you see more than this but I think it makes to create a file as the import on web look good. 

`import rudderAnalytics from 'client/rudder';`

lieut-data: Nope, I think that's exactly what I expected in the file. The benefit might be to add a comment explaining that this is effectively a singleton and not to import this resource yourself.

0/5 on the import path -- both your last example, and the current approach of importing `{rudderAnalytics}` from `client` as exported by the index seem fine.

sudheerDev: Done

src/client/rudder.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// the module exports a bunch of api's on a already defined object combined with node module caching,
+// selectively exporting which can be used throughout the project
```
lieut-data, line 5: nit:
```suggestion
// As per rudder-sdk-js documentation, import this only once and use like a singleton.
// See https://github.com/rudderlabs/rudder-sdk-js#step-1-install-rudderstack-using-the-code-snippet
```

package.json
```diff
@@ -38,6 +38,7 @@
     "redux-thunk": "2.3.0",
     "remote-redux-devtools": "0.5.16",
     "reselect": "4.0.0",
+    "rudder-sdk-js": "^1.0.4",
```
grundleborg, line 41: Should we pin an exact version here?

sudheerDev: Nice catch, changed it to be exact

sudheerDev:
Will be merging this one Webapp PR is QA approved

sudheerDev:
tested on webapp PR

PR #1179, opened 2020-06-25
Update NOTICE.txt

amyblais:
To match with what is here: https://github.com/mattermost/mattermost-redux/blob/master/package.json

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -41,63 +41,6 @@ THE SOFTWARE.
 
 ---
 
-## deep-equal
-
-This product contains 'deep-equal' by James Halliday.
-
-node's assert.deepEqual algorithm
-
-* HOMEPAGE:
-  * https://github.com/substack/node-deep-equal#readme
-
-* LICENSE: MIT
-
-This software is released under the MIT license:
-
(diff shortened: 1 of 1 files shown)


NOTICE.txt
```diff
@@ -491,6 +318,19 @@ SOFTWARE.
...
+## redux-persist-node-storage
+
+This product contains 'redux-persist-node-storage' by Pelle Jacobs.
+
+Storage adaptor to use node-localstorage with redux-persist
+
+* HOMEPAGE:
+  * https://github.com/pellejacobs/redux-persist-node-storage
+
+* LICENSE:
```
amyblais, line 330: No license

amyblais:
@grundleborg Kind reminder to review (+merge and cherry-pick if approved)

NOTICE.txt
grundleborg: This is the one I meant to comment on previously. NPM says MIT for the license, but given there's no license shown in the Github repo I'm not sure what we can assume.

amyblais: I've requested that they add a license file: https://github.com/pellejacobs/redux-persist-node-storage/issues/7.

PR #1182, opened 2020-07-02
[MM-26363] - Check if channel belong to category before dispatching addChannelToInitialCategory

nevyangelova:
#### Summary
Sometimes channels get added to two categories when we create a custom category because addChannelToInitialCategory is always ran. This fixes the issue by checking if the channel has a category assigned already. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26323

### DIFF
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -29,7 +29,7 @@ import {
     DispatchFunc,
     GetStateFunc,
 } from 'types/actions';
-import {CategorySorting, OrderedChannelCategories} from 'types/channel_categories';
+import {CategorySorting, OrderedChannelCategories, ChannelCategory} from 'types/channel_categories';
 import {Channel} from 'types/channels';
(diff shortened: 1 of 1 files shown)


src/actions/channel_categories.ts
```diff
@@ -128,7 +128,10 @@ export function addChannelToInitialCategory(channel: Channel, setOnServer = fals
         }
 
         // Add the new channel to the Channels category on the channel's team
-        const channelsCategory = getCategoryInTeamByType(state, channel.team_id, CategoryTypes.CHANNELS);
+        let channelsCategory: ChannelCategory | undefined;
+        if (!categories.some((category) => category.channel_ids.some((channelId) => channelId === channel.id))) {
```
hmhealey, line 132: Instead of only sometimes setting `channelsCategory`, could you instead return early if this condition is true? That'll help make it clear that we're not doing anything with the channel because it's already in a category.

PR #1183, opened 2020-07-02
Mm 23832 new system roles

mkraft:


src/utils/user_utils.ts
```diff
@@ -41,41 +41,51 @@ export function displayUsername(
     return name;
 }
 
-export function rolesIncludePermission(roles: string, permission: string): boolean {
-    const rolesArray = roles.split(' ');
-    return rolesArray.includes(permission);
+export function spaceSeparatedStringIncludes(spaceSeparated: string, item: string): boolean {
```
mkraft, line 44: Unrelated: renamed this func.

src/constants/permissions.ts
```diff
@@ -72,6 +73,37 @@ export default {
...
+    PERMISSION_READ_SYSCONSOLE_USERMANAGEMENT_USERS: 'read_sysconsole_user_management_users',
+    PERMISSION_WRITE_SYSCONSOLE_USERMANAGEMENT_USERS: 'write_sysconsole_user_management_users',
+    PERMISSION_READ_SYSCONSOLE_USERMANAGEMENT_GROUPS: 'read_sysconsole_user_management_groups',
+    PERMISSION_WRITE_SYSCONSOLE_USERMANAGEMENT_GROUPS: 'write_sysconsole_user_management_groups',
+    PERMISSION_READ_SYSCONSOLE_USERMANAGEMENT_TEAMS: 'read_sysconsole_user_management_teams',
+    PERMISSION_WRITE_SYSCONSOLE_USERMANAGEMENT_TEAMS: 'write_sysconsole_user_management_teams',
+    PERMISSION_READ_SYSCONSOLE_USERMANAGEMENT_CHANNELS: 'read_sysconsole_user_management_channels',
+    PERMISSION_WRITE_SYSCONSOLE_USERMANAGEMENT_CHANNELS: 'write_sysconsole_user_management_channels',
+    PERMISSION_READ_SYSCONSOLE_USERMANAGEMENT_PERMISSIONS: 'read_sysconsole_user_management_permissions',
+    PERMISSION_WRITE_SYSCONSOLE_USERMANAGEMENT_PERMISSIONS: 'write_sysconsole_user_management_permissions',
```
fmunshi, line 91: Does it make sense to prefix these with `PERMISSION_` when none of the other permissions are prefixed with `PERMISSION_`?

src/selectors/entities/roles_helpers.ts
```diff
@@ -14,6 +14,10 @@ export type PermissionsOptions = {
     permission: string;
 };
 
+export type SysConsoleItemOptions = {
+    resourceId: string;
+};
+
```
fmunshi, line 20: As far as i can tell this isnt used in redux at all does it make more sense to just keep it on the webapp side?

src/utils/user_utils.ts
fmunshi: I think you can actually just replace this entire function with `string.includes(item)` no need for the split and no need for this function in general

fmunshi: I guess this is more accurate if you have a permission that is a substring of another permission for example `sysconsole_usermanagement` and `read_sysconsole_usermanagement` ... you could get a false positive, ignore my above comment

src/selectors/entities/roles_helpers.ts
mkraft: Good point, and it's not used in webapp either. I've removed it. Thanks.

src/constants/permissions.ts
mkraft: Good point, it's not consistent. I've update them. Thanks.

PR #1185, opened 2020-07-03
Sidebar V2 Phase 2

hmhealey:
These are the mattermost-redux changes for the new sidebar. These changes are primarily to support the new channel categories information that comes from the server instead of deriving which channels are visible based on channel membership like we previously used to do.

Note that there are outstanding issues with this logic, but we've decided that they're small enough that they can be addressed once this is on master. We want to get in the core of this on Monday so that it'll have a week to soak on community-daily before the actual submission deadline.

Since this is coming from a feature branch, a thorough review is less necessary since the individual PRs going into this have already been reviewed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23730

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/5854
https://github.com/mattermost/mattermost-server/pull/14967


src/actions/channel_categories.ts
```diff
@@ -16,3 +48,343 @@ export function collapseCategory(categoryId: string) {
...
+        // Add the channel to the new category
+        const categories = [{
+            ...category,
+            sorting: (setManualSorting && category.type !== CategoryTypes.DIRECT_MESSAGES) ? CategorySorting.Manual : category.sorting,
+            channel_ids: insertWithoutDuplicates(category.channel_ids, channelId, newIndex),
+        }];
+
+        // And remove it from the old category
+        const originalCategory = getCategoryInTeamWithChannel(getState(), category.team_id, channelId);
+        if (originalCategory && originalCategory.id !== category.id) {
```
Willyfrog, line 182: why would we move to the same category? is it to change it's index?

src/actions/channels.ts
```diff
@@ -1394,11 +1418,21 @@ export function favoriteChannel(channelId: string): ActionFunc {
 
         Client4.trackEvent('action', 'action_channels_favorite');
 
-        return savePreferences(currentUserId, [preference])(dispatch);
+        // Add the channel to the Favorites category, unless this has been called from the categories update logic
```
Willyfrog, line 1421: why the different treatment? wouldn't this be a moveChannelToCategory from a default one?

src/actions/channel_categories.ts
```diff
@@ -16,3 +48,343 @@ export function collapseCategory(categoryId: string) {
...
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+
+        return dispatch({
+            type: ChannelCategoryTypes.RECEIVED_CATEGORY,
+            data: updatedCategory,
+        });
```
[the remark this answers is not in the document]
hmhealey: Nope, this is correct. A selector is responsible for sorting the category when it's not manually sorted, so only the sorting method should be changing here, not the order of the channels in the category.

Also, `RECEIVED_CATEGORY_ORDER` refers to the order of the categories in the sidebar, not the channels in the category.

src/actions/channel_categories.ts
hmhealey: Yep, that's correct

sudheerDev: Rearranging?

src/actions/channels.ts
hmhealey: Ideally it would be, but I never got around to changing all the favouriting logic to use the new method.

I did have a ticket for that though: https://mattermost.atlassian.net/browse/MM-23989

PR #1187, opened 2020-07-06
[MM-26322] - Optimistically update the order of channels/categories when dragging

nevyangelova:
#### Summary
Optimistically update the order of channels and categories in order to avoid 'jumping' when dragging.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26322


nevyangelova:
> A couple questions:
> 
> 1. Could dispatching the RECEIVED_CATEGORIES twice mess up the sidebar state, like if the user attempts to make another change in between the two actions? Maybe we could get away with only dispatching it once before the request is made.
> 2. What happens if the request fails? We don't really have a good way of rolling back the sidebar state since these changes made by `moveChannelToCategory` could get complicated. Perhaps we could cheat on the rollback by just reloading the affected categories if an error occurs.

1. Good idea, we can just use the request to persist the information on the backend and check that it didn't fail. 
2. I guess I should dispatch the action with the old categories when the request fails.

src/actions/channel_categories.ts
```diff
@@ -170,6 +170,10 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
         const category = getCategory(state, categoryId);
         const currentUserId = getCurrentUserId(state);
 
+        const originalCategories = [{
```
hmhealey, line 173: This name isn't great since it's easy to confuse with `originalCategory` below. I can't think of a better name for this, so maybe `originalCategory` can be renamed to `sourceCategory`?

src/actions/channel_categories.ts
```diff
@@ -186,12 +190,21 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
...
         try {
-            data = await Client4.updateChannelCategories(currentUserId, category.team_id, categories);
+            await Client4.updateChannelCategories(currentUserId, category.team_id, categories);
         } catch (error) {
             forceLogoutIfNecessary(error, dispatch, getState);
             dispatch(logError(error));
+
+            dispatch({
+                type: ChannelCategoryTypes.RECEIVED_CATEGORIES,
+                data: originalCategories,
```
hmhealey, line 206: Relating to above, this should also include `originalCategory` since that may have been changed as well

nevyangelova: Not sure I understand, but here we need to Null-ify the optimistic update, so we are trying to make it so as if nothing happened. We just have the previous state and that should be enough?

hmhealey: The problem is that `sourceCategories` only includes the original version of `category` and not `originalCategory`, so we're only reverting `category` if something fails

nevyangelova: That was the idea, we optimistically dispatch the changed ```category``` and then if that fails, we roll back to the original state. I think I might be missing something here, given what you are saying?

nevyangelova: I think I understood, we need to roll back both categories involved in the change, original and source category. Is that correct?

nevyangelova: Maybe we should rename them 'source' and 'target' since it was a bit confusing?

hmhealey: Yeah, I'd be up for renaming them like that. I think the renaming would go
- category -> targetCategory
- originalCategory -> sourceCategory
- categories stays the same
- sourceCategories -> originalCategories

src/actions/channel_categories.ts
```diff
@@ -183,6 +179,10 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
 
         // And remove it from the old category
         const originalCategory = getCategoryInTeamWithChannel(getState(), category.team_id, channelId);
+        const sourceCategories = [{
+            ...category,
+        }, originalCategory];
```
hmhealey, line 184: This should be moved into the if statement below in case `originalCategory` is somehow null or if it's the same as `category`

src/actions/channel_categories.ts
```diff
@@ -183,6 +179,10 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
 
         // And remove it from the old category
         const originalCategory = getCategoryInTeamWithChannel(getState(), category.team_id, channelId);
+        const sourceCategories = [{
+            ...category,
```
hmhealey, line 183: Also, this can just be `category`. It doesn't need to be copied since we're not editing it

src/actions/channel_categories.ts
nevyangelova: I don't think this would work because we need `sourceCategories` for the dispatch below.

hmhealey:
Manually cherry-picked this onto the release-5.26

PR #1189, opened 2020-07-08
POC Plugin Integrations

larkox:
#### Summary
This is a POC to allow plugins to work on mobile and webapp in a different way.

The main idea is to fetch the list of integrations to the server, and show them according to the location. The only location implemented so far in webapp is the post menu. At the moment we are only passing the text to show (we could also add an icon in future implementations). Apart of that we receive a requestURL.

The requestURL is just what to do when an action is performed (for channel header, a tap on the menu element). This will call the url. Here the response will have what action to do among a subset of actions. No actions are defined for webapp yet, but we may have actions like "GO_TO" to go to a different part of the app, or "OPEN_URL" to open an URL.

The idea is to add little by little more locations and extend what can be done with this kind of integrations. Eventually, I would like to populate complete views thanks to this (i.e. RHS).

Disclaimer 1: This approach was thought first for mobile, so the naming may still refer to mobile. When we have the naming more consolidated, we will update all the related PRs.

This is partly based on Maria's idea of Plugin Actions (https://docs.google.com/document/d/138uxYyUdjFFecB-tD3JWV7dTVxBBQjIo-cX7pGblI3E/edit). Here the main differences:

- My take tries to be more flexible, in order to add more locations as we go. For example, not all locations may need an icon, or a text.
- Type and Location would be the same concept.
- I did not consider permissions for my implementation, but it is something to keep in mind.
- Result actions are present in both ideas
- When fetching from the server, we fetch all plugins locations. That way we have everything cached for when it needs to be used.
- The content of what we fetch will be dependent on what is the location

#### Ticket Link
Webapp: https://github.com/mattermost/mattermost-webapp/pull/5887
Mobile: https://github.com/mattermost/mattermost-mobile/pull/4480
Server: https://github.com/mattermost/mattermost-server/pull/14906
Zoom: https://github.com/mattermost/mattermost-plugin-zoom/pull/139


larkox:
Apart of naming (discussion that will probably go through all the different PRs), and any change related to the other PRs (mainly server) this PR is from my side ready to merge.

larkox:
Closing since this effort is now driven by the "Apps" effort.

PR #1194, opened 2020-07-15
[MM-24300] Upgrade dependencies 5.28

calebroseland:
#### Summary

- ##### Patch upgrades
  Major upgrades excluded as fixes to breaking changes not worth it as `mattermost-redux` will be migrated to 
   `mattermost-webapp` in the near- to mid-term.

- ##### Fix failing CI builds like:
  - https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/20769/workflows/f6bac22a-ab8b-4719-843f-3f504a76526e/jobs/130867/parallel-runs/0/steps/0-102
- Quick return type correction

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24300


tsconfig.json
```diff
@@ -35,6 +35,7 @@
         "noUnusedParameters": false,
         "noImplicitReturns": true,
         "noFallthroughCasesInSwitch": true,
+        "skipLibCheck": true,
```
hmhealey, line 38: What does this change?

calebroseland: Probably a better explanation than I could come up with: 
https://stackoverflow.com/questions/52311779/usage-of-the-typescript-compiler-argument-skiplibcheck

> You could find two libraries define the same types, but incompatibly. 

It essentially skips checking ts declaration files, including in imported libs, which may often declare conflicting types or mis-matched versions of types. 

`skipLibCheck` is currently enabled in `mattermost-webapp`.

hmhealey: Makes sense. I didn't realize it was true in the web app as well. It seems like one of those things that probably shouldn't be enabled in ideal circumstances, but we're pretty much never writing code in ideal circumstances

PR #1197, opened 2020-07-17
MM-26410/MM-26825 Only use new sidebar API for favourites when enabled

hmhealey:
The web app is currently in a weird state where some parts use the sidebar to show favourites and other parts use the user preferences. This makes it so that we almost exclusively use the sidebar when it's enabled to mitigate against cases where the current channel is favourited in the sidebar, but not in the centre channel. This typically only applies to DMs/GMs because the favourites preferences are cross-team, but the favourites in the sidebar is team-specific.

There's still one place that uses the preferences even if the sidebar is enabled, and that's in the `isAutoClosed` function in `channel_utils`, but the code around it is a mess, and I don't think the code that calls it is even used with the new sidebar enabled.

I also took the opportunity to purge some mobile-specific code that relied on the favourites since it made it easier to switch this around.

This also fixes MM-26825 which was happening because there's a bug on the server, and it would constantly be triggered because we were telling the server to both favourite channels in the sidebar and prefs even though the server was already doing both itself.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26410
https://mattermost.atlassian.net/browse/MM-26825

#### Related Pull Requests


hmhealey:
This has to be merged at the same time as the webapp PR since it moves the `isFavoriteChannel` selector

PR #1198, opened 2020-07-17
[MM-26958] function to sort channels by supplied channel type list

ashishbhate:
#### Summary

- Function to sort channels by supplied channel type list. 
- I didn't want to update `sortChannelsByTypeAndDisplayName` since that would repercussions in many places, so I made a new function.  

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-26958
- Parent https://mattermost.atlassian.net/browse/MM-15477

### DIFF
--- a/src/utils/channel_utils.test.js
+++ b/src/utils/channel_utils.test.js
@@ -311,7 +311,7 @@ describe('ChannelUtils', () => {
         let sortfn = sortChannelsByTypeListAndDisplayName.bind(null, 'en', [General.OPEN_CHANNEL, General.PRIVATE_CHANNEL, General.DM_CHANNEL, General.GM_CHANNEL]);
         let actual = [channelOpen1, channelPrivate, channelDM, channelGM, channelOpen2].sort(sortfn);
         let expected = [channelOpen1, channelOpen2, channelPrivate, channelDM, channelGM];
-        expect(actual).toEqual(expected)
+        expect(actual).toEqual(expected);
 
         // Skipped Open Channel type should sort last but open channels should still sort in alphabetical order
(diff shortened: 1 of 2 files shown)


src/utils/channel_utils.test.js
```diff
@@ -260,4 +261,62 @@ describe('ChannelUtils', () => {
...
+            name: 'channelE',
+            team_id: 'teamId',
+            display_name: 'Unit Test channelE',
+            type: General.GM_CHANNEL,
+            delete_at: 0,
+            total_msg_count: 0,
+        };
+
+        let sortfn = sortChannelsByTypeListAndDisplayName.bind(null, 'en', [General.OPEN_CHANNEL, General.PRIVATE_CHANNEL, General.DM_CHANNEL, General.GM_CHANNEL]);
+        let actual = JSON.stringify([channelOpen1, channelPrivate, channelDM, channelGM, channelOpen2].sort(sortfn));
```
enahum, line 312: Why stringify? I think you can compare arrays in jest without convert them to string. 0/5

ashishbhate: jest complains that `Compared values have no visual difference.` 

According to https://github.com/facebook/jest/issues/5998 some object literals may not work correct with `equal`.

enahum: According to the link that happens with annonymous funtions, no?

hmhealey: Instead of `assert.equal`, what if you use `expect(actual).toEqual(expected)`? I don't remember if `assert.equal` expects both objects to be the same object or if it checks that they're deeply equal. With `expect`, there's `toEqual` for deep equality or `toBe` for ensuring that they're the same object

ashishbhate: `expect(actual).toEqual(expected)`  works. So does `assert.deepEqual`. I was incorrectly using `assert.Equal`.

PR #1201, opened 2020-07-20
Add Selectors to get all the recent channel Ids

sridhar02:
#### Summary
To get all the recent channel Ids and this will be used to display recent channels in channel switcher. 

#### Ticket Link
This PR is related to an issue in the web app.
https://github.com/mattermost/mattermost-server/issues/14633

### DIFF
--- a/src/selectors/entities/channels.ts
+++ b/src/selectors/entities/channels.ts
@@ -765,6 +765,80 @@ export const getSortedUnreadChannelIds: (state: GlobalState, lastUnreadChannel: 
     (unreadChannelIds, mappedAndSortedUnreadChannelIds) => mappedAndSortedUnreadChannelIds,
 );
 
+//recent channels
+
+export const getRecentChannelIds: (state: GlobalState) => Array<string> = createIdsSelector(
+    getAllChannels,
+    getMyChannelMemberships,
+    getChannelIdsForCurrentTeam,
+    (channels: IDMappedObjects<Channel>, members: RelationOneToOne<Channel, ChannelMembership>, teamChannelIds: Array<string>): Array<string> => {
+        const recentIds = teamChannelIds.filter((id) => {
+            const c = channels[id];
+            const m = members[id];
+
+            if (c && m) {
+                return true;
+            }
+
+            return false;
+        });
+
+        return recentIds;
+    },
+);
+
+export const getRecentChannels: (state: GlobalState) => Array<Channel> = createIdsSelector(
+    getCurrentUser,
+    getUsers,
+    getUserIdsInChannels,
+    getAllChannels,
+    getRecentChannelIds,
+    getTeammateNameDisplaySetting,
+    (currentUser, profiles, userIdsInChannels: any, channels, recentIds, settings) => {
+        // If we receive an unread for a channel and then a mention the channel
+        // won't be sorted correctly until we receive a message in another channel
+        if (!currentUser) {
+            return [];
+        }
+
+        const allRecentChannels = recentIds.filter((id) => channels[id] && channels[id].delete_at === 0).map((id) => {
+            const c = channels[id];
+
+            if (c.type === General.DM_CHANNEL || c.type === General.GM_CHANNEL) {
+                return completeDirectChannelDisplayName(currentUser.id, profiles, userIdsInChannels[id], settings!, c);
+            }
+
+            return c;
+        });
+        return allRecentChannels;
+    },
+);
+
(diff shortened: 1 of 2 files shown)


src/selectors/entities/channels.ts
```diff
@@ -795,6 +795,80 @@ export const getSortedUnreadChannelIds: (state: GlobalState, lastUnreadChannel:
...
+    const locale = currentUser.locale || General.DEFAULT_LOCALE;
+
+    const recentChannelIds = channels.
+        sort(sortChannelsByRecencyOrAlpha.bind(null, locale, lastPosts, sorting)).
+        map((channel) => channel.id);
+
+    return recentChannelIds;
+};
+
+export const getAllRecentChannelIds: (state: GlobalState, sorting: SortingType) => Array<string> = createIdsSelector(
```
hmhealey, line 862: It might be simpler to avoid some of the intermediate functions here if that makes your life easier. Part of the reason that we have multiple methods for `getUnreadChannels`, `getUnreadChannelIds`, `getUnreadChannelsSet`, etc for other types of channels is a bit of a legacy issue.

I'm not sure which version of the selector would be needed for the channel switcher modal in the web app. Do you have a webapp PR that you're working on that'll use these changes?

src/selectors/entities/channels.ts
```diff
@@ -795,6 +795,80 @@ export const getSortedUnreadChannelIds: (state: GlobalState, lastUnreadChannel:
...
+        });
+        return allRecentChannels;
+    },
+);
+
+export const mapAndSortRecentChannelIds = (
+    channels: Array<Channel>,
+    currentUser: UserProfile,
+    lastPosts: RelationOneToOne<Channel, Post>,
+    sorting: SortingType,
```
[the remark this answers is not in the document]
sridhar02: yes it can be hardcoded and I have updated them in latest changes

src/selectors/entities/channels.ts
```diff
@@ -795,6 +795,80 @@ export const getSortedUnreadChannelIds: (state: GlobalState, lastUnreadChannel:
     (unreadChannelIds, mappedAndSortedUnreadChannelIds) => mappedAndSortedUnreadChannelIds,
 );
 
+//recent channels
+
+export const getRecentChannelIds: (state: GlobalState) => Array<string> = createIdsSelector(
```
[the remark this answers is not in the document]
sridhar02: I have seen the channel.ts file there are no selectors to get the recent channels.

src/selectors/entities/channels.ts
sridhar02: Yes I have combined all the different functions into a single selector  and I need this selector to get data the recent channels to display in the switcher for this issue https://github.com/mattermost/mattermost-server/issues/14633, I have to raise a PR for the web app also but once this redux selector is ready I will raise a PR for the web app.

image

src/selectors/entities/channels.ts
```diff
@@ -765,6 +765,56 @@ export const getSortedUnreadChannelIds: (state: GlobalState, lastUnreadChannel:
...
+        const sorting = 'recent';
+        const recentIds = teamChannelIds.filter((id) => {
+            const c = channels[id];
+            const m = members[id];
+
+            if (c && m) {
+                return true;
+            }
+
+            return false;
```
hmhealey, line 794: Also, this can be simplified slightly to

```suggestion
            return Boolean(c && m);
```

hmhealey:
Also, if you'd like to start working on the webapp PR to accompany this, you can point the webapp branch to your modified version of mattermost-redux to work against. I think you can do that by running `npm install sridhar02/mattermost-redux#e4fe2e26103cd240062b062ed9b1d44df9738f1d`. You just need to make sure to switch it back to the master branch of mattermost-redux after this PR is merged.

hmhealey:
Thanks for changing that and adding some tests. It looks like those tests are failing though because part of the redux state is missing from testState [here](https://github.com/mattermost/mattermost-redux/pull/1201/commits/cd8c030e225e15956f3ae78b0e5a5287bfb3a7d2#diff-a1f43b00c397e0c96a24f874c53f86feR2665). It's an unfortunate side effect of complicated selectors like most of the ones in this file. You should be able to fix them by adding something like
```
users: {
    currentUserId: TestHelper.generateId(),
    profiles: {},
},
```
into `testState.entities`

sudheerDev:
i am going to merge this without Q review as it can be tested on webapp and the code here does not impact any existing code

PR #1205, opened 2020-07-23
Update NOTICE.txt

amyblais:


amyblais:
@grundleborg Kind reminder to review for v5.26,

PR #1206, opened 2020-07-23
MM-26023 Move handling of sidebar API responses to the websocket

hmhealey:
When making requests to the server, the client that made those requests also receives the corresponding websocket events which often leads to us duplicating logic for handling the websocket event and the HTTP response. I'm going to bring this up at the dev meeting next week, but it might make sense to just handle everything in response to the websocket event to avoid duplicating effort and making the code much more complicated and difficult to follow.

Unfortunately, this does make it harder to test these, but thankfully we have the new E2E tests for that 😄

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26023

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/5924

### DIFF
--- a/src/actions/channel_categories.test.js
+++ b/src/actions/channel_categories.test.js
@@ -21,7 +21,6 @@ import {CategorySorting} from 'types/channel_categories';
 import {isFavoriteChannelOld} from 'utils/channel_utils';
 
 import * as Actions from './channel_categories';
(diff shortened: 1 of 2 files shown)


amyblais:
@sudheerDev Kind reminder to review this for v5.26,

hmhealey:
Manually cherry-picked onto release-5.26

PR #1215, opened 2020-08-14
Remove WebSocket actions

hmhealey:
The WebSocket actions in the mattermost-redux repo were only used by the mobile app previously, so since the mobile app no longer uses mattermost-redux, they can be removed to stop people from being confused about where changes to websocket code need to be made.

I also removed the mattermost-redux version of the handling of new posts since it was also unused in the web app, and I need to make some changes to it to fix a race condition (https://mattermost.atlassian.net/browse/MM-25452)


hmhealey:
Skipping QA review since it's just dead code removal

PR #1217, opened 2020-08-17
[MM-27535] Adding new configuration for user limits

nickmisasi:
#### Summary
Expose new configuration to flatmapped clientconfig

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27535

### DIFF
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -99,6 +99,7 @@ export type ClientConfig = {
     ExperimentalChannelSidebarOrganization: string;
     ExperimentalClientSideCertCheck: string;
     ExperimentalClientSideCertEnable: string;
+    ExperimentalCloudUserLimit: string;
     ExperimentalDataPrefetch: string;
     ExperimentalEnableAuthenticationTransfer: string;
     ExperimentalEnableAutomaticReplies: string;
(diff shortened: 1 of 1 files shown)


src/types/config.ts
```diff
@@ -654,6 +655,7 @@ export type ExperimentalSettings = {
     ClientSideCertEnable: boolean;
     ClientSideCertCheck: string;
     EnableClickToReply: boolean;
+    CloudUserLimit: string;
```
devinbinnie, line 658: Is this supposed to be here twice? Looking at the server PR it looks like you have it declared twice in `GenerateClientConfig` as well.

nickmisasi: I can't say for sure, but I was following a pattern with a config that already existed. The ExperimentalSetting EnableClickToReply exists in 2 places, the client config flat map:
https://github.com/mattermost/mattermost-redux/blob/c60ae33b8d53b29c142d5cdfbc1ff13ac92d63da/src/types/config.ts#L106
And in the ExperimentalSettings object below it: 
https://github.com/mattermost/mattermost-redux/blob/c60ae33b8d53b29c142d5cdfbc1ff13ac92d63da/src/types/config.ts#L657

The client config `ExperimentalCloudUserLimit` config is the one I plan to use, I'm 0/5 on keeping the other one.

devinbinnie: It's possible it was pulled out of experimental at some point but no one removed it from the `ClientConfig` type. We're gonna need to make sure that process is followed by server changes going forward.

But yeah, just use the `ExperimentalCloudUserLimit` for now.

nickmisasi: Sounds good, removed the CloudUserLimit config in favour of ExperimentalCloudUserLimit

PR #1220, opened 2020-08-19
Admin Advisor: get metric status on refresh for admins only

catalintomai:
#### Summary
When the current user is non-admin, we should not be calling the Admin Advisor metric warning functionality on page refresh.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27960

#### Related PRs
https://github.com/mattermost/mattermost-webapp/pull/6212


srkgupta (approved):
Tested the PR along with its dependent PR and now the API call to the endpoint '/api/v4/warn_metrics/status' is made by webapp only if the logged in user is a sysadmin. Approving the PR.

src/selectors/entities/users.ts
```diff
@@ -133,7 +133,7 @@ export function getUserByEmail(state: GlobalState, email: $Email<UserProfile>):
 export const isCurrentUserSystemAdmin: (state: GlobalState) => boolean = createSelector(
     getCurrentUser,
     (user) => {
-        const roles = user.roles || '';
+        const roles = (user && user.roles) || '';
```
mkraft, line 136:
```suggestion
        const roles = user?.roles || '';
```

hahmadia: 👍

PR #1239, opened 2020-09-18
Change user search logic from startsWith to indexOf

haalcala:
Change user search logic from startsWith to indexOf

### DIFF
--- a/src/utils/user_utils.test.js
+++ b/src/utils/user_utils.test.js
@@ -131,6 +131,10 @@ describe('user utils', () => {
             assert.deepEqual(filterProfilesMatchingTerm(users, 'some'), [userB]);
         });
 
+        it('should match by nickname by substring', () => {
+            assert.deepEqual(filterProfilesMatchingTerm(users, 'body'), [userB]);
+        });
+
         it('should match by email prefix', () => {
             assert.deepEqual(filterProfilesMatchingTerm(users, 'left'), [userB]);
         });
--- a/src/utils/user_utils.test.js
+++ b/src/utils/user_utils.test.js
@@ -131,10 +131,6 @@ describe('user utils', () => {
             assert.deepEqual(filterProfilesMatchingTerm(users, 'some'), [userB]);
         });
 
-        it('should not match by nickname substring', () => {
-            assert.deepEqual(filterProfilesMatchingTerm(users, 'body'), []);
-        });
-
         it('should match by email prefix', () => {
             assert.deepEqual(filterProfilesMatchingTerm(users, 'left'), [userB]);
         });
--- a/src/utils/user_utils.ts
+++ b/src/utils/user_utils.ts
@@ -161,7 +161,8 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
 
         return profileSuggestions.
             filter((suggestion) => suggestion !== '').
-            some((suggestion) => suggestion.startsWith(trimmedTerm));
+            // some((suggestion) => suggestion.startsWith(trimmedTerm));
+            some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
     });
 }
 
--- a/src/utils/user_utils.ts
+++ b/src/utils/user_utils.ts
@@ -160,9 +160,7 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
         }
 
         return profileSuggestions.
-            filter((suggestion) => suggestion !== '').
-            // some((suggestion) => suggestion.startsWith(trimmedTerm));
-            some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
+            filter((suggestion) => suggestion !== '')      .some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
     });
 }
 
--- a/src/utils/user_utils.ts
+++ b/src/utils/user_utils.ts
@@ -159,9 +159,7 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
             profileSuggestions.push(split[1]);
         }
 
-    return profileSuggestions
-      .filter((suggestion) => suggestion !== "")
-      .some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
+        return profileSuggestions.filter((suggestion) => suggestion !== '').some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
     });
 }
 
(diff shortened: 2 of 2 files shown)


hanzei:
@haalcala Thanks for your PR. Is there a ticket that this PR targets?

haalcala:
> @haalcala Thanks for your PR. Is there a ticket that this PR targets?

Hi,

Uhmm ... no tickets. Am I suppose to create a ticket first? Sorry, I'm new to contributing to open-source projects.

Anyways, basically, we modified the server to do user wildcard search (front and back).  Current code only does "<term>%" which only hits words that starts with the search term.  It is difficult to know the first few letters of a person's name in a large user list.  That said, the client does it work filtering for some reason even when the server is already doing some filtering and returning some records.  So, this is to allow partial user search fix.  Hope you'll consider.

hanzei:
No, it's not required to open an issue first. 

@hahmadia @mgdelacroix Could you check if the use case makes sense to support and if so open a JIRA for QA?

src/utils/user_utils.ts
```diff
@@ -160,8 +160,7 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
         }
 
         return profileSuggestions.
-            filter((suggestion) => suggestion !== '').
-            some((suggestion) => suggestion.startsWith(trimmedTerm));
+            filter((suggestion) => suggestion !== '')      .some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
```
hahmadia, line 163: Can we move this to new line?
```suggestion
            filter((suggestion) => suggestion !== '').
            some((suggestion) => suggestion.indexOf(trimmedTerm) >= 0);
```

hahmadia (changes requested):
I don't see a problem with doing this. I did some research and it seems the performance of startsWith and indexOf are about the same and indexOf provides the plus of wildcard searching which I believe is an improvement. I don't see a problem with this.

src/utils/user_utils.ts
haalcala: Hi,

Thanks. Shouldn't be problem.  Thanks for looking into it.

hahmadia (approved):
Sorry for the not pickiness but can we change the double quotes back to single quotes and keep the dots on the lines prior 
```
Like.
This

Not
.like
.this
```
Thanks! Sorry about that but otherwise code looks good. I'm gonna approve as it's not a breaking change but would prefer that the code styling matches how we are doing it else where.
Thanks again

haalcala:
Hi, thank you all for this! I made a few more changes to comply with the test requirements (including the linting).  I hope it'll be carried when you merge/review.

haalcala:
Btw, this PR is pointing to release-5.24.  Do I need to do anything to make it carried to 5.28?

src/utils/user_utils.test.js
```diff
@@ -131,10 +131,6 @@ describe('user utils', () => {
             assert.deepEqual(filterProfilesMatchingTerm(users, 'some'), [userB]);
         });
 
-        it('should not match by nickname substring', () => {
```
hahmadia, line 134: Hey! Sorry again for another comment but I feel we shouldn't purge this test altogether. Instead, since now we are doing a wildcard search, I would say rename the test title to `should match by nickname substring` since the functionality has changed.

hahmadia (approved):
Everything looks good to me now!

src/utils/user_utils.test.js
hahmadia: I'm just not a big fan of completely purging a test unless absolutely required. I personally always look to add or modify test to match new behaviour. Thanks!

haalcala: Hi, you're correct. Sorry about that, I thought the test before the one I removed would be sufficient. I just did another commit. Sorry for all the trouble.

mgdelacroix:
Hi @haalcala! First of all, thanks for your contribution to Mattermost!

The feature you're proposing in this PR and in https://github.com/mattermost/mattermost-server/pull/15569 has been discussed by our team and I'm afraid that it is more complex that it can look at a first glance. There are some technical challenges involved; although you've solved the frontend part and the database part in the backend, we would need to implement these same changes in both the Elasticsearch and Bleve search engines, and those changes are not easy to add as they involve changes in how indexing is done, so the final implementation would require from a good chunk of effort from the team.

The feature itself has some UX challenges as well. When we analized the idea of allowing partial matches on user autocompletion, we realised that for servers that have large amounts of users the search became almost impossible to use (for reference, think about trying to match "mar" for "martin" in a server with 60.000 users). The conclusion we got from this analysis is that the most feasible path forward would be to implement contextual aware search, and that's a whole project on its own.

These are the reasons why we decided not to add partial matching to Mattermost yet and spend proper time designing a solution that contemplates all these changes. I'm sorry that the current implementation is causing problems with your user list, we hope that future changes will help to alleviate that.

Again, thanks a lot for taking the time to contribute.

haalcala:
> Hi @haalcala! First of all, thanks for your contribution to Mattermost!
> 
> The feature you're proposing in this PR and in [mattermost/mattermost-server#15569](https://github.com/mattermost/mattermost-server/pull/15569) has been discussed by our team and I'm afraid that it is more complex that it can look at a first glance. There are some technical challenges involved; although you've solved the frontend part and the database part in the backend, we would need to implement these same changes in both the Elasticsearch and Bleve search engines, and those changes are not easy to add as they involve changes in how indexing is done, so the final implementation would require from a good chunk of effort from the team.
> 
> The feature itself has some UX challenges as well. When we analized the idea of allowing partial matches on user autocompletion, we realised that for servers that have large amounts of users the search became almost impossible to use (for reference, think about trying to match "mar" for "martin" in a server with 60.000 users). The conclusion we got from this analysis is that the most feasible path forward would be to implement contextual aware search, and that's a whole project on its own.
> 
> These are the reasons why we decided not to add partial matching to Mattermost yet and spend proper time designing a solution that contemplates all these changes. I'm sorry that the current implementation is causing problems with your user list, we hope that future changes will help to alleviate that.
> 
> Again, thanks a lot for taking the time to contribute.

Hi, uhm, this is very sad conclusion to be honest, but I do understand your point.  I'm just thinking usability should more prioritised over technical difficulties.  The need for this fix arises for multi-cultural setup whereby non-Japanese may not remember certain person with Japanese pronunciation.  This should apply to other languages as well.

Anyways, it shouldn't be a problem for me.  I'll just keep this change together with the modified copy that I have (together with the clustering(multi-instance) bridge that I managed to figure out how it should work.).  Keeping up with the upstream is a bit of a hassle, so I thought this is one less thing to worry about.

Again, thank you for all your time looking at my PR. I'd be glad to submit more in the future.

PR #1247, opened 2020-09-27
Add post_id for FileInfo type

jespino:
#### Summary
Add `post_id` for FileInfo type


jespino:
Skipping QA because is clearly not needed (I only changing the a type, so if it compiles, there's nothing else to check).

devinbinnie (approved):

calebroseland (approved):

PR #1250, opened 2020-09-29
[MM-28211] Redux state for holding cloud products and subscription

devinbinnie:
#### Summary
Redux state for holding cloud products and subscription.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28211

### DIFF
--- a/src/reducers/entities/cloud.ts
+++ b/src/reducers/entities/cloud.ts
@@ -1,0 +1,24 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {combineReducers} from 'redux';
+import {CloudTypes} from 'action_types';
+
+import {GenericAction} from 'types/actions';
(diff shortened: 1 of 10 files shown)


nickmisasi (approved):
Pipelines seem to be failing, but not isolated to your PR. LGTM

src/reducers/entities/cloud.ts
```diff
@@ -5,7 +5,18 @@ import {combineReducers} from 'redux';
 import {CloudTypes} from 'action_types';
 
 import {GenericAction} from 'types/actions';
-import {CloudCustomer} from 'types/cloud';
+import {Product, Subscription, CloudCustomer} from 'types/cloud';
+import {Dictionary} from 'types/utilities';
+
+function subscription(state: Subscription | null = null, action: GenericAction) {
```
hmhealey, line 11: It's annoying that we need to use `| null` here because of redux. It would be nice to be able to just use `undefined` everywhere.

hmhealey (approved):

PR #1251, opened 2020-09-29
[MM-27142] Change dollar_per_seat to price_per_seat

nickmisasi:
#### Summary
This was changed in mattermost-server but the change was not made here.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28197


marianunez (changes requested):
I believe now that the server has changed that it may not be receiving the price properly from CWS given that it's still defined there as `dollars_per_seat`.  We may need to set server and webapp back if we don't want to change CWS as well.

marianunez (approved):

devinbinnie (approved):

PR #1252, opened 2020-10-07
Define OpenGraphMetadata

KuSh:
#### Summary

Define OpenGraphMetadata

This is needed to convert 'components/post_view/post_attachment_opengraph' module to TypeScript (issue https://github.com/mattermost/mattermost-server/issues/15447)

Regards,
Nicolas

### DIFF
--- a/src/types/posts.ts
+++ b/src/types/posts.ts
@@ -115,7 +115,7 @@ export type PostsState = {
     postsInChannel: Dictionary<Array<PostOrderBlock>>;
     postsInThread: RelationOneToMany<Post, Post>;
     reactions: RelationOneToOne<Post, Dictionary<Reaction>>;
-    openGraph: RelationOneToOne<Post, OpenGraphMetadata>;
+    openGraph: RelationOneToOne<Post, Dictionary<OpenGraphMetadata>>;
     pendingPostIds: Array<string>;
     selectedPostId: string;
     currentFocusedPostId: string;
@@ -131,6 +131,7 @@ export declare type OpenGraphMetadataImage = {
 }
 
 export declare type OpenGraphMetadata = {
+    type?: string;
     title?: string;
     description?: string;
     site_name?: string;
--- a/src/types/posts.ts
+++ b/src/types/posts.ts
@@ -123,4 +123,17 @@ export type PostsState = {
     expandedURLs: Dictionary<string>;
 };
 
-export type OpenGraphMetadata = any;
+export declare type OpenGraphMetadataImage = {
+    secure_url?: string;
+    url: string;
+    height?: number;
+    width?: number;
+}
+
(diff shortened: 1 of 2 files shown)


src/types/posts.ts
```diff
@@ -115,12 +115,26 @@ export type PostsState = {
     postsInChannel: Dictionary<Array<PostOrderBlock>>;
     postsInThread: RelationOneToMany<Post, Post>;
     reactions: RelationOneToOne<Post, Dictionary<Reaction>>;
-    openGraph: RelationOneToOne<Post, OpenGraphMetadata>;
+    openGraph: RelationOneToOne<Post, Dictionary<OpenGraphMetadata>>;
```
nevyangelova, line 118: Why does this need to be a Dictionary type here?

KuSh: Hi @nevyangelova 

This is needed because it is how openGraph is accessed, see https://github.com/mattermost/mattermost-redux/blob/f203bb743267cc1bf958ce3cea5a34fb19dd9bfe/src/selectors/entities/posts.ts#L78

nevyangelova (approved):
Thanks!

src/types/posts.ts
hmhealey: Good catch. The redux state was set up to potentially support multiple OpenGraph previews in the future

hmhealey (approved):
Thanks for adding a proper type for this! There's potentially more fields in that object, but we can add those if we actually start to use them later on. I don't have a full list on hand since it comes from the Go library we use on the server to generate that object.

KuSh:
> Thanks for adding a proper type for this! There's potentially more fields in that object, but we can add those if we actually start to use them later on. I don't have a full list on hand since it comes from the Go library we use on the server to generate that object.

Haaaaaaaaaaaaaa that explains many things ! I was searching where this object was defined and used every fields I could find that was used either in mattermost-redux or mattermost-webapp .. I didn't thought to check the server :facepalm: 
Do you want me to create a PR with missing fields ?

What was merge is sufficient to unblock https://github.com/mattermost/mattermost-server/issues/15447 , I'll just need to wait for a release :)

Thanks again,
Regards,
Nicolas

hmhealey:
I think it's totally fine to leave as-is. We have no plans on using the other fields anyway. Also, the server doesn't even have its own definition of the object since it comes indirectly from a third party library ([link](https://github.com/dyatlov/go-opengraph/blob/master/opengraph/opengraph.go#L68) for reference).

As for the PR, you can run `npm install mattermost/mattermost-redux#fcdb54580b8ee88b17b40fe99609505f66f1418d` to update to this version of mattermost-redux. The web app doesn't use proper releases of mattermost-redux from npm since we update it frequently.

PR #1257, opened 2020-10-14
MM-28833 Remove computed details from getChannel selector

hmhealey:
As I mentioned earlier this week, `getChannel` is problematic because its a selector that takes an argument without being behind a factory function, so if you have two or more components both using it, they'll constantly be overwriting each others' results, leading to those components re-rendering constantly. The main reason that this causes problems is that `getChannel` returns a new object each time for DMs and GMs because it adds on some additional derived fields such as the dynamic display_name for DM and GM channels. On the current version of master, this leads to things like parts of the new sidebar and all of the RHS re-rendering constantly.

I changed `getChannel` to bea plain function that doesn't compute any data so that it can be re-used easily wherever we don't need the computed fields, and I've used `makeGetChannel` to generate one-off selectors for everywhere that does need that data.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28833

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/6794

### DIFF
--- a/src/selectors/entities/channels.test.js
+++ b/src/selectors/entities/channels.test.js
@@ -355,7 +355,7 @@ describe('Selectors.Channels.getOtherChannels', () => {
     });
 });
 
-describe('Selectors.Channels.getChannel', () => {
+describe('getChannel', () => {
     const team1 = TestHelper.fakeTeamWithId();
     const team2 = TestHelper.fakeTeamWithId();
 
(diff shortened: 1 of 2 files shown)


nevyangelova (approved):

sudheerDev (approved):

sudheerDev:
@hmhealey Does this require a testing label on redux?

hmhealey:
It'll be tested on the webapp branch, but I want to make sure it's tested there before it's merged. Merging this on its own might break anyone using the master version of mattermost-redux without the webapp changes

jgilliam17 (approved):
Tested on the related webapp PR.

PR #1266, opened 2020-10-19
[MM-28214][MM-28216] Redux hooks for getting invoices

devinbinnie:
#### Summary
This PR contains the redux state and methods for getting invoices from the Customer Web Server for the current MM Cloud subscription.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28216
https://mattermost.atlassian.net/browse/MM-28214


src/selectors/entities/cloud.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {createSelector} from 'reselect';
+
+import {Invoice} from 'types/cloud';
+import {GlobalState} from 'types/store';
+import {Dictionary} from 'types/utilities';
+
+export function makeGetLastInvoice(): (state: GlobalState) => Invoice | undefined {
```
marianunez, line 10: I believe that part of the changes was to have the `Subscription` struct have a field for `LastInvoice` the same way it is modeled in Stripe so we wouldn't need to calculate this. @ethervoid could you confirm if this is something you are adding?

ethervoid: Yes, I'm adding that as Ali suggested in the Subscription model as you can see [here](https://github.com/mattermost/customer-web-server/pull/257/files#diff-23410cc4a3f0578104eca2027b742790534897f8c5e79856678663e295f2736fR24)

devinbinnie: Cool, I'll remove this selector then and add that field in the `Subscription` front-end model

src/client/client4.ts
```diff
@@ -3307,51 +3311,62 @@ export default class Client4 {
...
         return this.doFetch<Subscription>(
-            `${this.getBaseRoute()}/cloud/subscription`,
+            `${this.getCloudRoute()}/subscription`,
+            {method: 'get'},
+        );
+    }
+
+    getInvoices = () => {
+        return this.doFetch<Invoice[]>(
+            `${this.getCloudRoute()}/subscription/invoices`,
```
ethervoid, line 3361: We have to pass the subscription id like `/subscription/${subscriptionId}/invoices` . My main concern on this is that we have to pass the stripe subscription id, eg. `sub_IEtgAkaTisBQTG`, because we don't store the subscription information

// @marianunez

src/client/client4.ts
```diff
@@ -3307,51 +3311,62 @@ export default class Client4 {
...
             {method: 'get'},
         );
     }
 
+    getInvoicePdfUrl = (invoiceId: string) => {
+        return `${this.getCloudRoute()}/subscription/invoices/${invoiceId}/pdf`;
```
ethervoid, line 3367: Same here

ethervoid (commented):
Some comments to discuss about the subscription id

src/client/client4.ts
devinbinnie: We shouldn't need to do this actually, if we mimic what I've done for getting the subscription (ie. using the installation ID) then we can avoid this.

marianunez: Correct, the subscription id will be passed in the enterprise by looking at the env var for installation id. It will be transparent from the front-end side.

marianunez (approved):

src/client/client4.ts
ethervoid: Understood :)

ethervoid (approved):
LGTM!

PR #1270, opened 2020-10-22
MM-22845 Fix quick switcher results

sudheerDev:
#### Summary
 * Change existing searchprofiles selector to be searchProfilesStartingWithTerm
 * Add a new selector searchProfileMatchingWithTerm for substring matches

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22854

### DIFF
--- a/src/selectors/entities/users.ts
+++ b/src/selectors/entities/users.ts
@@ -372,9 +372,7 @@ export function getFilteredUsersStats(state: GlobalState): any {
     return state.entities.users.filteredStats;
 }
 
-export function searchProfilesStartingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
-    const users = getUsers(state);
-    const profiles = filterProfilesStartingWithTerm(Object.keys(users).map((key) => users[key]), term);
+function filterFromProfiles(state: GlobalState, profiles: Array<UserProfile>, skipCurrent = false, filters?: Filters): Array<UserProfile> {
     const filteredProfilesMap = filterProfiles(profileListToMap(profiles), filters);
     const filteredProfiles = Object.keys(filteredProfilesMap).map((key) => filteredProfilesMap[key]);
 
@@ -385,17 +383,16 @@ export function searchProfilesStartingWithTerm(state: GlobalState, term: string,
     return filteredProfiles;
 }
 
+export function searchProfilesStartingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
+    const users = getUsers(state);
+    const profiles = filterProfilesStartingWithTerm(Object.keys(users).map((key) => users[key]), term);
+    return filterFromProfiles(state, profiles, skipCurrent, filters);
+}
+
 export function searchProfilesMatchingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
     const users = getUsers(state);
     const profiles = filterProfilesMatchingWithTerm(Object.keys(users).map((key) => users[key]), term);
-    const filteredProfilesMap = filterProfiles(profileListToMap(profiles), filters);
-    const filteredProfiles = Object.keys(filteredProfilesMap).map((key) => filteredProfilesMap[key]);
-
-    if (skipCurrent) {
-        removeCurrentUserFromList(filteredProfiles, getCurrentUserId(state));
-    }
-
-    return filteredProfiles;
+    return filterFromProfiles(state, profiles, skipCurrent, filters);
 }
 
 export function makeSearchProfilesInChannel() {
--- a/src/selectors/entities/users.ts
+++ b/src/selectors/entities/users.ts
@@ -17,7 +17,8 @@ import {getDirectShowPreferences, getTeammateNameDisplaySetting} from 'selectors
 
 import {
     displayUsername,
-    filterProfilesMatchingTerm,
+    filterProfilesStartingWithTerm,
+    filterProfilesMatchingWithTerm,
     isSystemAdmin,
     includesAnAdminRole,
     profileListToMap,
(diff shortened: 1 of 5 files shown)


src/utils/user_utils.ts
```diff
@@ -154,29 +174,34 @@ export function filterProfilesMatchingTerm(users: Array<UserProfile>, term: stri
...
+    return users.filter((user: UserProfile) => {
+        if (!user) {
+            return false;
+        }
+
+        const profileSuggestions = nameSuggestionsForUser(user);
+
+        return profileSuggestions.
+            filter((suggestion) => suggestion !== '').
+            some((suggestion) => suggestion.includes(trimmedTerm));
```
nevyangelova, line 201: this can be in the `some`, `some((suggestion) => suggestion !== '' && suggestion.includes(trimmedTerm));`

nevyangelova (approved):
LGTM, minus two non-blocking comments

src/selectors/entities/users.ts
```diff
@@ -371,9 +372,22 @@ export function getFilteredUsersStats(state: GlobalState): any {
     return state.entities.users.filteredStats;
 }
 
-export function searchProfiles(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
+export function searchProfilesStartingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
```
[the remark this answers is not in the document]
sudheerDev: I could make it an if else with keeping the default as term search. That way changes are minimal. I was hoping to remove the startWithTerm entirely at the end of the epic but I probably won't be able to do it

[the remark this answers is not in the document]
sudheerDev: Sorry i misunderstood your question i guess, the filter is optional and what kinda of composition do you have in mind?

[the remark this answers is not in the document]
nevyangelova: Basically I was thinking that the structure of the selectors `searchProfilesStartingWithTerm` and `searchProfilesMatchingWithTerm` is the same so we could have them as one function but it's not a necessary change.

src/utils/user_utils.ts
sudheerDev: Sure

src/selectors/entities/users.ts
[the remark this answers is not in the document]
sudheerDev: i abstracted a bit more code as it is similar.

> Basically I was thinking that the structure of the selectors `searchProfilesStartingWithTerm` and `searchProfilesMatchingWithTerm` is the same so we could have them as one function but it's not a necessary change.

Based. on the suggestions above what i understood is you want one function instead of these two. For that, we have two options both similar HOF or normal functions they pretty much will be the same as we don't have much composition going on here for other functions to leverage.

HOF pattern

```
searchProfiles(state, term, skipCurrent, filters) => {
  return (newFilterFunc) => {
            const users = getUsers(state);
           const profiles = newFilterFunc(Object.keys(users).map((key) => users[key]), term);
           .....
  }
}
```

and the usage will be in the way

`searchProfiles(state, term, skipCurrent, filters)(filterProfilesMatchingWithTerm)` or 
`searchProfiles(state, term, skipCurrent, filters)(filterProfilesStartingWithTerm)`

I am 1/5 to leave the way it is as it is more readable IMO at the place of usage albeit the extra lines of code. Looks weird with multiple filters at the place of usage.

Let me know if you have strong opinions otherwise, I am gonna let it be the way it is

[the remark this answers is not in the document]
nevyangelova: Yes that's what I had in mind but I am 1/5 on it as I already approved, I was just thinking out loud what could 'potentially' be improved but it looks good regardless!

src/selectors/entities/users.ts
```diff
@@ -384,10 +383,22 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
+export function searchProfilesStartingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
+    const users = getUsers(state);
+    const profiles = filterProfilesStartingWithTerm(Object.keys(users).map((key) => users[key]), term);
```
hmhealey, line 388: Could `Object.keys(users).map((key) => users[key])` be replaced with `Object.values(users)` instead?

src/selectors/entities/users.ts
```diff
@@ -384,10 +383,22 @@ export function searchProfiles(state: GlobalState, term: string, skipCurrent = f
     return filteredProfiles;
 }
 
+export function searchProfilesStartingWithTerm(state: GlobalState, term: string, skipCurrent = false, filters?: Filters): Array<UserProfile> {
```
hmhealey, line 386: Could you turn these into selector factories like `makeSearchProfiles...`? We should be doing that with all selectors that have parameters to avoid cases like the https://github.com/mattermost/mattermost-redux/pull/1257 where they cause a lot of re-renders.

hmhealey (changes requested):

src/selectors/entities/users.ts
sudheerDev: yes, we need memoization, Can we consider these changes out of the scope of the PR? these are not. new changes and I like to have them separately filed

src/selectors/entities/users.ts
sudheerDev: Sure

src/selectors/entities/users.ts
sudheerDev: Done, but i don't think they usually add much value here. Most of these sectors are used in channel provider and user provider. They. invalidate with new search terms.  Nevertheless changed it to be a factory

hmhealey: Yeah, I agree that it's kind of unnecessary, but I want to make a habit of using the factories since it will help prevent mistakes like with `getChannel`. Alternatively, if `term`, `skipCurrent`, and `filters` also lived in the redux state, then we wouldn't need to worry about the parameters or the factory in the first place.

hmhealey (approved):

PR #1279, opened 2020-11-02
MM-19509 Prioritize autocomplete results based on interactions and threads

sudheerDev:
#### Summary
  * Add selector makeGetProfilesForThread for getting profiles for threds
  * Add selector makeAddLastViewAtToProfiles to add last_viewed_at for profiles if
    membership exists

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19509

### DIFF
--- a/src/selectors/entities/posts.ts
+++ b/src/selectors/entities/posts.ts
@@ -7,6 +7,7 @@ import {Posts, Preferences} from '../../constants';
 
 import {getCurrentUser} from 'selectors/entities/common';
 import {getMyPreferences} from 'selectors/entities/preferences';
+import {getUsers, getCurrentUserId} from 'selectors/entities/users';
 
 import {Channel} from 'types/channels';
 import {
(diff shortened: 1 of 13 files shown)


src/selectors/entities/posts.ts
```diff
@@ -366,6 +367,29 @@ export function makeGetPostsForThread(): (state: GlobalState, props: {rootId: $I
     );
 }
 
+// The selector below filters current user if it exists. Excluding currentUser is just for convinience
+export function makeGetProfilesForThread(): (state: GlobalState, props: {rootId: $ID<Post>}) => Array<UserProfile> {
```
hmhealey, line 371: 0/5 on this, but I don't think we necessarily need to keep passing the arguments into these selectors as objects. I think reselect intended you to pass component props directly into the selector, but that feels awkward to me.

src/types/users.ts
```diff
@@ -56,6 +56,10 @@ export type UserProfile = {
     terms_of_service_create_at: number;
 };
 
+export type userProfileWithLastViewAt = UserProfile & {
```
hmhealey, line 59: This should be capitalized

src/selectors/entities/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {GlobalState} from 'types/store';
+import {UserProfile, userProfileWithLastViewAt} from 'types/users';
+import {getDirectChannelName} from 'utils/channel_utils';
+
+export function makeAddLastViewAtToProfiles(): (state: GlobalState, profiles: UserProfile[]) => Array<userProfileWithLastViewAt> {
+    return createSelector(
+        getCurrentUserId,
+        getMyChannelMemberships,
+        getAllDirectChannelsNameMapInCurrentTeam,
```
hmhealey, line 17: Why use this selector as opposed to just getting all channels here? I'm wondering if we could use a selector that does slightly less work on the channels. Also 0/5 here though

hmhealey (changes requested):

src/selectors/entities/utils.ts
sudheerDev: Ya, i used simple get channels selector and made necessary changes based on it

src/selectors/entities/posts.ts
sudheerDev: i am not sure i understand. I could avoid even defining them, but TS is complaining about it 🤷‍♂️

nevyangelova (approved):

src/selectors/entities/posts.ts
hmhealey: TS wants them because it's passed into `getPostsForThread`, but I think if you define the selector like
```js
function makeGetSomething(): (state: GlobalState, rootId: string) => void {
    return createSelector(
        (state, rootId) => getPostsForThread(state, {rootId}),
        () => {},
    );
}
```
you could remove the extra object if you wanted though

hmhealey (approved):

PR #1282, opened 2020-11-04
MM-30090 Add ManagedResourcePaths setting

hmhealey:
Due to some changes in 5.28 around the internal/external link handling in the web app, we broke the newly added [Desktop Managed Resources](https://docs.mattermost.com/install/desktop-managed-resources.html) feature which basically lets non-MM services host stuff at certain routes under the MM server.

Since most paths in the server could be a team URL, the new link handling is correct for anyone not using that feature, so instead of reverting the link handling changes, we're adding a way to the server to manually specify these paths that should not be handled by the web app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30090

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/16213
https://github.com/mattermost/mattermost-webapp/pull/7024

### DIFF
--- a/src/selectors/entities/general.test.js
+++ b/src/selectors/entities/general.test.js
@@ -339,5 +339,34 @@ describe('Selectors.General', () => {
             assert.equal(Selectors.getAutolinkedUrlSchemes(state), Selectors.getAutolinkedUrlSchemes(state));
         });
     });
+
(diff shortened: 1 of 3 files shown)


jespino (approved):
LGTM

Willyfrog (approved):
LGTM!

PR #1284, opened 2020-11-05
[MM-20581] Change `Groups` constant to `enum`

devinbinnie:
#### Summary
Our type for `SyncableType` expects some form of string literal, which is not provided by our current constant implementation. This PR changes the type of the constant to `enum` so that the TS compiler expects that the value of the constant is always the value specified and not some other string.

NOTE:
The correct way of doing this is to actually change the type of `SyncableType` to an `enum` as well, but that would require a lot of changes to the webapp, so we'll do that across the board in another PR.

Brought up in this PR: https://github.com/mattermost/mattermost-webapp/pull/6924#discussion_r516829735

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20581

### DIFF
--- a/src/constants/emoji.ts
+++ b/src/constants/emoji.ts
@@ -1,7 +1,5 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-const Emoji = {
+export default {
     SORT_BY_NAME: 'name',
 };
-
-export default Emoji;
--- a/src/constants/emoji.ts
+++ b/src/constants/emoji.ts
@@ -1,5 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export default {
+const Emoji = {
     SORT_BY_NAME: 'name',
 };
+
+export default Emoji;
--- a/src/constants/groups.ts
+++ b/src/constants/groups.ts
@@ -1,6 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export default {
-    SYNCABLE_TYPE_TEAM: 'team',
-    SYNCABLE_TYPE_CHANNEL: 'channel',
-};
+
+enum Groups {
+    SYNCABLE_TYPE_TEAM = 'team',
+    SYNCABLE_TYPE_CHANNEL = 'channel',
+}
+
+export default Groups;
--- a/src/constants/groups.ts
+++ b/src/constants/groups.ts
@@ -1,6 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export default {
+
+import {SyncableType} from 'types/groups';
+
+// See LICENSE.txt for license information.
+const Groups: {[key: string]: SyncableType} = {
     SYNCABLE_TYPE_TEAM: 'team',
     SYNCABLE_TYPE_CHANNEL: 'channel',
 };
+
+export default Groups;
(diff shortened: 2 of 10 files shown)


src/constants/groups.ts
```diff
@@ -1,6 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export default {
+
+import {SyncableType} from 'types/groups';
+
+// See LICENSE.txt for license information.
```
hmhealey, line 6: Accidentally duplicated comment

src/constants/emoji.ts
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-export default {
+const Emoji = {
```
hmhealey, line 3: Have you checked if this makes a difference to the compiled code? I'm pretty sure this is functionally the same as before.

src/constants/groups.ts
```diff
@@ -1,6 +1,12 @@
...
+const Groups: {[key: string]: SyncableType} = {
```
hmhealey, line 7: This is actually a type change since the export is going from the type `{SYNCABLE_TYPE_TEAM: X, SYNCABLE_TYPE_CHANNEL: X}` to `{[key: string]: X}`, so `Groups.Foobarasdf` goes from being a type error to just returning undefined.

In this case, I think we need to cast the `'team'` and `'channel'` literals to `SyncableType` or find some way to tell TypeScript that it should do that TypeScript without using a type like this that makes keys potentially be undefined

src/constants/roles.ts
```diff
@@ -1,8 +1,13 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-export default {
+import {ChannelModerationRoles} from 'types/roles';
+import {Dictionary} from 'types/utilities';
+
+const Roles: Dictionary<ChannelModerationRoles> = {
```
hmhealey, line 7: Ditto here. Dictionary vs `{[key: string]: X}` vs Record is definitely confusing.

src/constants/groups.ts
hmhealey: Also, I think this is the big thing you're wanting to fix here since the problem seems to be that Groups.SYNCABLE_TYPE_TEAM is not actually a SyncableType currently

src/constants/emoji.ts
reflog: I'm not sure there's any difference between the two ways of writing this. What's the benefit?

src/constants/groups.ts
reflog: Suggestion - replace all {[key: string]: SyncableType} / Dictionary / OneToOneRelashionship stuff that we already have with the built in 'Record' type.

src/constants/emoji.ts
devinbinnie: I was under the impression that when we declared these objects, all of their values defaulted to the type of the value we assigned. If we declare as `const`, we know that the object can't change and thus issues with comparing these values to a string literal type are gone since the value can only be the specified string literal.

I might be wrong in this belief, maybe the issue is actually just the typing of the object itself, like in the `groups.ts` case.

src/constants/groups.ts
devinbinnie: @reflog That's a much bigger undertaking, and I'd like to do it. But without this change we are currently blocking this contributor PR: https://github.com/mattermost/mattermost-webapp/pull/6924.

Can we merge as-is and come back to fixing this properly later?

src/constants/emoji.ts
hmhealey: `const` in JS is like final in Java, so the contents of Emoji can change in both cases. When you use `export default`, I don't think the reference can change anyway because it doesn't have a name.

src/constants/groups.ts
hmhealey: Agree on doing that separately. We know that there's issues with the types defined here, but we haven't been able to devote much time to fixing them as a whole

hmhealey (approved):

reflog (approved):

PR #1286, opened 2020-11-06
MM-20897 Add category muting

hmhealey:
A new sidebar feature is muting entire categories at once. Muting a category mutes all channels in it and makes it so that channels moved into it get automatically muted.

Nothing super special to note here, but I went back on just relying on the websocket events to update categories because it caused a weird pop-in where each channel in a category muted one-by-one because different websocket events were muting each channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20897

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/16225
https://github.com/mattermost/mattermost-webapp/pull/7033

### DIFF
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -52,7 +52,7 @@ export function setCategorySorting(categoryId: string, sorting: CategorySorting)
         const state = getState();
         const category = getCategory(state, categoryId);
 
-        dispatch(updateCategory({
+        return dispatch(updateCategory({
             ...category,
             sorting,
         }));
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -59,6 +59,18 @@ export function setCategorySorting(categoryId: string, sorting: CategorySorting)
     };
 }
 
+export function setCategoryMuted(categoryId: string, muted: boolean) {
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
+        const category = getCategory(state, categoryId);
+
+        return dispatch(updateCategory({
+            ...category,
+            muted,
+        }));
+    };
+}
+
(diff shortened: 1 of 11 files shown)


src/actions/channel_categories.ts
```diff
@@ -52,13 +52,45 @@ export function setCategorySorting(categoryId: string, sorting: CategorySorting)
         const state = getState();
         const category = getCategory(state, categoryId);
 
-        return dispatch(updateCategory({
+        dispatch(updateCategory({
```
devinbinnie, line 55: Don't we want most actions to return data? This might just be a TS headache but when we have actions that don't return some `{data: any; error: Error}` the redux connector complains from a TS POV because the return type doesn't match `ActionFunc`.

src/constants/channels.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const NotificationLevel = {
```
devinbinnie, line 4: Should we start moving towards `enum` like we discussed?

devinbinnie (approved):

src/constants/channels.ts
hmhealey: I'm leaving this one as-is for now since it's existing field in a poorly-typed data structure (NotifyProps), but I agree with enums for new types going forward.

src/actions/channel_categories.ts
hmhealey: Good catch. I think I removed this by mistake.

In general, I'd like to get away from requiring every action to return something, but that's something for later.

devinbinnie: Agreed, I'd like to fix the TS headache since there's no requirement to return anything.

sudheerDev (approved):
LGTM

PR #1295, opened 2020-11-12
[MM-30158] Added method for moving multiple channels from any number of categories to one category.

devinbinnie:
#### Summary
For multi-select drag and drop, we needed a reliable way of moving an arbitrary selection of channels from n categories to a single index in a category. I've leveraged the existing code and modified it to work with multiple and the re-arranging necessary to do so. Luckily the back-end already supports this!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30158

### DIFF
--- a/src/actions/channel_categories.ts
+++ b/src/actions/channel_categories.ts
@@ -31,7 +31,7 @@ import {CategorySorting, OrderedChannelCategories, ChannelCategory} from 'types/
 import {Channel} from 'types/channels';
 import {$ID} from 'types/utilities';
 
-import {insertWithoutDuplicates, removeItem} from 'utils/array_utils';
+import {insertMultipleWithoutDuplicates, insertWithoutDuplicates, removeItem} from 'utils/array_utils';
 
 export function expandCategory(categoryId: string) {
     return {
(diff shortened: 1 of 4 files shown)


src/utils/array_utils.test.ts
```diff
@@ -26,6 +26,34 @@ describe('insertWithoutDuplicates', () => {
...
+        expect(insertMultipleWithoutDuplicates(['a', 'b', 'c', 'd'], ['z', 'y', 'x', 'a', 'c'], 0)).toEqual(['z', 'y', 'x', 'a', 'c', 'b', 'd']);
+        expect(insertMultipleWithoutDuplicates(['a', 'b', 'c', 'd'], ['z', 'y', 'x', 'a', 'c'], 1)).toEqual(['b', 'z', 'y', 'x', 'a', 'c', 'd']);
+        expect(insertMultipleWithoutDuplicates(['a', 'b', 'c', 'd'], ['z', 'y', 'x', 'a', 'c'], 2)).toEqual(['b', 'd', 'z', 'y', 'x', 'a', 'c']);
+    });
+
+    test('should return the original array if nothing changed', () => {
+        const input = ['a', 'b', 'c', 'd'];
+
+        expect(insertMultipleWithoutDuplicates(input, ['a', 'b', 'c'], 0)).toStrictEqual(input);
+    });
```
Willyfrog, line 54: Just for completeness maybe adding three test cases for edge cases:
- array = [], should return whatever is in the items array
- items = [], should return the same original array
- newIndex to be greater than the array or be negative

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
...
+
+            const originalCategories = Object.values(unmodifiedCategories).reduce((allCategories: ChannelCategory[], category) => {
+                allCategories.push(category);
+                return allCategories;
+            }, []);
+
+            dispatch({
+                type: ChannelCategoryTypes.RECEIVED_CATEGORIES,
+                data: originalCategories,
+            });
```
Willyfrog, line 299: not sure if we shouldn't trigger a request for the categories from the server, if there was an error there is the possibility that client's info is not right.

Willyfrog (approved):
Looking good, left a couple of comments, but I don't think any of them are really blockers.

src/actions/channel_categories.ts
devinbinnie: This is just replacing the categories that we updated above outside of the `catch` back to what they originally were. We can assume the client data that existed before any updates happened is correct since there was no error leading up to that data existing.

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
     };
 }
 
+export function moveChannelsToCategory(categoryId: string, channelIds: string[], newIndex: number, setManualSorting = true) {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
+        const targetCategory = getCategory(state, categoryId);
+        const currentUserId = getCurrentUserId(state);
+
+        // Add the channel to the new category
```
hmhealey, line 243:
```suggestion
        // Add the channels to the new category
```

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
...
+                sorting: (setManualSorting && targetCategory.type !== CategoryTypes.DIRECT_MESSAGES) ? CategorySorting.Manual : targetCategory.sorting,
+                channel_ids: insertMultipleWithoutDuplicates(targetCategory.channel_ids, channelIds, newIndex),
+            },
+        };
+
+        // Needed if we have to revert categories and for checking for favourites
+        let unmodifiedCategories = {[targetCategory.id]: targetCategory};
+        let sourceCategories: Record<string, string> = {};
+
+        // And remove it from the old category
```
hmhealey, line 256:
```suggestion
        // And remove it from the old categories
```

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
...
+        channelIds.forEach((channelId) => {
+            const sourceCategory = getCategoryInTeamWithChannel(getState(), targetCategory.team_id, channelId);
```
hmhealey, line 258: 0/5 but I wonder if there's any performance benefit to checking `sourceCategories` first instead of using the selector. That could also avoid the repeated `categories[sourceCategory.id] || sourceCategory` below

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
     };
 }
 
+export function moveChannelsToCategory(categoryId: string, channelIds: string[], newIndex: number, setManualSorting = true) {
```
hmhealey, line 237: You probably don't need the `setManualSorting` parameter with this one. `moveChannelToCategory` singular only has that for `addChannelToCategory`.

src/actions/channel_categories.ts
```diff
@@ -234,6 +234,85 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
...
+
+            dispatch({
+                type: ChannelCategoryTypes.RECEIVED_CATEGORIES,
+                data: originalCategories,
+            });
+            return {error};
+        }
+
+        // Update the favorite preferences locally on the client in case we have any logic relying on that
+        await Promise.all(channelIds.map(async (channelId) => {
```
hmhealey, line 304: I think this might not work right. I think the map might return an array of async functions without actually calling them, but I'm not 100% sure

src/utils/array_utils.ts
```diff
@@ -24,6 +24,19 @@ export function insertWithoutDuplicates<T>(array: T[], item: T, newIndex: number
     return newArray;
 }
 
+export function insertMultipleWithoutDuplicates<T>(array: T[], items: T[], newIndex: number) {
+    let newArray = [...array];
```
hmhealey, line 28: This does return a new array even if nothing moved, but that's probably fine. I can't imagine people will use the bulk move within a category very often. I can't think of a concise way to check if nothing moved anyway without just doing a shallow equals check.

hmhealey (changes requested):

src/actions/channel_categories.ts
devinbinnie: Don't think so, `sourceCategories` shouldn't have a record for this channel ID unless the channel was duplicated somehow, so I think we're stuck with the selector.

src/actions/channel_categories.ts
devinbinnie: `Promise.all` executes all of the functions in the array provided and will wait for all of them to complete.

src/actions/channel_categories.ts
devinbinnie: We'll need this for favouriting and muting as a bulk operation.

src/actions/channel_categories.ts
hmhealey: Right

hmhealey (approved):

PR #1300, opened 2020-11-16
[MM-27927] - Add user preference for limit on dms and gms

ghost:
#### Summary
As part of https://github.com/mattermost/mattermost-webapp/pull/7068 this PR adds the user preference to set a limit on unread DMs and GMs from the new sidebar menu dropdown.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27927

NOTE: Need to add tests for the new selector

### DIFF
--- a/src/selectors/entities/channel_categories.ts
+++ b/src/selectors/entities/channel_categories.ts
@@ -239,19 +239,21 @@ export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channe
                     }
                 }
 
-                // Skip the rest of the checks if autoclosing inactive DMs is disabled
-                if (!autocloseDMs) {
-                    return true;
-                }
-
-                return false;
+                return true;
             });
 
             // The limit of DMs user specifies to be rendered in the sidebar
             const limitPref = myPreferences[getPreferenceKey(Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.LIMIT_VISIBLE_DMS_GMS)];
             const limit = parseInt(limitPref ? limitPref.value! : '0', 10);
 
-            const remaining = limit - unreadCount;
+            let remaining;
+
+            if (limit < unreadCount) {
+                remaining = unreadCount;
+            } else {
+                remaining = limit;
+            }
+
             return filtered.slice(0, remaining);
         },
     );
--- a/src/selectors/entities/channel_categories.ts
+++ b/src/selectors/entities/channel_categories.ts
@@ -249,7 +249,7 @@ export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channe
 
             // The limit of DMs user specifies to be rendered in the sidebar
             const limitPref = myPreferences[getPreferenceKey(Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.LIMIT_VISIBLE_DMS_GMS)];
-            const limit = parseInt(limitPref?.value || '0', 10);
+            const limit = parseInt(limitPref ? limitPref.value! : '0', 10);
 
             const remaining = limit - unreadCount;
             return filtered.slice(0, remaining);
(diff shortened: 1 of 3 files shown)


devinbinnie (approved):

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,76 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
+                    return true;
+                }
+
+                if (currentChannelId === channel.id) {
+                    // The current channel is always visible
+                    return true;
+                }
+
+                // openTime is the time the channel was last opened (like from the More DMs list) after having been closed
+                const openTimePref = myPreferences[getPreferenceKey(Preferences.CATEGORY_CHANNEL_OPEN_TIME, channel.id)];
```
hmhealey, line 228: Instead of looking at this, let's compare the deactivated time against the channel member's `last_viewed_at` since that's a lot more widely used. The OPEN_TIME preference is something I didn't know existed until I started digging into this, and I'm pretty sure it has issues where it's not set right. I'd like to remove it eventually.

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,76 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
     );
 }
 
+export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channel[], categoryType: string) => Channel[] {
+    return createSelector(
+        (state: GlobalState, channels: Channel[]) => channels,
+        (state: GlobalState, channels: Channel[], categoryType: string) => categoryType,
+        getMyPreferences,
```
hmhealey, line 196: Also, if you stop using the OPEN_TIME preferences, you can just select your new preference as a number to pass in here and make this recalculate a bit less often.

src/selectors/entities/channel_categories.ts
```diff
@@ -384,18 +454,20 @@ export function makeGetChannelsForCategory() {
...
     const sortChannels = makeSortChannels();
 
     return (state: GlobalState, originalChannels: Channel[], category: ChannelCategory) => {
+        const sidebarOrganisationPreference = getNewSidebarPreference(state);
         let channels = originalChannels;
 
         channels = filterArchivedChannels(state, channels);
 
-        channels = filterAutoclosedDMs(state, channels, category.type);
+        channels = sidebarOrganisationPreference ? filterAutoclosedDMs(state, channels, category.type) : legacyFilterAutoclosedDMs(state, channels, category.type);
```
hmhealey, line 470: Double checking, but is this actually used by the DM/GM channel loading code? If not, these selectors aren't used by the old sidebar, so we don't need to keep `legacyFilterAutoclosedDMs` around any more since the old sidebar is controlled by a different selector.

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,76 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
+
+            let remaining;
+
+            if (limit < unreadCount) {
+                remaining = unreadCount;
+            } else {
+                remaining = limit;
+            }
+
+            return filtered.slice(0, remaining);
```
hmhealey, line 257: This isn't going to work unfortunately since it'll slice the first X channels alphabetically or with the X most recent posts (based on category sorting) instead of the last X viewed. We'll need to do something like constructing an array of the last X viewed channels and then use that as part of the above filter instead.

hmhealey: Also, if you could add the `return filtered.length === channels.length ? channels : filtered;` check like above, that'd help reduce re-renders if this doesn't actually end up filtering anything out.

hmhealey (changes requested):

src/selectors/entities/channel_categories.ts
nevyangelova: Oh yes I remember you did mention that, I forgot this part of the conversation SORRY!

hmhealey (approved):

PR #1310, opened 2020-12-04
[MM-27927] - Add user preference for limit on dms and gms

nevyangelova:
There is an error with category sorting, when you are currently on a DM which gets sorted last (or in between the last messages) alphabetically, the DM will disappear when you limit the view count (from 20 to 10 for example)


src/selectors/entities/channel_categories.ts
```diff
@@ -248,7 +248,12 @@ export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channe
                 remaining = limitPref;
             }
 
+            const currentChannel = filtered.find((channel) => channel.id === currentChannelId);
             const slicedChannels = filtered.slice(0, remaining);
+            if (currentChannel && !slicedChannels.includes(currentChannel)) {
+                slicedChannels.pop();
```
hmhealey, line 254: Which channel gets popped off by this? Maybe we should just let the count be one over in this case?

src/selectors/entities/channel_categories.ts
```diff
@@ -248,7 +248,12 @@ export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channe
                 remaining = limitPref;
             }
 
+            const currentChannel = filtered.find((channel) => channel.id === currentChannelId);
```
hmhealey, line 251: Instead of looking through filtered, I think it would be better to pass the current channel into this instead of the current channel ID. That way, we don't need to null check `currentChannel`

hmhealey (changes requested):

hmhealey:
Also, could you add a test case for this case if it's not too difficult to recreate that state?

devinbinnie (approved):
LGTM other than Harrison's suggestions

src/selectors/entities/channel_categories.ts
nevyangelova: In that case we still need to add another selector and get the channel as an extra step, IMO I think this fits more with the flow

nevyangelova:
> Also, could you add a test case for this case if it's not too difficult to recreate that state?

there is a test case, but it checks for only one channel I would add a second one to recreate more accurately :P

hmhealey (approved):

hmhealey:
Leaving this to merge once the webapp PR is ready so that we don't end up breaking master like we did last time.

nevyangelova:
I went with a different approach and sorted channels by unreads first and last_viewed_at second, and that determines the cutout.

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,98 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
+                    }
+
+                    return -1;
+                }
+
+                if (isUnreadChannel(myMembers, channelB)) {
+                    return 1;
+                }
+
+                return sortByLastViewedAt(channelA, channelB);
```
hmhealey, line 241: I think you can reorder this a bit to save some typing by checking if they're different first and then handling both of them being either true or false afterwards
```suggestion
                
                if (isUnreadChannel(myMembers, channelA) && !isUnreadChannel(myMembers, channelB)) {
                    return -1;
                } else if (!isUnreadChannel(myMembers, channelA) && isUnreadChannel(myMembers, channelB)) {
                    return 1;
                }

                return sortByLastViewedAt(channelA, channelB);
```

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,98 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
+
+            function sortByLastViewedAt(channelA: Channel, channelB: Channel) {
+                const channelALastViewedAt = myMembers[channelA.id]?.last_viewed_at || 0;
+                const channelBLastViewedAt = myMembers[channelB.id]?.last_viewed_at || 0;
+
+                if (channelBLastViewedAt === channelALastViewedAt) {
+                    return 0;
+                }
+
+                if (channelALastViewedAt === 0) {
```
hmhealey, line 217: I don't think these cases are necessary since we'd want a channel that has never been viewed to be last. If it's something the user was just invited to, it should be unread, so that part of the sorting will handle it.

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,98 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
+                    return -1;
+                }
+
+                if (channelBLastViewedAt === 0) {
+                    return 1;
+                }
+
+                return Math.sign(channelBLastViewedAt - channelALastViewedAt);
```
hmhealey, line 225: Strictly speaking, I don't think we need `Math.sign` here since any positive or negative values are treated as 1 and -1 respectively. I do have a funny story about a production bug I caused with this sort of code at my previous job though...

hmhealey (changes requested):

src/selectors/entities/channel_categories.ts
nevyangelova: If I remove this check it will sort the channels with last_viewed_at === 0 last and from my understanding we want to prioritise them.

src/selectors/entities/channel_categories.test.js
```diff
@@ -541,6 +541,193 @@ describe('makeFilterAutoclosedDMs', () => {
...
+            },
+        });
+
+        expect(filterAutoclosedDMs(state, [dmChannel1, gmChannel1], CategoryTypes.DIRECT_MESSAGES)).toEqual([gmChannel1]);
+    });
+
+    describe('Should always show the exact number of channels specified by the user', () => {
+        const filterAutoclosedDMs = Selectors.makeFilterAutoclosedDMs();
+        const tigerKing = {id: 'tigerKing'};
+        const bojackHorseman = {id: 'bojackHorseman'};
```
devinbinnie, line 658: Nice lol

src/selectors/entities/channel_categories.ts
devinbinnie: Is it possible to have a DM that has never been viewed open at all? Even one that was manually closed should have a `last_viewed_at` > 0.

devinbinnie (approved):

src/selectors/entities/channel_categories.ts
nevyangelova: From what I understood brand new DM's will have a last_viewed_at === 0

hmhealey: If another user creates it and sends a message, it will be 0 but it will be visible since it's unread. That's the only case I can think of where we'd have one like that naturally

src/selectors/entities/channel_categories.ts
```diff
@@ -189,6 +189,96 @@ export function makeFilterAutoclosedDMs(getAutocloseCutoff = getDefaultAutoclose
...
 
+export function makeFilterAutoclosedDMs(): (state: GlobalState, channels: Channel[], categoryType: string) => Channel[] {
+    return createSelector(
+        (state: GlobalState, channels: Channel[]) => channels,
+        (state: GlobalState, channels: Channel[], categoryType: string) => categoryType,
+        getCurrentChannelId,
+        (state: GlobalState) => state.entities.users.profiles,
+        getCurrentUserId,
+        getMyChannelMemberships,
+        (state: GlobalState) => getInt(state, Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.LIMIT_VISIBLE_DMS_GMS),
```
hmhealey, line 200: This needs a default value since it defaults to showing 0 DMs right now

hmhealey (changes requested):

hmhealey (approved):

devinbinnie (approved):

PR #1311, opened 2020-12-06
MM-31126 Collapsed threads support

reflog:


src/types/users.ts
```diff
@@ -23,7 +23,7 @@ export type UserNotifyProps = {
 };
 
 export type UserProfile = {
-    id: string;
+    id: 'me' | string;
```
jespino, line 26: Is there any reason to have this distinction here?

reflog: nope. will remove

src/actions/threads.ts
```diff
@@ -0,0 +1,111 @@
...
+                threads: userThreadList.threads.map((thread) => ({...thread, is_following: true})),
+                team_id: teamId,
+            },
+        });
+
+        return {data: userThreadList};
+    };
+}
+
+export function markThreadsRead(userId: string, teamId: string) {
```
jespino, line 46: maybe something like `markAllTeamThreadsAsRead` from the parameters you can infer that, but is not obvious enough that you are marking as read all the team threads.

src/actions/threads.ts
```diff
@@ -0,0 +1,111 @@
...
+        try {
+            await Client4.updateThreadsReadForUser(userId, teamId);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+
+        dispatch({
+            type: ThreadTypes.THREADS_READ,
```
jespino, line 57: Same with this, this action id is not obvious enough. something like `ALL_TEAM_THREADS_READ` would be more explicit.

src/reducers/entities/threads.test.js
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {ThreadTypes} from 'action_types';
+import deepFreeze from 'utils/deep_freeze';
+import threadsReducer from './threads';
+
+describe('threads', () => {
+    describe('RECEIVED_THREADS', () => {
```
jespino, line 9: Missing a the tests here around the other events that can happen on the threads reducers, like the `FOLLOW_CHANGED_THREAD` and the `LEAVE_TEAM`.

And missing tests for the other existing reducers like `threadsReducer`,  `threadsInTeamReducer` and `countsReducer`.

src/reducers/entities/threads.ts
```diff
@@ -0,0 +1,106 @@
...
+        const {threads} = action.data;
+        return {
+            ...state,
+            ...threads.reduce((results: IDMappedObjects<UserThread>, thread: UserThread) => {
+                results[thread.id] = thread;
+                return results;
+            }, {}),
+        };
+    }
+    case ThreadTypes.FOLLOW_CHANGED_THREAD: {
```
jespino, line 23: What happens if the followed thread is not yet in the store?

src/selectors/entities/preferences.ts
```diff
@@ -8,7 +8,7 @@ import {General, Preferences} from '../../constants';
 import {getConfig, getLicense} from 'selectors/entities/general';
 import {getCurrentTeamId} from 'selectors/entities/teams';
 
-import {PreferenceType} from 'types/preferences';
+import {PreferenceType, Theme} from 'types/preferences';
```
jespino, line 11: Why this changes appear here? Looks like a merge conflict that is resolved reverting changes.

jespino (changes requested):
Some improvement suggestions and a doubt about a possible badly resolved merge conflict.

src/selectors/entities/preferences.ts
calebroseland: @jespino I'm using the `getTheme` selector in a small part the main feature, and I wanted to cleanup the types around it as it's currently typed as returning `any`. Having clearer types during development helps me to be more sure, but by all means, it could be separated in another PR.

src/constants/preferences.ts
```diff
@@ -2,9 +2,8 @@
 // See LICENSE.txt for license information.
 
 import {Theme} from 'types/preferences';
-import {Dictionary} from 'types/utilities';
 
-const Preferences: Dictionary<any> = {
+const Preferences = {
```
calebroseland, line 6: @jespino this is also related to that (`getTheme` cleanup)

src/types/threads.ts
```diff
@@ -0,0 +1,35 @@
...
+    total_unread_mentions: number;
+    threads: UserThread[];
+}
+
+export type ThreadsState = {
+    threadsInTeam: RelationOneToMany<Team, UserThread>;
+    threads: IDMappedObjects<UserThread>;
+    counts: RelationOneToOne<Team, {
+        total: number;
+        total_unread_replies: number;
```
calebroseland, line 32: @reflog I spoke with Matt on this, and indeed this will be used for "You have N threads with unread messages", so we may want to rename this after it's changed server-side.

src/actions/threads.ts
```diff
@@ -0,0 +1,111 @@
...
+            dispatch(logError(error));
+            return {error};
+        }
+
+        dispatch({
+            type: ThreadTypes.READ_CHANGED_THREAD,
+            data: {
+                id: threadId,
+                team_id: teamId,
+                timestamp,
```
calebroseland, line 82: Should we update `unread_replies` and `unread_mentions` here? Or are we just waiting for the websocket event to update these?

src/actions/threads.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        dispatch({
+            type: ThreadTypes.FOLLOW_CHANGED_THREAD,
+            data: {
+                id: threadId,
+                team_id: teamId,
+                following: newState,
+            },
+        });
```
calebroseland, line 107: Should we dispatch this lazily and then again if need be after success, to prevent this on slow connections?
CleanShot 2021-01-11 at 08 51 26

calebroseland (commented):
Just a couple questions after taking a second look at this.

src/actions/threads.ts
reflog: Yeah, I'd leave this for websocket events to keep it consistent.

src/actions/threads.ts
```diff
@@ -0,0 +1,110 @@
...
+        let userThreadList: undefined | UserThreadList;
+
+        try {
+            userThreadList = await Client4.getUserThreads(userId, teamId, {page, pageSize: perPage, extended: true});
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(logError(error));
+            return {error};
+        }
+
```
calebroseland, line 27:
```suggestion
        dispatch({
            type: PostTypes.RECEIVED_PROFILES_LIST,
            data: userThreadList.threads.map(({participants: users}) => users).flat(),
        });

```

calebroseland (changes requested):
One last one to help with profile/user selectors.

calebroseland (approved):

jespino (approved):
LGTM

PR #1313, opened 2020-12-08
Merge latest master as of dec 1st into cloud branch

jwilander:
#### Summary
Forgot to do this one earlier. Shouldn't affect anything since the cloud branch of the webapp points to the correct commit it needs but we should keep this branch up to date.


marianunez (approved):
Do we need this branch at all?

jwilander:
Only if we need to cherry-pick a fix before a release I think

nickmisasi (approved):

PR #1316, opened 2020-12-09
[MM-30851] Replace ExperimentalChannelSidebarOrganization with EnableLegacySidebar

devinbinnie:
#### Summary
Redux changes to change the new sidebar to be on by default and only turned off when `EnableLegacySidebar` is on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30851


nevyangelova (approved):

src/constants/preferences.ts
```diff
@@ -34,7 +34,6 @@ const Preferences: Dictionary<any> = {
     MENTION_KEYS: 'mention_keys',
     USE_MILITARY_TIME: 'use_military_time',
     CATEGORY_SIDEBAR_SETTINGS: 'sidebar_settings',
-    CHANNEL_SIDEBAR_ORGANIZATION: 'channel_sidebar_organization',
```
hmhealey, line 37: Can you leave this constant? I still need it for the backwards compatibility code

src/actions/channels.ts
```diff
@@ -1429,31 +1429,32 @@ export function favoriteChannel(channelId: string, updateCategories = true): Act
 
         Client4.trackEvent('action', 'action_channels_favorite');
 
-        if (getNewSidebarPreference(state)) {
-            // The new sidebar is enabled, so favorite the channel by moving it into the current team's Favorites category
-            if (updateCategories) {
-                const channel = getChannelSelector(state, channelId);
-                const category = getCategoryInTeamByType(state, channel.team_id || getCurrentTeamId(state), CategoryTypes.FAVORITES);
+        if (config.EnableLegacySidebar === 'true') {
```
hmhealey, line 1432: 0/5 on changing it back now, but you could've reduced the diff on these other files by making `getNewSidebarPreference` look at the config instead of the prefs instead of checking the config everywhere.

hmhealey (changes requested):

src/actions/channels.ts
devinbinnie: Reducing code bloat, since we don't need that selector anymore and this is simpler.

hmhealey (approved):

src/actions/channels.ts
hmhealey: I don't think that's bloat. I think it's encapsulating the details of how we know which sidebar is enabled no matter how that's done, so even if we were to change the settings again in the future, we could do that without changing the code that uses it. That said, it's kind of a moot point here since we're just going to be removing it anyway.

PR #1319, opened 2020-12-14
Fix cache

metanerd:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,9 +7,13 @@ include:
   file: private.yml
 
 cache:
-  key: ${CI_COMMIT_REF_SLUG}
+  key:
+    files:
+      - package.json
+    prefix: ${CI_COMMIT_REF_SLUG}
   paths:
     - .npm/
+    - node_modules/
 
 lint:
   stage: test


jaydeland (approved):
Lgtm

.gitlab-ci.yml
```diff
@@ -7,9 +7,13 @@ include:
   file: private.yml
 
 cache:
-  key: ${CI_COMMIT_REF_SLUG}
+  key:
+    files:
+      - package.json
```
hanzei, line 12: Should we also check for changes in  `package-lock.json`? E.g. security updates could only update `package-lock.json`.

.gitlab-ci.yml
```diff
@@ -7,9 +7,13 @@ include:
...
+    prefix: ${CI_COMMIT_REF_SLUG}
   paths:
     - .npm/
```
hanzei, line 15: I'm yet to encounter a `.npm` directory. What is it used for?

hanzei (changes requested):
Thanks for taking a look at the issue!

Should we merge this into `master` directly to ensure other don't run into the same issue?

.gitlab-ci.yml
metanerd: https://docs.gitlab.com/ee/ci/caching/#cache-nodejs-dependencies

.gitlab-ci.yml
metanerd: `package.json` is fine for now, we can add package-lock.json when the issue arises.

jaydeland (commented):
Lgtm

metanerd:
We can merge it into master with your PR @hanzei

hanzei (approved):

PR #1342, opened 2021-01-25
[MM-32087] - Add theme variable for sidebar team bar

ghost:
#### Summary
This PR adds a new theme variable for team sidebar text which would replace the currently constructed colour of the team bar, inclusive of channelHeaderColour + opacity. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32087


devinbinnie (approved):

nevyangelova:
@devinbinnie @hmhealey the theme was getting updated and overwritten by the default theme in multiple places in the code, so I think its best to keep it all in the 'getTheme' selector.

src/selectors/entities/preferences.ts
```diff
@@ -174,7 +175,13 @@ export const getTheme: (state: GlobalState) => Theme = createShallowSelector(
             theme.mentionBg = theme.mentionBj;
         }
 
-        return Object.assign({}, defaultTheme, theme);
+        return {
```
devinbinnie, line 178: We don't have to do this now, but we did talk about having this mapping put in a different function so that the mapping logic wasn't directly mixed with the logic for getting the theme.

devinbinnie (approved):

src/selectors/entities/preferences.ts
nevyangelova: @devinbinnie I remember this discussion but then I saw that we can spread all object properties together here since we are already mutating the theme object. Let me know how you would approach this 👍

src/utils/theme_utils.ts
```diff
@@ -96,20 +93,4 @@ export function blendColors(background: string, foreground: string, opacity: num
     );
 
     return `rgba(${red},${green},${blue},${alpha})`;
-}
-
-// setThemeDefaults will set defaults on the theme for any unset properties.
-export function setThemeDefaults(theme: Theme): Theme {
```
hmhealey, line 102: 0/5 on this right now, but if we leave this function around for the time being, it'll mean that this PR can be merged without breaking the web app. As it is, this PR needs to be merged at the same time as the web app one since anyone else trying to use a new version of mattermost-redux will get an error if the webapp PR hasn't gone in yet.

src/selectors/entities/preferences.test.js
```diff
@@ -175,6 +181,10 @@ describe('Selectors.Preferences', () => {
     });
 
     describe('get theme', () => {
+        beforeEach(async () => {
+            jest.spyOn(ThemeUtils, 'blendColors').mockReturnValue('#efefef');
```
hmhealey, line 185: Why do we need this mock? Shouldn't `getTheme` be able to use the theme you've given it in the test case to actually blend it?

src/selectors/entities/preferences.test.js
```diff
@@ -299,6 +309,34 @@ describe('Selectors.Preferences', () => {
...
+                            DefaultTheme: 'default',
+                        },
+                    },
+                    teams: {
+                        currentTeamId,
+                    },
+                    preferences: {
+                        myPreferences: {
+                            [getPreferenceKey(Preferences.CATEGORY_THEME, '')]: {
+                                category: Preferences.CATEGORY_THEME, name: '', value: JSON.stringify({}),
```
hmhealey, line 329: You can just define this value instead of making it empty and having a separate team-specific theme defined next. The two different preferences are only needed if you're testing team-specific themes.

src/selectors/entities/preferences.test.js
nevyangelova: it requires the function to be mocked to be able to call it and get a value from it otherwise it defaults to undefined.

devinbinnie (approved):

src/selectors/entities/preferences.test.js
hmhealey: But aren't we just calling it normally? It doesn't hit do anything fancy, so it should always just give us the same result whenever we call it.

nevyangelova: It's failing without the mock, as far as I saw its taking some properties from the store that aren't there so I either had to mock the function or build up the store.

src/utils/theme_utils.ts
```diff
@@ -95,6 +95,24 @@ export const blendColors = (background: string, foreground: string, opacity: num
         opacity,
     );
 
+    if (hex) {
+        let r = red.toString(16);
+        let g = green.toString(16);
+        let b = blue.toString(16);
+
+        if (r.length === 1) {
+            r = '0' + r;
```
hmhealey, line 104: This makes me wish we had some sort of `sprintf` equivalent in JS. If this was Go, we could do this in a one-liner like `fmt.Sprintf("#%02X%02X%02X", red, green, blue)`.

Now that I've written that out, it also looks kind of like gibberish though

hmhealey (approved):

PR #1348, opened 2021-01-30
[Server #16802]: Update string values to boolean in "client config".

haardikdharma10:
#### Summary

Updated string values to boolean in `ClientConfig`

#### Ticket Link

Fixes : https://github.com/mattermost/mattermost-server/issues/16802


devinbinnie:
Thanks for the PR @haardikdharma10!
Looks like some of the unit tests are failing, please run `make test` on your local machine and have a look at which tests are failing. Once they're fixed, we can review your PR.

src/selectors/entities/users.test.js
```diff
@@ -592,15 +592,15 @@ describe('Selectors.Users', () => {
...
                         },
                         license: {
-                            LockTeammateNameDisplay: 'true',
+                            LockTeammateNameDisplay: true,
                         },
                     },
                 },
             };
-            assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'username');
+            assert.deepEqual(Selectors.makeGetDisplayName()(newTestState, testUser1.id), 'First Last');
```
devinbinnie, line 603: I don't think this is the intention of the test. We should be seeing the username and not the full name of the user here. Sounds like something is wrong is the selector perhaps.

devinbinnie (changes requested):
Thanks for the PR @haardikdharma10!
I've made some comments for you to review.

src/selectors/entities/users.test.js
haardikdharma10: Hmm, when changing it back to `username`, the test fails. What could be wrong here? 🤔

devinbinnie: Have a look at the selector and investigate a bit. Perhaps some `console.log` statements might help out?

hmhealey:
I've added the "Awaiting PR" label to this since it won't work without https://github.com/mattermost/mattermost-server/issues/16839 being implemented. Right now, the server can only send us the client config values as strings, but once that's fixed with a new API (or a tweak to the existing one), then this change can be done.

At that point, it'll also require accompanying webapp changes as well.

src/selectors/entities/users.test.js
hmhealey: Taking another look at this, it's failing because you changed `license.LockTeammateNameDisplay` to a boolean in the tests, when it's still expecting a string in the other code. You'll want to revert that since we're not changing the types in the license with this.

haardikdharma10: Thank you @hmhealey, that was exactly the reason tests were failing. Should be good to go now:)

haardikdharma10:
@devinbinnie, gentle ping for review:)

devinbinnie (commented):
Code looks good, just waiting on the server changes before I make a final approval.
Thanks for the work on this @haardikdharma10 :)

hmhealey:
@mkraft I just had a thought about this. We're changing boolean fields to be their correct type, but what about numeric fields? It's also possible that we have arrays and other non-string fields we might want to fix the type for as well when doing this. Since we're changing the format coming from the server, we might have to do these all at once.

mkraft:
> @mkraft I just had a thought about this. We're changing boolean fields to be their correct type, but what about numeric fields? It's also possible that we have arrays and other non-string fields we might want to fix the type for as well when doing this. Since we're changing the format coming from the server, we might have to do these all at once.

Yikes, I didn't realize we were doing numbers and arrays as string! Good callout. Would we be adding a bunch of overhead to do those as one separate PR? I just don't want to bait-and-switch the work for @haardikdharma10.

hmhealey:
It all needs to be done at once since it requires a breaking API change on the server to return the correct types. These redux changes would have to use the new API, so it'd need to include all the new types.

hmhealey:
@haardikdharma10 We're going to be shortly deprecating mattermost-redux as a standalone repository and moving it into [mattermost-webapp](https://github.com/mattermost/mattermost-webapp) as part of an effort to make development of both codebases easier by using a monorepo. This is going to require either merging or closing all outstanding pull requests in this repository in preparation for the migration planned for tomorrow.

For this PR, I know it's waiting on server changes in a separate ticket, and there's still some outstanding issues regarding changing other field types as well. I'd be glad to help getting this reopened in the webapp repo after our migration is done tomorrow though. Let me know what you think.

haardikdharma10:
Hi @hmhealey, sounds good to me. I'll close this for now.

hmhealey:
Thanks! I'll let you know when it can be reopened against the webapp repo.

hmhealey:
@haardikdharma10 Thanks for waiting on that. mattermost-redux now lives [here](https://github.com/mattermost/mattermost-webapp/tree/master/packages/mattermost-redux), so future mattermost-redux changes can be submitted there.

For this PR, I still think we want to do this, but I think we need to figure out what we want to do with the client config as a whole since there's more fields that need their types changes as I mentioned before. I'd be glad to work with you on that since it's something that would be really beneficial, but I think the existing tickets need to be discussed and updated since they only cover one part of the structure.

haardikdharma10:
Hi @hmhealey, thanks for the info. I'd be happy to open up this PR again in the webapp repo in a while. I'd also be happy to discuss with you regarding other changes. Please let me know when is a good time to ping you on the community server :)

hmhealey:
Hey. We discussed this a bit more internally, and there's some bigger issues we discovered since changing the type in the redux state will break all plugins relying on the config only containing strings. We still need to figure out how what parts of mattermost-redux we want accessible by plugins and how we can do that to best ensure backwards compatibility, so we're going to close these tickets for the time being while we sort that all out. We don't have a specific timeline for that yet, but that will likely involve breaking changes, so it would be a good time to break this for as well.

Still, we appreciate your work on this, and we can definitely let you know when we start trying to fix these types later if you're still interested in working on it then.

PR #1349, opened 2021-02-01
Custom status feature

chetanyakan:
#### Summary
- Add types and actions to set and unset custom status.

#### Ticket Link
- https://github.com/mattermost/mattermost-server/pull/16835
- Specs: https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes

### DIFF
--- a/src/actions/preferences.ts
+++ b/src/actions/preferences.ts
@@ -107,7 +107,7 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
     };
 }
 
-export function setCustomStatusInitialisationState(initializationState: string) {
+export function setCustomStatusInitialisationState(initializationState: Record<string, boolean>) {
     return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
         const state = getState();
         const currentUserId = getCurrentUserId(state);
@@ -115,7 +115,7 @@ export function setCustomStatusInitialisationState(initializationState: string) 
             user_id: currentUserId,
             category: Preferences.CATEGORY_CUSTOM_STATUS,
             name: Preferences.NAME_CUSTOM_STATUS_TUTORIAL_STATE,
-            value: initializationState,
+            value: JSON.stringify(initializationState),
         };
         await dispatch(savePreferences(currentUserId, [preference]));
     };
--- a/src/actions/preferences.ts
+++ b/src/actions/preferences.ts
@@ -107,6 +107,20 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
     };
 }
 
+export function setCustomStatusInitialisationState(initializationState: string) {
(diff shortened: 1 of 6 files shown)


esethna:
Test server from brightscout that includes webapp, redux and server changes: http://ec2-3-137-148-232.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=559fit8sj7bm8n7mcpqcz9cy5y

reflog (changes requested):
I see actions to update the status, but nothing in the store is updated. How are the changes reflected back to the UI?

chetanyakan:
@reflog 

Here, we just make the API call to update the user's custom status.
In the server, in the API implementation, we update the user's props with the current status.

Following that, any update to any `User` is sent to all logged-in users via the WebSocket events.
In the web-app, we just use the updated values of the user props to show the custom status set by that user.


***
This implementation follows the one suggested by @jespino here: https://mattermost.atlassian.net/wiki/spaces/CS/pages/1096810497/Development+Spec

reflog:
@chetanyakan - sounds reasonable, thanks!

reflog (approved):

chetanyakan:
Rebased to the HEAD of master

esethna:
@hmhealey can you help with reviews here and on the webapp PR as needed? We're hoping to merge this for the v5.32 deadline (early next week)

src/actions/preferences.ts
```diff
@@ -102,6 +102,20 @@ export function makeGroupMessageVisibleIfNecessary(channelId: string): ActionFun
     };
 }
 
+export function setCustomStatusInitialisationState(initializationState: string) {
```
hmhealey, line 105: 0/5 at this point, but it might be nice to pass in an object here and have this function be smart enough to stringify it for you. Similarly, you could add a selector to `selectors/entities/preferences` that knows to parse the JSON and return it as an object.

hmhealey (commented):
Code looks good, but I still have to finish reviewing the webapp, so I'm holding off on approving right now in case there's redux changes required by webapp ones.

src/actions/preferences.ts
hmhealey: Never mind on the second half of that. I see you have selectors that do that on the webapp-side

chetanyakan:
Rebased to the HEAD of master.

jespino (approved):
Looks good to me. Thanks @chetanyakan! :tada:

src/actions/preferences.ts
chetanyakan: Done.

ogi-m (approved):
Approving the PR (notes are on the webapp PR)

chetanyakan:
Rebased to the HEAD of master.

PR #1351, opened 2021-02-03
MM-31503 - create new restrict screens 4 users

pablovelezvidal:
#### Summary
Adds the selector for the subscriptionStats used in https://github.com/mattermost/mattermost-webapp/pull/7468

#### Ticket Link
[MM-31503](https://mattermost.atlassian.net/browse/MM-31503)

### DIFF
--- a/src/selectors/entities/general.ts
+++ b/src/selectors/entities/general.ts
@@ -37,7 +37,7 @@ export function warnMetricsStatus(state: GlobalState): any {
     return state.entities.general.warnMetricsStatus;
 }
 
-export function subscriptionStatsFromState(state: GlobalState): any {
+export function getSubscriptionStats(state: GlobalState): any {
     return state.entities.cloud.subscriptionStats;
 }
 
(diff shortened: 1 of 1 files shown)


src/selectors/entities/general.ts
```diff
@@ -37,6 +37,10 @@ export function warnMetricsStatus(state: GlobalState): any {
     return state.entities.general.warnMetricsStatus;
 }
 
+export function subscriptionStatsFromState(state: GlobalState): any {
+    return state.entities.cloud.subscriptionStats;
+}
+
```
nickmisasi, line 43: I think the name getSubscriptionStats is more in-line with the rest of the selectors in this file, and I think it highlights that this is a selector a little better at a glance

nickmisasi (changes requested):

AGMETEOR (commented):
Other than what Nick has recommended, all looks good.

src/selectors/entities/general.ts
pablovelezvidal: Thanks @nickmisasi ,ok, I will change it then!

AGMETEOR (approved):

nickmisasi (approved):

PR #1355, opened 2021-02-04
MM-28736 Update dependencies

hmhealey:
I'll make notes below if any changes are particularly notable.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28736


package.json
```diff
@@ -37,59 +37,52 @@
...
+    "cross-env": "7.0.3",
+    "eslint": "7.18.0",
+    "eslint-plugin-header": "3.1.0",
+    "eslint-plugin-import": "2.22.1",
     "eslint-plugin-mattermost": "github:mattermost/eslint-plugin-mattermost#46ad99355644a719bf32082f472048f526605181",
-    "husky": "4.2.5",
+    "husky": "4.3.8",
     "isomorphic-fetch": "3.0.0",
-    "jest": "25.2.7",
-    "jest-junit": "10.0.0",
```
hmhealey, line 79: I need to confirm whether or not this is still needed

src/actions/posts.ts
```diff
@@ -206,8 +206,8 @@ export function createPost(post: Post, files: any[] = []) {
         dispatch({
             type: PostTypes.RECEIVED_NEW_POST,
             data: {
-                id: pendingPostId,
```
hmhealey, line 209: These had to change because TypeScript expected `newPost` to be a complete Post object which meant that it thought this ID was being overwritten.

The only case where this ID was being overwritten was in the corresponding test that I changed, and that test was invalid since the request was never actually made to create the post. To fix that, I changed it to use `createPostImmediately` instead which does send the request, and then I had the server respond with the post ID instead.

src/client/websocket_client.ts
```diff
@@ -13,12 +14,12 @@ class WebSocketClient {
     token: string|null;
     sequence: number;
     connectFailCount: number;
-    eventCallback?: Function;
-    firstConnectCallback?: Function;
-    reconnectCallback?: Function;
-    errorCallback?: Function;
-    closeCallback?: Function;
-    connectingCallback?: Function;
+    eventCallback?: (message: any) => void;
```
hmhealey, line 17: The new version of typescript-eslint doesn't want us to use Function any more, so I had to properly define what types of functions we were looking for.

src/utils/helpers.ts
```diff
@@ -1,14 +1,18 @@
...
 // See LICENSE.txt for license information.
+
 import * as reselect from 'reselect';
 import shallowEqual from 'shallow-equals';
+
 import {Dictionary} from 'types/utilities';
-export function memoizeResult<F extends Function>(func: F): any {
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function memoizeResult<F extends Function>(func: F): F {
```
hmhealey, line 10: This is the only case where I couldn't figure out how to get rid of `Function`, so we have to do some ugly casting through `unknown` below as well.

src/reducers/entities/gifs.ts
```diff
@@ -4,9 +4,10 @@
 import {combineReducers} from 'redux';
 import {GifTypes} from 'action_types';
 import {GenericAction} from 'types/actions';
-import {Dictionary} from 'types/utilities';
 
-const SEARCH_SELECTORS: Dictionary<Function> = {
+type ReducerMap = {[actionType: string]: (state: any, action: GenericAction) => any};
```
devinbinnie, line 8: Might be worth making this a utility type rather than just for this reducer.

devinbinnie (approved):

src/reducers/entities/gifs.ts
hmhealey: Yeah, I was thinking about that. I do kind of like this way of writing reducers instead of using a big switch statement, but I'm not wanting to make any big changes right now

calebroseland (approved):

jgilliam17 (approved):
After discussing with Harrison, QA will test these changes on a related webapp PR when ready.

PR #1368, opened 2021-02-21
MM-32882: Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item

catalintomai:
#### Summary
Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item until an admin user visits the marketplace for the first time

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32882

#### Related PRs
server: https://github.com/mattermost/mattermost-server/pull/16992
api: https://github.com/mattermost/mattermost-api-reference/pull/616
webapp: https://github.com/mattermost/mattermost-webapp/pull/7563


src/actions/general.ts
```diff
@@ -200,6 +200,37 @@ export function getWarnMetricsStatus(): ActionFunc {
     };
 }
 
+export function setFirstAdminVisitMarketplaceStatus(): ActionFunc {
+    return async (dispatch: DispatchFunc) => {
+        try {
+            Client4.trackEvent('api', 'api_request_set_first_admin_visit_marketplace_status');
```
hanzei, line 206: I'm 3/5 that we should track event on the components instead of Redux. That simplifies the number of place to look for a track event and seems to be the idiomatic way in the code base.

src/client/client4.ts
```diff
@@ -2276,6 +2279,20 @@ export default class Client4 {
         );
     }
 
+    setFirstAdminVisitMarketplaceStatus = async () => {
+        return this.doFetch<StatusOK>(
+            `${this.getBaseRoute()}/first_admin_visit_marketplace`,
+            {method: 'post', body: JSON.stringify({first_admin_visit_marketplace_status: true})},
```
hanzei, line 2285: Strictly speaking the server doesn't care about the payload. Should we not send one then?

src/selectors/entities/general.ts
```diff
@@ -104,3 +104,7 @@ export const getManagedResourcePaths: (state: GlobalState) => string[] = createS
 export const getServerVersion = (state: GlobalState): string => {
     return state.entities.general.serverVersion;
 };
+
+export function firstAdminVisitMarketplaceStatus(state: GlobalState): any {
```
hanzei, line 108: Should the code return an `boolean` here instead of `any`?

src/types/general.ts
```diff
@@ -15,4 +15,10 @@ export type GeneralState = {
     serverVersion: string;
     timezones: string[];
     warnMetricsStatus: Dictionary<WarnMetricStatus>;
+    firstAdminVisitMarketplaceStatus: boolean;
 };
+
+export type SystemSetting = {
+    name: string;
+    value: string;
+ };
```
hanzei, line 24: Nit pick: Add empty line

hanzei (changes requested):

src/selectors/entities/general.ts
hanzei, line 108: Maybe 
```suggestion
export function getFirstAdminVisitMarketplaceStatus(state: GlobalState): any {
```
?

larkox (approved):
LGTM

src/actions/general.ts
catalintomai: I'm not particular to one or another (an action is an action), but I'd like to know more about: "seems to be the idiomatic way" - I don't see a definitive pattern one way or another - can you help substantiate the statement?

src/client/client4.ts
catalintomai: While more verbose, I find it easier to follow - maps to how we store the state of the flag in the store (as opposed to interpreting the call at the server as "true" and doing the conversion at the receiving end). Also, it keeps the door open to more interesting scenarios in the future.

src/types/general.ts
```diff
@@ -15,4 +15,10 @@ export type GeneralState = {
     serverVersion: string;
     timezones: string[];
     warnMetricsStatus: Dictionary<WarnMetricStatus>;
+    firstAdminVisitMarketplaceStatus: boolean;
```
jfrerich, line 18: nit: keep the alphabetical organization

src/client/client4.ts
```diff
@@ -34,6 +34,9 @@ import {
 import {CustomEmoji} from 'types/emojis';
 import {ServerError} from 'types/errors';
 import {FileInfo, FileUploadResponse} from 'types/files';
+import {
```
jfrerich, line 37: nit: make a one-liner

jfrerich (approved):
LGTM with two nitpicks

src/client/client4.ts
hanzei: The server does ignore the payload. Hence, the any call to the server is currently interpreted as "true", correct?

Also: How does ignoring the payload closes a door to future feature? It seems the other way around: If we are now conservative with sending the payload, we can add semantics to the payload later.

src/actions/general.ts
hanzei: My initiation comment was a misleading. IMO the event category should be `plugins` because that is what all other marketplace events are also using e.g. `ui_marketplace_search` or `ui_marketplace_opened`. Therefore the event should triggered on the component, as only `api` events are triggered in Redux.

src/client/client4.ts
catalintomai: "Hence, the any call to the server is currently interpreted as "true", correct?" - not disputing that. As I said, I find it easier to follow - also, to me it aligns better with the next layer - don't have to convert the "true" into a  name-value pair.
My second comment was more about - if we ever send the "false"/other state, the door is already opened (still need to do a few changes).

src/actions/general.ts
catalintomai: Ok, I get your comment now. I can see the benefit of using the event category 'plugins', wrt to the redux vs. component, don't really mind one vs. another since there is a 1-1 mapping of the UI action to the redux action (and apparently redux captures other categories besides `api`), but one could argue that the first/original action is the UI one so I'll move it there.

src/reducers/entities/general.test.js
```diff
@@ -0,0 +1,44 @@
...
+                type: GeneralTypes.RECEIVED_FIRST_ADMIN_VISIT_MARKETPLACE_STATUS,
+                data: true,
+            };
+            const expectedState = true;
+
+            const actualState = reducer({firstAdminVisitMarketplaceStatus: state}, action);
+            assert.deepStrictEqual(actualState.firstAdminVisitMarketplaceStatus, expectedState);
+        });
+    });
+});
```
hanzei, line 44: Nit pick: Add empty line

hanzei (approved):
LGTM :+1:

DHaussermann (approved):
LGTM! 
tested as part of https://github.com/mattermost/mattermost-server/pull/16992

hmhealey:
As mentioned elsewhere, I'm going to close this for now while we do the mattermost-redux migration. I'll let you know when it can be reopened against the webapp some time tomorrow.

hmhealey:
@catalintomai This can be resubmitted against the mattermost-webapp version of mattermost-redux now

PR #1378, opened 2021-02-26
Add support for search files

jespino:
#### Summary
Add support for search files

#### Ticket Link
[MM-28019](https://mattermost.atlassian.net/browse/MM-28019)


jespino:
From QA perspective, this shouldn't affect anything until is used for the new search files feature. Maybe a quick smoke testing around searches is more than enough.

src/actions/search.ts
```diff
@@ -102,7 +127,68 @@ export function getMorePostsForSearch(): ActionFunc {
...
+        dispatch({
+            type: SearchTypes.SEARCH_FILES_REQUEST,
+            isGettingMore,
+        });
+
+        let files: FileSearchResults;
+        try {
+            files = await Client4.searchFilesWithParams(teamId, params);
+
+            // await dispatch(getMissingChannelsFromFiles(files.file_infos));
```
mgdelacroix, line 148: :eyes:

jespino: image

jespino: Done

mgdelacroix (approved):

jespino:
@ogi-m this is not changing anything important in the rest of the code, is only adding the machinery to access the new API endpoint. I don't think this need testing by itself, it can be tested after merged as part of the whole feature.

ogi-m (approved):
Thanks @jespino 👍🏻

PR #1382, opened 2021-02-27
[GH-16739] Add optional props attribute to channel type

Crimson-riot:
#### Summary
* Adds optional prop type in the same syntax as `posts.ts`, `integrations.ts` and `sessions.ts` so that the channel_info_moda l(mattermost webapp) can correctly use the channel type when converted to typescript.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16739

### DIFF
--- a/src/types/channels.ts
+++ b/src/types/channels.ts
@@ -42,6 +42,7 @@ export type Channel = {
     status?: string;
     fake?: boolean;
     group_constrained: boolean;
+    props?: Record<string, any>;
 };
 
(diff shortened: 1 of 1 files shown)


hmhealey (approved):
Thanks for the PR, @Crimson-riot! Strange that that field wasn't added previously since it's been around for a while on the server, but I guess we never needed to access it before now.

calebroseland (approved):

PR #1393, opened 2021-03-03
MM-24009: remove redux-action-buffer

lieut-data:
#### Summary
This is a precursor to removing redux-offline altogether, by noting first that redux-action-buffer is only used when `enableBuffer` is left unset or passed as `true`, but the webapp unconditionally passes it as `false`:

https://github.com/mattermost/mattermost-webapp/blob/8676ca150396b1eed898d7a114264bfa4278e884/store/index.js#L215

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-24009

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -213,41 +213,6 @@ SOFTWARE.
 
 ---
 
-## redux-action-buffer
-
-This product contains 'redux-action-buffer' by rt2zz.
-
-Buffer redux actions until a breaker condition is met.
-
-* HOMEPAGE:
-  * https://github.com/rt2zz/redux-action-buffer
-
-* LICENSE: MIT
-
-The MIT License (MIT)
-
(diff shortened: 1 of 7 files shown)


lieut-data:
@hmhealey, I'd love to get a sanity check on my investigation here. From what I can see, this code path was only ever used by the mobile app, but of course their having diverged from `mattermost-redux` proper, we no longer need this.

reflog (approved):

hmhealey (approved):

hmhealey:
Yeah, that looks right to me. It looks like it was used to hold any redux actions until the state was finished rehydrating from disk, but that only really mattered on the mobile app where everything is saved to disk.

hmhealey:
This can be tested in the corresponding web app PR

jgilliam17 (approved):
Thanks @lieut-data 
Tested on the corresponding WebApp PR, looks good to merge.

PR #1396, opened 2021-03-05
MM-24009: remove redux-offline

lieut-data:
#### Summary
Follow the lead of https://github.com/mattermost/mattermost-mobile/pull/4120 and remove redux-offline. The library was responsible for automatically retrying select requests, but the vast majority of API calls remained unaffected. Removing the library simplifies the surface area of the Redux store and corresponding code. Note that this library is discrete from redux-persist, currently responsible for the magic of `storage.storage` and rehydrating previously saved drafts, etc.

I've broken this PR into various discrete PRs that are worth reviewing independently. But to summarize the other changes that were tangled with removing this library:
* Replace `remote-redux-devtools` with `redux-devtools-extension` -- the former has support for React Native apps and routing requests through remotedev.io, but none of this is needed for the WebApp now that Mattermost Redux is unused by the mobile app.
* Fix frozen stores in development: after peeling back the layers of code, it looked like freezing the Redux store in development wasn't actually working. This is completed by corresponding changes in the WebApp.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-24009

Also, see https://github.com/mattermost/mattermost-webapp/pull/7648


src/actions/channels.ts
```diff
@@ -650,26 +650,28 @@ export function leaveChannel(channelId: string): ActionFunc {
...
+                    type: ChannelTypes.RECEIVED_CHANNEL,
+                    data: channel,
+                },
+                {
+                    type: ChannelTypes.RECEIVED_MY_CHANNEL_MEMBER,
+                    data: member,
+                },
+            ]));
+
+            // The category here may not be the one in which the channel was originally located,
```
lieut-data, line 669: I spent a bit of time here, but ultimately decided not to worry about recovering the exact category again.

src/actions/posts.ts
```diff
@@ -381,17 +380,21 @@ export function deletePost(post: ExtendedPost) {
...
-            });
+            try {
+                dispatch({
+                    type: PostTypes.POST_DELETED,
+                    data: delPost,
+                });
+
+                await Client4.deletePost(post.id);
+            } catch (e) {
+                // Recovering from this state doesn't actually work. The deleteAndRemovePost action
```
lieut-data, line 391: I played with this for a while too: calling this comment out.

src/actions/preferences.ts
```diff
@@ -25,22 +25,19 @@ export function deletePreferences(userId: string, preferences: PreferenceType[])
...
-                        type: PreferenceTypes.DELETED_PREFERENCES,
-                    },
-                    rollback: {
-                        type: PreferenceTypes.RECEIVED_PREFERENCES,
-                        data: currentPreferences,
-                    },
-                },
-            },
-        });
+        try {
```
lieut-data, line 28: I kept these, but I'm not convinced they are /useful/ without first-class support in the WebApp. Notably, the UI was a bit wonky when I reproduced these steps.

src/store/configureStore.dev.ts
```diff
@@ -1,28 +1,14 @@
...
-import {createOfflineReducer, networkStatusChangedAction, offlineCompose} from 'redux-offline';
-import defaultOfflineConfig from 'redux-offline/lib/defaults';
+import {createStore, applyMiddleware} from 'redux';
+import {composeWithDevTools} from 'redux-devtools-extension';
 import reducerRegistry from './reducer_registry';
-import devTools from 'remote-redux-devtools';
 
-const windowAny = window as any;
-
-const devToolsEnhancer = typeof windowAny !== 'undefined' && windowAny.__REDUX_DEVTOOLS_EXTENSION__ ? // eslint-disable-line no-underscore-dangle
```
lieut-data, line 12: Note that I opted to depend on `redux-devtools-extension` to contain this logic.

src/actions/channels.ts
hmhealey: Makes sense. Managing the side effects that other operations have on the sidebar is something I really struggled with in that code. That's the reason that I ended up just going back to the server to resync categories so often since it was hard to figure out what the effects of any given operation were. The only other thing I could think to do here would be to get the channel's category before leaving it and then restoring, but then you have to worry about potentially paving over other changes done in the meantime.

src/actions/posts.ts
```diff
@@ -196,79 +196,78 @@ export function createPost(post: Post, files: any[] = []) {
...
-                                type: FileTypes.RECEIVED_FILES_FOR_POST,
-                                postId: payload.id,
-                                data: files,
-                            });
-                        }
-
-                        dispatch(batchActions(actions));
+        });
+
+        dispatch(batchActions(actions, 'BATCH_CREATE_POST_INIT'));
```
hmhealey, line 214: Why give a name to the batched action? Is it to make debugging easier?

src/actions/preferences.ts
hmhealey: I feel like I was adding Preferences right around the time we were getting into the idea of having more optimistic actions, so that might be why this is here. I don't think it's the first time I've said this, but I think it might end up being better to be optimistic about specific preference changes (like favouriting a post or setting your theme) while making the general case non-optimistic for simplicity's sake.

src/store/configureStore.dev.ts
hmhealey: I'd react with thumbs up more times if I could. I've never been certain if anyone used this, and it's always bugged me how much console log spam it generates for something that might not be used.

src/store/configureStore.dev.ts
```diff
@@ -31,40 +17,30 @@ import {GlobalState} from 'types/store';
...
+ * persistConfig - Any additional configuration data to be passed into redux-persist aside from the default values.
  * getAppReducer - A function that returns the appReducer as defined above. Only used in development to enable hot reloading.
  * clientOptions - An object containing additional options used when configuring the redux store. The following options are available:
  *     additionalMiddleware - func | array - Allows for single or multiple additional middleware functions to be passed in from the client side.
  *     enableThunk - bool - default = true - If true, include the thunk middleware automatically. If false, thunk must be provided as part of additionalMiddleware.
  */
-export default function configureServiceStore(preloadedState: any, appReducer: any, userOfflineConfig: any, getAppReducer: any, clientOptions: any) {
-    const baseOfflineConfig = Object.assign({}, defaultOfflineConfig, offlineConfig, userOfflineConfig);
+export default function configureServiceStore(preloadedState: any, appReducer: any, persistConfig: any, getAppReducer: any, clientOptions: any): Store {
```
hmhealey, line 26: Strictly speaking, this is a breaking change for anyone using mattermost-redux to configure their own store. This isn't one of the areas I think we've really implied backwards compatibility though. If we ever do that, we should make this take named arguments in an object instead.

src/store/configureStore.dev.ts
```diff
@@ -75,15 +51,15 @@ export default function configureServiceStore(preloadedState: any, appReducer: a
...
+            store.replaceReducer(enableFreezing(createDevReducer(baseState, reducerRegistry.getReducers(), nextServiceReducer, nextAppReducer)));
         });
     }
 
     return store;
 }
 
 function createDevReducer(baseState: any, ...reducers: any) {
-    return enableFreezing(createReducer(baseState, ...reducers));
+    return createReducer(baseState, ...reducers);
```
hmhealey, line 62: Why move `enableFreezing` out of here? It looks like you're always calling them together anyway.

src/store/configureStore.dev.ts
lieut-data: I definitely rely on Redux DevTools, and hope we can do so /more/ as we shift more logic to the Redux bus. So just to avoid ambiguity here -- the integration still exists -- but the `redux-devtools-extension` and it exposing `composeWithDevTools` just encapsulates all the magic here.

src/store/configureStore.dev.ts
lieut-data: Agreed re: breaking change, and that we're not explicitly supporting backwards compatibility here. I'm not really sure we should, either, but open for debate!

src/store/configureStore.dev.ts
lieut-data: Ooh, good catch -- when I originally removed `createOfflineReducer`, I started getting exceptions around the store being mutated. So to isolate and fix that, I called them in the order `enableFreezing(createOfflineReducer(createDevReducer))` and made sure freezing worked before ripping out `createOfflineReducer` afterwards. The final delta is unnecessary, though -- and in fact `createDevReducer` is largely superfluous at this point.

In fact, I'm not entirely convinced it's necessary to distinguish `.dev.ts` from `.prod.ts` in the first place, given tree-shaking and what not. Either way, I'll clean up this anomaly.

src/actions/posts.ts
devinbinnie: Is this a follow-up ticket to have the delete modal report back to the user? Or maybe pop a toastr or something to say the post was never deleted?

devinbinnie (approved):

src/actions/posts.ts
lieut-data: Good call: filed https://mattermost.atlassian.net/browse/MM-33634 to track.

src/actions/posts.ts
lieut-data: I admit I copied this directly from the corresponding mobile repository, but yes, it does enhance the experience when debugging via Redux DevTools:

image

hmhealey: I wonder why it's "INIT" then. Regardless, that looks fine since it's probably better than seeing dozens of BATCH_ACTION in the dev tools

lieut-data: In this case, it's to distinguish from the `BATCH_CREATE_POST` batch action a few lines later (and `BATCH_CREATE_POST_FAILED` later on). 0/5 on the naming.

hmhealey (approved):

hmhealey:
@lieut-data I've added the Awaiting PR label since if this goes in first and someone updates their version of mattermost-redux in the webapp, then it'll break due to the change in parameters. One of the many reasons I'm looking forward to getting rid of mattermost-redux

lieut-data:
Re-requesting dev review given the changes. Might be easiest to review with whitespace diff disabled. See https://github.com/mattermost/mattermost-webapp/pull/7648#issuecomment-797692291 for context.

src/actions/channels.ts
```diff
@@ -652,25 +652,27 @@ export function leaveChannel(channelId: string): ActionFunc {
...
-                    data: member,
-                },
-            ]));
-
-            // The category here may not be the one in which the channel was originally located,
-            // much less the order in which it was placed. Treating this as a transient issue
-            // for the user to resolve by refreshing or leaving again.
-            dispatch(addChannelToInitialCategory(channel, false));
-        }
+        (async function removeFromChannelWrapper() {
```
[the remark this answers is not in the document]
hmhealey: Oh, I see. It's virtually the same as just dispatching another async action that does this without waiting for it. I'm not a fan of that sort of set and forget stuff, but I guess it's no different than what we did before.

hmhealey (changes requested):

src/actions/channels.test.js
```diff
@@ -2125,6 +2131,49 @@ describe('Actions.Channels', () => {
...
+
+            nock(Client4.getBaseRoute()).
+                delete(`/channels/${channel.id}/members/${user.id}`).
+                reply(500, {});
+
+            await store.dispatch(Actions.leaveChannel(channel.id));
+
+            const state = store.getState();
+
+            setTimeout(() => {
```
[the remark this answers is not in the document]
hmhealey: Actually, you could probably get away with `await Promise.resolve(true)` to further simplify it since we only need to preempt this code to let the async block in the action run.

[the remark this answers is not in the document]
lieut-data: Ooh, I like the promise approach instead of wrapping. However, I tried the `await`, and it failed for one test case, I think because of the interdependence on mocking the responses which probably requires a chain of pre-emptions.

hmhealey (approved):

tkiethanom:
It doesn't look like this change was pushed to the npm version 5.33.1. I'm still seeing `redux-offline` in the dependencies list.

PR #1408, opened 2021-03-11
Remove async/await and some mocks from tests

hmhealey:
async/await and this mocking seemed to cause issues after this was moved into the web app. I'm not 100% sure that these will solve those issues, but they seem decent enough to clean up anyway.

### DIFF
--- a/test/test_helper.js
+++ b/test/test_helper.js
@@ -453,7 +453,7 @@ class TestHelper {
         this.basicGroup = this.fakeGroupWithId();
     }
 
-    initBasic = async (client4 = this.createClient4()) => {
+    initBasic = (client4 = this.createClient4()) => {
         client4.setUrl(DEFAULT_SERVER);
         this.basicClient4 = client4;
 
--- a/test/test_helper.js
+++ b/test/test_helper.js
@@ -469,7 +469,7 @@ class TestHelper {
         };
     };
 
-    tearDown = async () => {
+    tearDown = () => {
         nock.restore();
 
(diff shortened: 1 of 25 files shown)


lieut-data:
To confirm, are all these `async` / `await` spurious, in that none of the methods return promises?

test/test_helper.js
```diff
@@ -453,7 +453,7 @@ class TestHelper {
         this.basicGroup = this.fakeGroupWithId();
     }
 
-    initBasic = async (client4 = this.createClient4()) => {
+    initBasic = (client4 = this.createClient4()) => {
```
lieut-data, line 456: Oh, I see we've actually stripped this here -- that makes more sense now!

lieut-data (approved):
Reviewed more carefully and realized we're stripping it holistically -- nice!

test/test_helper.js
hmhealey: Yeah, I don't know why it was even there to begin with. It was probably something we needed at some point, but never ended up using after that. We have had a tendency to overuse async/await frequently in this code.

nevyangelova (approved):
LGTM not sure why some of those were there to begin with though? Are we sure its not needed somewhere, maybe a promise is returned somewhere?

hmhealey:
I think we could've used a promise somewhere, but it doesn't seem necessary any more at least. I'd prefer to simplify things by removing the async/awaits wherever is possible since it makes integrating the tests with the webapp code easier.

PR #1410, opened 2021-03-15
Add NoAccounts property to ClientConfig type

EricMontague:
#### Summary
This PR adds the NoAccounts property to the ClientConfig type alias. While migrating the SignupEmail component to Typescript in the webapp repo, I discovered that this property wasn't present on the ClientConfig type.

#### Ticket Link
N / A

### DIFF
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -143,7 +143,7 @@ export type ClientConfig = {
     MaxFileSize: string;
     MaxNotificationsPerChannel: string;
     MinimumHashtagLength: string;
-    NoAccounts: string;
+    NoAccounts: 'true' | 'false';
     OpenIdButtonText: string;
     OpenIdButtonColor: string;
     PasswordMinimumLength: string;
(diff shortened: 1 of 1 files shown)


devinbinnie:
So this field was added in this PR just today: https://github.com/mattermost/mattermost-redux/pull/1412
Not sure if we need this change, though I do like that we restrict the field to just a stringified boolean.

hmhealey:
I think it's safe to close this then. We can always make the the typing stricter later if wanted, but that one solves the main problem here.

@EricMontague If you could update your webapp PR to the latest master and its mattermost-redux version, everything should work then.

EricMontague:
> I think it's safe to close this then. We can always make the the typing stricter later if wanted, but that one solves the main problem here.
> 
> @EricMontague If you could update your webapp PR to the latest master and its mattermost-redux version, everything should work then.

Okay I pulled in the changes from master into my branch and pointed it the main mattermost-redux repo.

PR #1412, opened 2021-03-15
[GH_16717/MM_20389] Migrate 'components/signup/signup_controller' module and associated tests to TypeScript

aggmoulik:
Signed-off-by: Moulik Aggarwal <qwertymoulik@gmail.com>

#### Summary
It changes the name of the SubscriptionStats in camelcase as it was written in bad mannered and added a property which is used in the react web app.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-20389

### DIFF
--- a/src/types/cloud.ts
+++ b/src/types/cloud.ts
@@ -110,6 +110,6 @@ export type InvoiceLineItem = {
 }
 
 export type SubscriptionStats = {
-	remainingSeats: number;
-	isPaidTier: string;
+	remaining_seats: number;
+	is_paid_tier: string;
 }
--- a/src/types/cloud.ts
+++ b/src/types/cloud.ts
@@ -110,6 +110,6 @@ export type InvoiceLineItem = {
 }
 
 export type SubscriptionStats = {
-	remaining_seats: number;
-	is_paid_tier: string;
+	remainingSeats: number;
+	isPaidTier: string;
 }
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -179,6 +179,7 @@ export type ClientConfig = {
     WebsocketPort: string;
     WebsocketSecurePort: string;
     WebsocketURL: string;
+    NoAccounts: string;
 };
 
 export type License = {
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -143,6 +143,7 @@ export type ClientConfig = {
     MaxFileSize: string;
     MaxNotificationsPerChannel: string;
     MinimumHashtagLength: string;
+    NoAccounts: string;
     OpenIdButtonText: string;
     OpenIdButtonColor: string;
     PasswordMinimumLength: string;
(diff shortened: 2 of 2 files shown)


src/types/cloud.ts
```diff
@@ -110,6 +110,6 @@ export type InvoiceLineItem = {
 }
 
 export type SubscriptionStats = {
-	remaining_seats: number;
-	is_paid_tier: string;
+	remainingSeats: number;
```
devinbinnie, line 113: We can't change the names of these without changing what's being passed down from the server, since it's mapped using snake case naming. I think these names will need to be changed on the webapp PR instead.

devinbinnie (changes requested):
Added a comment for you to review @aggmoulik.
Thanks for the contribution!

src/types/cloud.ts
aggmoulik: I created a PR for the same on the web-app repository and if you want me to remove these naming convention changes then it will not pass the type-checking.

lieut-data: Agree with @devinbinnie -- until we go to the effort of renaming all the "incoming" fields from the server, we're stuck with the snack case here. Can we supply instructions to override the linter in this case?

lieut-data (changes requested):

src/types/cloud.ts
devinbinnie: Just to confirm @aggmoulik, you're saying the linter is causing the error or the type checker? If it's the type checker, there's probably another instance that needs to be changed somewhere as well.

If it's the linter, then I think we need to change our ruleset to allow for snake case unless we plan on changing this convention app-wide.

aggmoulik:
@lieut-data I can do it in the same PR of webapp.

src/types/cloud.ts
hmhealey: Having the server code return snake case was intentional way back when, so I don't think there's any plan to change that any time soon.

I'd say we could update the ESLint rules for that now, but we're going to be getting rid of mattermost-redux as a separate repository later this week, so I'd say to leave the casing as-is and use an `// eslint-ignore` wherever the warnings are thrown for the time being.

aggmoulik:
Hy @lieut-data @devinbinnie , I have removed the naming in redux and also fixed the web-app PR. So, please review this so that web-app PR can pass all checks.

devinbinnie (approved):
LGTM!

lieut-data (changes requested):
Thanks, @aggmoulik! Just one comment below.

lieut-data:
Weird, my comment didn't actually show up -- the ask was to sort the new entry alphabetically with the rest.

aggmoulik:
> Weird, my comment didn't actually show up -- the ask was to sort the new entry alphabetically with the rest.

Updated @lieut-data

lieut-data (approved):
Thanks!!

devinbinnie (approved):

PR #1417, opened 2021-03-19
Proposal: Introduce EPICS to handle side effects from redux actions

AGMETEOR:
#### Summary
[Motivation Ticket 1](https://mattermost.atlassian.net/browse/MM-33132?atlOrigin=eyJpIjoiMDg1Y2Q1NTdkMGE5NDJiYjhmNGNiMGUyZWY3MTU3Y2EiLCJwIjoiaiJ9)
[Motivation Ticket 2](https://mattermost.atlassian.net/browse/MM-33374?atlOrigin=eyJpIjoiYWE2NjFiMTkxYzE3NDQwMjg3N2IyYmM0NTNmZmIzMzgiLCJwIjoiaiJ9)

#### Example 1
In Motivation Ticket 2,  we had a bug caused by this flow. First, at boot time, if we had a cloud licence we would load `subscriptionStats` [here](https://github.com/mattermost/mattermost-webapp/blob/eb12f73a071bd0a4b49c4fa5b1cf673ac251521a/actions/views/root.js#L28). The bug occurred when, at boot time, the license was not cloud, and then later a user witches the license to cloud. This means there'll be components expecting to use ```subscriptionStats``` but they were not loaded. The components were nolonger working properly. To fix, we might have to go into each of the components and in ```compnentDidMount``` or ```useEffect``` hook check whether the subscriptionStats exist and then fetch them.

A more compact and predictable solution would be to listen to redux events and then react to them. In this demo PR, I use ```redux-observable``` library and ```rxjs``` to listen to ```UPLOAD_LISENCE_SUCCESS``` redux action and then as a side effect trigger fetching of ```subscriptionStats``` if the uploaded lisence is a Cloud one.

In Motivation Ticket 1, the solution was to listen to a webhook event and then update the data. This can also easily be solved with an appropriate epic.

#### Ticket Link
None


hmhealey:
Hey Allan. Just a heads up that this repo is technically frozen right now pending it being moved into the webapp with https://github.com/mattermost/mattermost-webapp/pull/7723.

AGMETEOR:
> Hey Allan. Just a heads up that this repo is technically frozen right now pending it being moved into the webapp with [mattermost/mattermost-webapp#7723](https://github.com/mattermost/mattermost-webapp/pull/7723).

Hey Harrison, thanks for the heads up.

hmhealey:
For the Epics themselves, we'd definitely be interested in discussing them with the web guild. I haven't looked into Epics myself, but we're actually going to be trying out Redux Sagas at some point since Jesse has previous experience with them. It sounds like they'll be able to fix similar problems too!

AGMETEOR:
> For the Epics themselves, we'd definitely be interested in discussing them with the web guild. I haven't looked into Epics myself, but we're actually going to be trying out Redux Sagas at some point since Jesse has previous experience with them. It sounds like they'll be able to fix similar problems too!

Yes. I think redux saga and redux-observable do the same thing.

hmhealey:
@AGMETEOR mattermost-redux has been merged into the web app now, so you can feel free to resubmit this there.

